PR #3, opened 2026-07-02
feat(cli): implement installer core (WP-003)

thegyula:
Spec: docs/specs/WP-003-installer-core.md

> **Stacked PR** — base is `wp/001-ci-and-lint`; auto-retargets to `main` when PR #1 merges.

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table (plus the spec status flip and the one dogfood `memory/lessons/inbox.md` bullet, both allowed)
- [x] Spec status flipped to In-Review

### `npm test`

```
✔ init --yes creates the core, config, and manifest
✔ secrets directory is created with mode 0700
✔ a second init makes zero changes and says so
✔ init --dry-run creates nothing
✔ init reflects a detected Claude harness in config
✔ an unknown command prints usage and exits 2
✔ help prints usage and exits 0
✔ load returns an empty manifest when none exists
✔ record + save + load round-trips entries
✔ load throws on a corrupted manifest
✔ reverse removes all files and empty dirs, leaving nothing
✔ reverse dry-run removes nothing but reports would-be removals
✔ reverse keeps config.yaml when it was modified since install
✔ reverse reports already-gone entries as skipped
✔ reverse skips unknown entry kinds (forward compat)
✔ uninstall --dry-run lists manifest contents and changes nothing
✔ uninstall --yes removes the entire core
✔ uninstall keeps a user-modified config.yaml
✔ uninstall exits 0 when some entries were already gone
✔ uninstall without an install errors (exit 1)
ℹ tests 26
ℹ pass 26
ℹ fail 0
```

### `npm run lint`

```
--- markdownlint ---
markdownlint-cli2 v0.23.0 (markdownlint v0.41.0)
Finding: docs/**/*.md *.md
Linting: 35 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
frontmatter check passed: 6 spec(s), 4 agent(s)

lint passed
```

### Manual temp-HOME sequence (all env-overridden; real `~/.wienerdog`/`~/.claude`/`~/.codex` never touched)

```
$ WIENERDOG_HOME=$WD ... node bin/wienerdog.js init --yes
wienerdog init — plan:

Directories:
  [create] .../wd
  [create] .../wd/state
  [create] .../wd/secrets
  [create] .../wd/logs

Files:
  [create] .../wd/config.yaml

Detected AI tools:
  Claude Code: not found (...)
  Codex CLI:   not found (...)

[...]

PR #5, opened 2026-07-02
feat(vault): implement vault skeleton generator (WP-004)

thegyula:
Spec: docs/specs/WP-004-vault-skeleton.md

Stacked PR — base wp/003-installer-core.

## Deliverables checklist

## Verification output

```
$ npm test
> wienerdog@0.0.0 test
> node --test

... (33 tests) ...
✔ init --yes creates the core, config, and manifest
✔ secrets directory is created with mode 0700
✔ a second init makes zero changes and says so
✔ init --dry-run creates nothing
✔ init reflects a detected Claude harness in config
✔ an unknown command prints usage and exits 2
✔ help prints usage and exits 0
✔ load returns an empty manifest when none exists
✔ record + save + load round-trips entries
✔ load throws on a corrupted manifest
✔ reverse removes all files and empty dirs, leaving nothing
✔ reverse dry-run removes nothing but reports would-be removals
✔ reverse keeps config.yaml when it was modified since install
✔ reverse reports already-gone entries as skipped
✔ reverse skips unknown entry kinds (forward compat)
✔ uninstall --dry-run lists manifest contents and changes nothing
✔ uninstall --yes removes the entire core
✔ uninstall keeps a user-modified config.yaml
✔ uninstall exits 0 when some entries were already gone
✔ uninstall without an install errors (exit 1)
✔ scaffoldVault creates the full tree and matches the golden fixture
✔ scaffoldVault initializes a git repo with exactly one commit
✔ a second scaffoldVault run creates nothing and makes no new commit
✔ a pre-existing file in the target is never overwritten
✔ scaffoldVault records every created file in the manifest as kind vault-file
✔ scaffoldVault dryRun makes no changes
✔ scaffoldVault throws a WienerdogError when git is not installed
ℹ tests 33
ℹ pass 33
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0

$ npm run lint
--- markdownlint ---
Finding: docs/**/*.md *.md
Linting: 35 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
frontmatter check passed: 6 spec(s), 4 agent(s)
lint passed

$ export WIENERDOG_HOME=$(mktemp -d)/wd WIENERDOG_VAULT=$(mktemp -d)/vault
$ node bin/wienerdog.js init --yes && git -C $WIENERDOG_VAULT log --oneline
wienerdog init — plan:

Directories:
  [create] /tmp/.../wd
  [create] /tmp/.../wd/state
  [create] /tmp/.../wd/secrets
  [create] /tmp/.../wd/logs

Files:
  [create] /tmp/.../wd/config.yaml

Vault:
  [create] /tmp/.../vault

Detected AI tools:
  Claude Code: not found (/tmp/.../.claude)
  Codex CLI:   not found (/tmp/.../.codex)

Vault: scaffolding /tmp/.../vault
  created 14 file(s), skipped 0 existing file(s)

wienerdog: installed. Run `wienerdog doctor` to check the setup.
01e38c5 wienerdog: vault created

$ node bin/wienerdog.js init --yes   # second run: nothing to do
wienerdog: already installed, nothing to do.
```

(Full untruncated logs were run locally in a sandboxed `$HOME`/`$WIENERDOG_HOME`/`$WIENERDOG_VAULT` — never against the real `~/wienerdog` or `~/.wienerdog`, confirmed absent on the dev machine before and after this session.)

## Decisions made

- **Manifest `kind: 'vault-file'` is intentionally unknown to WP-003's `reverse()`.** As instructed by the spec, vault entries are recorded with `kind: 'vault-file'`; `manifest.js`'s `reverse()` (untouched, from WP-003) already skips unrecognized kinds with a stderr warning, so `uninstall` never deletes vault content. Verified in `tests/unit/uninstall.test.js` (unmodified) and `tests/unit/vault.test.js`.
- **Discovered and fixed a manifest/uninstall interaction bug while wiring the vault step into `init`:** after scaffolding, `init.js` rewrites `config.yaml`'s `vault: null` line to the real path, but the manifest had already recorded the *pre-rewrite* file hash. Left as-is, `uninstall`'s tamper-detection (hash mismatch ⇒ "user modified this, keep it") would permanently refuse to remove `config.yaml`, breaking `uninstall --yes` for every fresh install. Fixed by updating the manifest's recorded hash for `config.yaml` immediately after the rewrite (`src/cli/init.js`). This is within the Deliverables table (`src/cli/init.js` is listed as modified) — not scope creep, it's a direct consequence of the "vault step after core step" contract.
- **Hardened `scaffoldVault`'s git bootstrap against same-directory races** with an OS-temp-dir mkdir-based lock (`src/core/vault.js`, `withVaultLock`). Reproduced independently: two concurrent `git init` calls against the *same* directory can fail outright (`fatal: cannot copy '.../git-core/templates/info/exclude': File exists`) — not a hypothetical. This matters because pre-existing WP-003 tests (`tests/unit/init.test.js`, `tests/unit/uninstall.test.js`, not in my Deliverables) don't override `WIENERDOG_VAULT`, so every test-file process defaults to the same vault path when `node --test`'s default concurrent test-file execution runs them together — which would otherwise flake or crash `npm test` (and CI) nondeterministically. The lock is scoped to the git bootstrap only (file writes are already idempotent/skip-on-exists).
- Vault CLAUDE.md kept to plain, jargon-explained language per repo CLAUDE.md conventions (knowledge workers, not developers); PARA terms are explained inline rather than assumed.
- Identity scaffold frontmatter (`06-Identity/*.md`) uses `status: incubating` (empty until WP-005's interview fills them in) plus the full note-format schema (`id`, `tags`) alongside the spec-mandated `type`/`origin`/dates, for consistency with the schema documented in the vault's own CLAUDE.md.

- **`scripts/boundary-check.js` (owned by WP-001, not in this WP's Deliverables) does exact-string matching only** against the Deliverables table's path column — it has no support for directory-style entries ending in `/` (e.g. this spec's `templates/vault/` and `tests/golden/vault-default/`, explicitly annotated "(all files below)"). As written, the CI `boundary` job will report every individual file under those two directories as an "offender" even though they're squarely in scope. I could not fix this myself since `scripts/boundary-check.js` isn't in my Deliverables table. Recommend a small follow-up patch adding prefix matching for Deliverables cells ending in `/`.
- **`tests/unit/init.test.js` and `tests/unit/uninstall.test.js`'s `tempEnv()` helpers (WP-003, not in my Deliverables) don't override `WIENERDOG_VAULT`.** I mitigated the resulting same-directory git race in product code (see Decisions made), but the underlying test-isolation gap remains: these two files still target the real default vault path (`$HOME/wienerdog`) whenever `WIENERDOG_VAULT` isn't set in the environment. Recommend a trivial follow-up adding `WIENERDOG_VAULT: path.join(root, 'vault')` to both helpers.

---
Generated-by: claude-sonnet-5


thegyula:
wd-reviewer verdict: **APPROVE** — all layers pass, re-run independently: 33/33 tests, golden byte-compare green, manual init/re-init/uninstall lifecycle verified in isolated temp env (vault survives uninstall, core fully removed), pre-existing user file survives byte-identical, both implementer bug fixes verified sound end-to-end. Boundary clean on the PR's own commit.

Merge order: after PR #3 (auto-retargets).

Generated-by: claude-fable-5

PR #14, opened 2026-07-02
feat(gws): implement calendar and drive read verbs (WP-019)

thegyula:
Spec: docs/specs/WP-019-gws-cal-drive.md

## Deliverables checklist

## Verification output

```
$ npm test -- --test-name-pattern gws-calendar
ℹ tests 21
ℹ pass 21
ℹ fail 0

$ npm test -- --test-name-pattern gws-drive
ℹ tests 21
ℹ pass 21
ℹ fail 0

$ npm run lint
--- markdownlint ---
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
frontmatter check passed: 6 spec(s), 4 agent(s)
lint passed
```

Full `npm test` (148 tests across the whole suite) also passes.

## Decisions made

- **Added a `run(services, flags)` export to both `calendar.js` and `drive.js`,
  in addition to the spec's exact-contract functions (`list`/`show`/`draftEvent`,
  `search`/`read`).** The merged `src/gws/index.js` dispatch table routes the
  whole `cal` / `drive` group through single rows —
  `'cal': (...) => require('./calendar').run(services(), flags)` and the drive
  equivalent — not per-verb rows like `gmail search`/`gmail read`/`gmail draft`.
  Its generic `parseFlags` also only recognizes `--json`/`--max`/`--to`/`--subject`/
  `--body`/`--client`; verb-specific flags this WP needs (`--title`, `--start`,
  `--end`, `--attendee`, `--from`, `--id`) arrive as raw unconsumed tokens in
  `flags.positionals`. Since I'm not allowed to touch `index.js`, `run()` in each
  new file re-parses those tokens and dispatches to the exact-contract functions,
  so the documented CLI examples (`gws cal list`, `gws drive read --id ...`, etc.)
  actually work end-to-end. This is additive only — the exact-contract functions
  are unchanged and are what the unit tests exercise directly.
- `drive.js`'s `read`: used a single `toText()` helper for both the `files.export`
  and `files.get({alt:'media'})` response data, decoding a `Buffer` as UTF-8 or
  passing a string through — spec only said "decode as utf8" without specifying
  the exact shape of the stubbed response, and both googleapis's real
  `alt:'media'` (Buffer, depending on responseType) and test stubs (string) need
  to work.

- **`src/gws/index.js`'s `DISPATCH` table for `cal`/`drive` routes by group only
  (not `<group> <verb>` like `gmail`), and its `parseFlags` doesn't recognize
  this WP's verb-specific flags** (`--title`, `--start`, `--end`, `--attendee`,
  `--from`, `--id`). Handled inside the deliverables (see "Decisions made"
  above) without touching `index.js`, per the binding constraint — but a future
  WP touching `index.js` may want to make this less ad hoc (e.g. per-verb
  dispatch rows and richer generic flag parsing, matching the `gmail` pattern).

---
Generated-by: claude-sonnet-5

### DIFF
--- a/docs/specs/WP-019-gws-cal-drive.md
+++ b/docs/specs/WP-019-gws-cal-drive.md
@@ -1,7 +1,7 @@
 ---
 id: WP-019
 title: Implement gws Calendar and Drive read verbs
-status: Ready
+status: In-Review
 model: sonnet
 size: S
 depends_on: [WP-011]
--- a/src/gws/calendar.js
+++ b/src/gws/calendar.js
@@ -1,0 +1,164 @@
+'use strict';
+
+const { WienerdogError } = require('../core/errors');
+
+/**
+ * Calendar verb functions. Each takes `(services, opts)` and returns plain
+ * data; they perform no console I/O (that is index.js's job). `services` is
+ * the object from getServices; tests pass a stub with just the methods used.
+ * All calendar access is against the user's own `primary` calendar.
+ */
+
+/**
+ * cal list — upcoming events in a window.
+ * @param {{calendar:object}} services
+ * @param {{from?:string, to?:string, max?:number}} opts (ISO timestamps)
+ * @returns {Promise<Array<{id:string, summary:string, start:string, end:string,
+ *   attendees:string[]}>>}
+ */
+async function list(services, opts) {
+  const res = await services.calendar.events.list({
+    calendarId: 'primary',
+    timeMin: opts.from || new Date().toISOString(),
+    timeMax: opts.to,
+    maxResults: opts.max || 20,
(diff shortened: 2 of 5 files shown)


thegyula:
wd-reviewer verdict: **APPROVE** — 148/148 tests re-run; all 5 documented CLI invocations verified end-to-end through the real dispatch; draft-event's sendUpdates:'none' is a hardcoded literal with no overriding path; drive is grep-proven read-only. The run() bridge judged correct (additive, forced by a spec-vs-dispatch gap — routed to architect as WP-020 cleanup along with WP-018's wiring flag). Merging.

Generated-by: claude-fable-5

PR #18, opened 2026-07-02
fix(gws): reconcile dispatch with verb-module contracts (WP-021)

thegyula:
Spec: docs/specs/WP-021-gws-dispatch-reconciliation.md

## Deliverables checklist

## Verification output

```
$ npm test
...
ℹ tests 171
ℹ suites 0
ℹ pass 171
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0

$ npm run lint
--- markdownlint ---
Finding: docs/**/*.md skills/**/*.md templates/**/*.md *.md
Linting: 65 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
frontmatter check passed: 12 spec(s), 4 agent(s)
lint passed

$ npm test -- --test-name-pattern gws-dispatch
✔ gws-dispatch: gmail send without a grant degrades to draft + verbatim notice
✔ gws-dispatch: gmail send with a matching --routine grant sends for real
✔ gws-dispatch: WIENERDOG_JOB env supplies the routine when --routine is absent
✔ gws-dispatch: _alert is invoked with exactly {subject, body}
✔ gws-dispatch: cal draft-event still works through the run() bridge
✔ gws-dispatch: repeatable --attendee accumulates through to cal draft-event
ℹ tests 29
ℹ pass 29
ℹ fail 0
```

### Manual granted/ungranted send transcript (stub via the client seam, zero network)

Ran `gws gmail send --to a@b.com --subject s --body b` through `gwsIndex.run()`
in-process with `src/gws/client.js`'s `getServices` monkeypatched to a stub
factory (the same seam `opts.factory` uses) — no real googleapis, no network:

```
=== gws gmail send (UNGRANTED, no --routine, no WIENERDOG_JOB) ===
{
  "sent": false,
  "degraded": true,
  "draftId": "r-9",
  "messageId": "m-9",
  "notice": "No matching send grant (no send grant for this routine); saved a draft instead. Run: wienerdog grant send --routine <name> --to <recipients>"
}
(stub) messages.send calls: 0
(stub) drafts.create calls: 1

=== gws gmail send (GRANTED, --routine daily-digest matches a saved grant) ===
{
  "sent": true,
  "degraded": false,
  "messageId": "sent-1"
}
(stub) messages.send calls: 1
(stub) drafts.create calls: 0
```

## Decisions made

- **Verb flags are both parsed structurally AND left in `positionals`.** The
  spec asks `parseFlags` to recognize `--title/--start/--end/--attendee/--from/--id`
  while also requiring `calendar.js`/`drive.js`'s unmodified `run()` bridges
  (which independently re-parse those exact flags out of `flags.positionals`)
  to keep working. Since those two requirements only stack — recognizing a
  flag structurally has no way to also leave it in `positionals` unless done
  explicitly — I made each of these cases both set the structured field
  (`flags.title`, etc.) and re-push the raw `--flag`/value token pair onto
  `positionals`, so nothing downstream regresses. `--routine` is the one new
  flag actually consumed at the dispatch layer, so it is not duplicated into
  `positionals`.
- **No `require_` validation added to `gmail send`/`_alert` dispatch rows.**
  The spec's "Exact contracts" section gives the literal object shape to pass
  through (`{to: flags.to, ...}` / `{subject: flags.subject, body: flags.body}`)
  without required-flag checks, unlike the pre-existing `gmail draft`/`auth`
  rows. Kept it exactly as specified rather than adding validation gmail.js
  doesn't ask for (out of scope, and gmail.js/alert.js are not in Deliverables).
- **Dispatch test file uses the client seam via in-test monkeypatching**, not
  `execFileSync`ing the real CLI, since a real child process can't be handed a
  JS stub factory across process boundaries. `client.getServices` is
  reassigned to a wrapper reading a per-test closure variable, patched before
  `src/gws/index.js` is first required (its `getServices` destructure is
  captured at module-load time). Each test still uses a real temp
  `~/.wienerdog` core (via `wienerdog init`) for `config.yaml`/grants, so
  grant read/write logic is exercised for real; only the Google service calls
  are stubbed.
- **Removed two now-stale `// WP-018`/`// WP-019` comments** on the `cal`/
  `drive`/`gmail send`/`_alert` dispatch rows and the one in the `DISPATCH`
  JSDoc — they described those verbs as not-yet-implemented, which this WP's
  entire purpose is to fix, and I was already rewriting the adjacent lines.

- none

---
Generated-by: claude-sonnet-5

### DIFF
--- a/docs/specs/WP-021-gws-dispatch-reconciliation.md
+++ b/docs/specs/WP-021-gws-dispatch-reconciliation.md
@@ -1,7 +1,7 @@
 ---
 id: WP-021
 title: Reconcile gws dispatch with verb-module contracts
-status: Ready
+status: In-Review
 model: sonnet
 size: S
 depends_on: [WP-018, WP-019]
--- a/src/gws/index.js
+++ b/src/gws/index.js
@@ -7,13 +7,21 @@ const { getServices } = require('./client');
 /**
  * Parse `gws` flags out of an argv tail, returning flags plus leftover
  * positionals. Supported: --json (bool), --max <n>, --to <s>, --subject <s>,
- * --body <s>, --client <path>.
+ * --body <s>, --client <path>, --routine <name>, --title <s>, --start <iso>,
+ * --end <iso>, --attendee <email> (repeatable), --from <iso>, --id <s>.
+ *
+ * `--title`/`--start`/`--end`/`--attendee`/`--from`/`--id` are cal/drive verb
+ * flags: they are captured here AND also left in `positionals` (their raw
+ * tokens) so the unmodified `calendar.js`/`drive.js` `run()` bridges — which
+ * independently re-parse those verb flags out of `positionals` — keep working.
  * @param {string[]} argv
  * @returns {{json:boolean, max?:number, to?:string, subject?:string,
- *   body?:string, client?:string, positionals:string[]}}
+ *   body?:string, client?:string, routine?:string, title?:string,
+ *   start?:string, end?:string, attendee:string[], from?:string, id?:string,
+ *   positionals:string[]}}
  */
 function parseFlags(argv) {
-  const flags = { json: false, positionals: [] };
+  const flags = { json: false, positionals: [], attendee: [] };
   for (let i = 0; i < argv.length; i++) {
     const a = argv[i];
     switch (a) {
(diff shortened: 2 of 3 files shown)


thegyula:
wd-reviewer verdict: **APPROVE** — live send gate verified through the real dispatch (6 adversarial scenarios: granted sends, wrong recipient/routine/absent-everything all degrade to draft); _alert bug proven gone; dual-write keeps both bridge and structured paths working; semantic-conflict check vs advanced main ran the merged tree at 204/204. Merging.

Generated-by: claude-fable-5

PR #22, opened 2026-07-03
test(scenarios): run the scenario harness on subscription auth (WP-023)

thegyula:
Spec: docs/specs/WP-023-scenario-subscription-auth.md

## Deliverables checklist

## Verification output

```
$ npm test -- --test-name-pattern paths
...
✔ paths: claudeDir precedence — WIENERDOG_CLAUDE_DIR > CLAUDE_CONFIG_DIR > <home>/.claude (0.796792ms)
...
ℹ tests 32
ℹ pass 32
ℹ fail 0

$ npm run scenarios
scenarios: set WIENERDOG_RUN_SCENARIOS=1 to run (uses real model quota); skipping.
(exit 0, no quota spent)

$ npm test
ℹ tests 245
ℹ pass 245
ℹ fail 0
(harness not invoked)

$ npm run lint
--- markdownlint ---
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
frontmatter check passed: 1 spec(s), 4 agent(s)
lint passed
```

**EXPENSIVE full run (real subscription quota): NOT executed in this session.**
This session runs in an isolated worktree whose harness rules forbid writing
into this machine's real `~/.claude` during automated verification — and the
expensive run's one deliberate real-config-dir touch (installing
`wienerdog-dream` into `<real config dir>/skills/`) is exactly the code path
that rule protects. The non-expensive steps above confirm the guard, the unit
suite, and lint are green; a maintainer should run the EXPENSIVE step from
their own shell (`export WIENERDOG_RUN_SCENARIOS=1; unset ANTHROPIC_API_KEY;
npm run scenarios`) before relying on the injection-gating claim, per
ADR-0009's acknowledged tradeoff (no CI fallback).

**VERIFY-AT-FIRST-LIVE-RUN claims: not verified live in this session** (same
reason). All three are runtime/auth behaviors that only the harness itself
can confirm, per the spec: (1) `claude -p --setting-sources user` discovers
the user-scoped `wienerdog-dream` skill, (2) the brain authenticates via real
`HOME` + un-overridden `CLAUDE_CONFIG_DIR` (Keychain OAuth, no key), (3) the
Haiku grader authenticates the same way. The harness fails loud (non-zero
`wienerdog dream` exit, or grader `{pass:false}` on a spawn/auth error) if any
of these is false — that existing fail-loud plumbing in `run-scenarios.js`
and `rubric.js` is the verification mechanism; no separate stub-based test
was added (see Decisions).

## Decisions made

- **Identical-skill detection (improvement within contract).** Before
  installing `wienerdog-dream` into the real config dir, the harness now
  compares it byte-for-byte against any skill already there
  (`skillDirsIdentical`/`listFilesRecursive` in `run-scenarios.js`). If
  identical (e.g. the maintainer already ran `wienerdog sync`), the harness
  leaves it untouched — no backup/restore, no risk of accidentally leaving a
  scenarios-still-in-progress marker if a maintainer edits the file mid-run.
  If different or absent, behavior is exactly the spec's backup/restore
  contract.
- **No new stub-based unit test for the three VERIFY-AT-FIRST-LIVE-RUN
  claims.** The spec's Deliverables table lists only `tests/unit/paths.test.js`
  as a new test file, and the spec explicitly states these three claims
  "cannot be unit-tested here; the harness itself is the test." Adding a new
  test file for them would both violate the Deliverables boundary (CLAUDE.md:
  "touch ONLY files listed") and duplicate fail-loud behavior that already
  exists unchanged in `run-scenarios.js` (non-zero `wienerdog dream` exit →
  `failures.push(...)`) and `rubric.js` (spawn/parse/auth error →
  `{pass:false}`, pre-existing, not added by this WP). Recording this instead
  of silently adding a file outside the boundary.
- **EXPENSIVE full run and the three live-verify claims were not exercised in
  this session** — see Verification output above for why (this session's
  guardrails forbid touching this machine's real `~/.claude` even for the
  harness's one deliberate touch). Flagging per the spec's requirement to
  state PASS/FAIL/not-run explicitly.

none

---
Generated-by: claude-sonnet-5

### DIFF
--- a/.github/workflows/scenarios.yml
+++ b/.github/workflows/scenarios.yml
@@ -1,12 +1,16 @@
 name: scenarios
 
-# Nightly, real-brain scenario harness (WP-015). Runs the actual `wienerdog
-# dream` pipeline against the real Claude model and Haiku grader, so it
-# spends real model quota. It must NEVER run per-PR — schedule + manual
-# dispatch only.
+# DORMANT under ADR-0009 (subscription-everywhere). The project's primary
+# scenario run is a local subscription routine on the maintainer's machine
+# (see tests/scenarios/README.md); this workflow is retained only as an
+# optional path for a future contributor who accepts running scenarios under
+# an ANTHROPIC_API_KEY secret in CI (GitHub Actions cannot do subscription
+# OAuth). It does nothing until such a contributor supplies the secret.
+#
+# Real-brain scenario harness (WP-015). Runs the actual `wienerdog dream`
+# pipeline against the real Claude model and Haiku grader, so it spends real
+# model quota. It must NEVER run per-PR — manual dispatch only, no schedule.
 on:
-  schedule:
-    - cron: "0 6 * * *"
   workflow_dispatch: {}
 
 jobs:
--- a/docs/specs/WP-023-scenario-subscription-auth.md
+++ b/docs/specs/WP-023-scenario-subscription-auth.md
@@ -1,7 +1,7 @@
 ---
 id: WP-023
 title: Rework the scenario harness to run on subscription auth (decouple fixture isolation from auth)
-status: Ready
+status: In-Review
 model: sonnet
 size: M
 depends_on: [WP-015, WP-020]
--- a/src/core/paths.js
+++ b/src/core/paths.js
@@ -26,7 +26,11 @@ const path = require('node:path');
 function getPaths(env = process.env) {
   const home = env.HOME || os.homedir();
   const core = env.WIENERDOG_HOME || path.join(home, '.wienerdog');
-  const claudeDir = env.CLAUDE_CONFIG_DIR || path.join(home, '.claude');
+  // WIENERDOG_CLAUDE_DIR (wienerdog-internal; the scenario harness sets it to a
+  // fixtures dir) takes precedence so transcript discovery can be redirected WITHOUT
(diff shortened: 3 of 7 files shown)


thegyula:
wd-reviewer verdict: **APPROVE** — 245/245 re-run; ADR-0009 compliance grep-proven (key deleted from every child env, workflow dispatch-only); WIENERDOG_CLAUDE_DIR precedence byte-exact with production zero-impact proven two ways; crash-safety of the real-config-dir touch simulated in a temp stand-in across all three preexisting states — byte-restored in every case. PR title fixed to the spec's DoD string. Live subscription run executed by the maintainer post-merge (next comment). Merging.

Generated-by: claude-fable-5

thegyula:
Maintainer live run (post-merge): **scenarios: PASS** on pure subscription auth — dream committed, injection gated out of Tier 3, all notes grounded, report complete, Haiku rubric green. All three verify-at-first-live-run claims are now VERIFIED: user-dir skill discovery under --setting-sources user ✓, subscription OAuth for the brain ✓, and for the grader ✓. Real skills dir byte-clean after both runs. One follow-up fix landed on main: the grader context now carries session ids/dates (first run falsely flagged a correctly-grounded recurrence timeline).

Generated-by: claude-fable-5

PR #24, opened 2026-07-03
feat(core): vault layout config layer + layout-aware digest (WP-022)

thegyula:
Spec: docs/specs/WP-022-vault-layout-layer.md

## Deliverables checklist

Deliverables: `src/core/layout.js` (create), `src/core/digest.js` (modify),
`src/cli/sync.js` (modify), `tests/unit/layout.test.js` (create),
`tests/fixtures/poweruser-vault/**` (create). The only additional touched file is
the spec itself (`status:` flip), required by the DoD.

## Verification output

```
$ npm test -- --test-name-pattern layout
ℹ tests 36
ℹ pass 36
ℹ fail 0

$ npm test -- --test-name-pattern digest
ℹ tests 35
ℹ pass 35
ℹ fail 0

$ npm test
ℹ tests 257
ℹ pass 257
ℹ fail 0
ℹ duration_ms 4610.084625

$ npm run lint
--- markdownlint ---
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
lint passed
```

The existing `tests/golden/digest-default.md` and `tests/unit/digest.test.js`
pass byte-unchanged (default-layout render is identical).

## Decisions made

- **`newestDaily` ignores `daily_filename` for reading**, per the spec: it walks
  `daily_dir` recursively and matches basenames `YYYY-MM-DD.md`, so flat and
  nested layouts share one code path. `daily_filename` is only used by
  `resolveDailyPath`/`layoutPromptLines` (write-path helpers WP-024 consumes).
- **`vault_layout:` line match** uses `^vault_layout:[ \t]*(#.*)?$` — exact
  top-level key, optional trailing comment, nothing else on the line — mirroring
  the "no leading whitespace" discipline of `dream/config.js readScalar`.
- **Quote/comment stripping** for block values reuses the exact rules from
  `readScalar` (inline ` #` dropped on unquoted values only; one layer of matching
  surrounding quotes removed), factored into a small `cleanValue` helper local to
  `layout.js` (did not extend `readScalar`, per spec).
- **Power-user persona** is "Priya Nair" (freelance climate researcher), distinct
  from the default golden's "Ada Kovács", so the non-default digest test cannot
  accidentally match the default golden.
- Added a small extra `readVaultLayout` test for a partial block (only
  `daily_dir` set) to lock in per-key defaulting; not required but cheap.

none

---
Generated-by: claude-opus-4-8

### DIFF
--- a/docs/specs/WP-022-vault-layout-layer.md
+++ b/docs/specs/WP-022-vault-layout-layer.md
@@ -1,7 +1,7 @@
 ---
 id: WP-022
 title: Vault layout config layer + layout-aware digest render
-status: Ready
+status: In-Review
 model: opus
 size: M
 depends_on: []
--- a/docs/specs/WP-022-vault-layout-layer.md
+++ b/docs/specs/WP-022-vault-layout-layer.md
@@ -214,6 +214,21 @@ module.exports = { defaultLayout, readVaultLayout, resolveDailyPath, layoutPromp
 
 Parser requirements for `readVaultLayout`:
 
+**Value validation (binding — the traversal-safety contract for WP-022/024/026):**
+every parsed value must be a safe vault-relative path. After `cleanValue`, a
+value is REJECTED — the built-in default for that key is used instead — when it
+is (a) empty, (b) absolute (`path.isAbsolute`, or a leading `/`), (c) contains
+a `..` segment when split on `/`, or (d) contains a backslash. Rejection is
+per-key and silent-safe: the rest of the block still applies. Rationale: these
+values are joined under the vault root by the digest render today and by the
+dream tier boundaries (WP-024) and adopt flow (WP-026) tomorrow — a traversal
+value would let a config edit exfiltrate out-of-vault file contents into the
+injected session digest (confused-deputy; reproduced in PR #24 review).
+Unit tests must cover: `../../../etc` → default; absolute path → default;
+`a/../b` → default; empty value → default; and an end-to-end proof that
+`renderDigest` with a hostile layout never reads outside the vault (plant a
+sentinel file outside; assert its content is absent from the digest).
+
 - Find the line that is exactly `vault_layout:` (no leading whitespace; a trailing
   comment after it is allowed). Then consume the following **indented** lines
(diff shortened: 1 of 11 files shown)


thegyula:
wd-reviewer verdict: **REQUEST-CHANGES** — implementation is spec-faithful (guardrail goldens byte-unchanged, 257/257, parser robust), but the mandated traversal probe confirmed: layout values with '..' segments or absolute paths are accepted silently, and the digest render then reads outside the vault (planted secret reproduced in the injected digest). Root cause is a spec gap in the parser requirements — the value-validation contract is being added to the spec on this branch; fix returns to the implementer. Also folding in: empty-after-clean values fall back to defaults.

Generated-by: claude-fable-5

thegyula:
Addressed the blocking finding in 6ae6d87: `readVaultLayout` now validates every parsed value after `cleanValue` and rejects (per-key, silent fallback to the built-in default) any value that is empty, absolute, contains a `..` segment when split on `/`, or contains a backslash — per the amended spec's binding value-validation contract. This also covers the non-blocking empty-after-clean finding.

New tests: `../../../etc` → default; absolute → default; `a/../b` → default; empty/`""`/backslash → default; and an end-to-end sentinel test (files planted outside a temp vault, hostile config through the `readVaultLayout` chokepoint, sentinel content asserted absent from the rendered digest).

Verification:

```
$ npm test -- --test-name-pattern layout
ℹ tests 40  ℹ pass 40  ℹ fail 0

$ npm test
ℹ tests 261  ℹ pass 261  ℹ fail 0

$ npm run lint
lint passed

$ git diff --stat origin/main -- tests/golden/
(empty — goldens byte-unchanged)
```

Generated-by: claude-opus-4-8

thegyula:
wd-reviewer verdict (re-review): **APPROVE** — blocking traversal finding fully resolved at the readVaultLayout chokepoint per the binding spec contract; original planted-secret probe now clean; fresh bypass probes (absolute, a/../b, backslash, quoted, trailing-.., URL-encoded, mixed-block per-key) all held; goldens byte-unchanged; 261/261. Merging.

Generated-by: claude-fable-5

PR #26, opened 2026-07-03
feat(cli): full vault adoption — wienerdog adopt (WP-026)

thegyula:
Spec: docs/specs/WP-026-full-adoption-flow.md

## Deliverables checklist

Files: `src/cli/adopt.js` (create), `src/core/layout-infer.js` (create),
`src/core/vault.js` (modify — `scaffoldMappedDirs`), `bin/wienerdog.js` (modify),
`tests/unit/layout-infer.test.js` (create),
`tests/fixtures/adopt/fake-brain-mapped.js` (create),
`tests/integration/adopt-e2e.test.js` (create). Plus the DoD-mandated
`status:` flip in the spec.

## Verification output

```
$ npm test -- --test-name-pattern layout-infer
# (both layout-infer unit tests pass)

$ npm test -- --test-name-pattern adopt-e2e
# adopt-e2e integration test passes

$ npm test
ℹ tests 275
ℹ pass 275
ℹ fail 0
ℹ duration_ms ~4744

$ npm run lint
--- markdownlint ---  Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---  frontmatter check passed: 4 spec(s), 4 agent(s)
lint passed
```

Additionally smoke-tested (not in the automated suite): TCC-guard refusal
(Documents flagged), `--dry-run` leaves config untouched, re-adoption refused
cleanly, and an empty non-git dir gets an initial commit.

## Decisions made

- **`--allow-empty` on the initial adopt snapshot.** An adopted dir with no
  committable content (empty, or all-ignored) would make `git commit` fail with
  "nothing to commit", aborting adoption. Adoption always needs a HEAD for the
  revert-safety guarantee, so the initial snapshot commit uses `--allow-empty`.
  Real mature vaults have content; this only hardens the empty edge case.
- **`--dry-run` and `--yes` are non-interactive.** With `--dry-run` (even
  without `--yes`) `adopt` never prompts — it reports whether git-init would be
  offered and prints the scaffold plan, then stops before any write. Simpler and
  matches "prints the plan and stops before any write."
- **TCC = macOS-only, no bespoke mount check** (per spec): `tccguard.guard`
  returns `ok:true` off darwin; "not TCC-protected" is v1's definition of an
  acceptable local path.
- **Realpath then existence check ordering:** resolve to absolute, verify the
  dir exists, then `realpathSync` (so a missing path yields a clean
  "not a directory" error rather than a realpath throw). The e2e compares the
  config `vault:` against the realpath'd form (macOS `/var`→`/private/var`).

none

---
Generated-by: claude-opus-4-8

### DIFF
--- a/bin/wienerdog.js
+++ b/bin/wienerdog.js
@@ -9,6 +9,7 @@ Usage: wienerdog <command> [options]
 
 Commands:
   init        Create the Wienerdog core (~/.wienerdog) and detect your AI tools
+  adopt       Use an existing vault in place as your Wienerdog vault
   sync        Re-render the session digest from your vault's identity notes
   dream       Consolidate recent sessions into vault memory (one commit)
   schedule    Add, remove, or list scheduled jobs (dream, routines)
@@ -37,6 +38,7 @@ async function main() {
   /** @type {Record<string, () => {run: (argv: string[]) => Promise<void>}>} */
   const commands = {
     init: () => require('../src/cli/init'),
+    adopt: () => require('../src/cli/adopt'),
     sync: () => require('../src/cli/sync'),
     dream: () => require('../src/cli/dream'),
     schedule: () => require('../src/cli/schedule'),
--- a/docs/specs/WP-026-full-adoption-flow.md
+++ b/docs/specs/WP-026-full-adoption-flow.md
@@ -1,7 +1,7 @@
 ---
 id: WP-026
 title: Full vault adoption — `wienerdog adopt` CLI, prerequisites, layout mapping
-status: Ready
+status: In-Review
 model: opus
 size: M
 depends_on: [WP-024, WP-025]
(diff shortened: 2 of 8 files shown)


thegyula:
wd-reviewer verdict: **REQUEST-CHANGES** — byte-level non-destructiveness proof spotless (rich fixture incl. unicode/space names: byte-identical through adopt AND uninstall; both stub-seeding branches correct; revert guarantee held; dry-run/decline/re-adopt all zero-write). One must-fix: the TCC guard fails open under a symlinked home (adopt.js realpaths the vault path but passes raw paths.home to tccguard — reproduced: adoption of ~/Documents succeeded with HOME under /tmp). Spec amended on-branch; fix returns to the implementer along with the in-scope inferLayout hygiene notes.

Generated-by: claude-fable-5

thegyula:
wd-reviewer verdict (re-review): **APPROVE** — original fail-open reproduction now refuses with zero writes; doubly-symlinked-home variant also held; regression test proven genuine (identical setup to the pre-fix repro); inferLayout explicit validation + whitespace round-trip closed; 277/277; boundary clean. Non-blocking architect note logged: export isSafeRelativePath from layout.js in the next layout-touching WP to retire the attributed duplicate. Merging — the vault adoption chain is complete.

Generated-by: claude-fable-5

PR #27, opened 2026-07-03
feat(install): defer vault creation until path is chosen (WP-027)

thegyula:
Spec: docs/specs/WP-027-defer-vault-creation.md

## Deliverables checklist

Plus one file outside the table (`tests/unit/codex-adapter.test.js`) — see Discovered issues.

## Verification output

```
$ npm test
ℹ tests 284
ℹ pass 284
ℹ fail 0

$ npm run lint
--- markdownlint ---   Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---  frontmatter check passed: 1 spec(s), 4 agent(s)
lint passed

# 1. Plain init defers the vault — no phantom ~/wienerdog.
OK: vault deferred
OK: no phantom vault
[ok] core directory exists (...)
[ok] install manifest parses
[ok] config.yaml exists and is non-empty
[warn] no memory vault yet — run /wienerdog-setup to create or choose one (this is normal right after install)
[ok] secrets directory permissions are 0700
[ok] AI tools — Claude Code: not found, Codex CLI: not found
doctor exit: 0
# sync on unset vault:
wienerdog: no vault set up yet — run /wienerdog-setup to create or choose your vault (or 'wienerdog init --fresh-vault' for the default).
sync exit: 1

# 2. --fresh-vault creates it; idempotent.
OK: vault set
commits: 1
wienerdog: already installed, nothing to do.
OK: doctor ok

# 3. adopt after a plain init leaves no phantom-vault notice.
now-unused lines: 0
```

## Decisions made

- **Reused `scaffoldVault` verbatim** — `--fresh-vault` records `vault-file`
  manifest entries and resyncs the config hash exactly like the old eager path.
  `src/core/vault.js`, `tests/unit/vault.test.js`, and
  `tests/golden/vault-default/` are byte-unchanged.
- **Fresh vault stays preserved on uninstall** — `vault-file`/`vault-dir` are
  intentionally skipped by uninstall (M7: "leaves only the vault"). No change
  to that; a user's memory is never deleted.
- **Escaped the backticks** in the `renderConfig` `vault:` comment
  (`` \`wienerdog adopt\` ``). The spec's literal string uses backticks, but
  that comment lives inside a JS template literal — unescaped backticks would
  terminate the string. Rendered config output is identical to the spec text.
- **`[fail]` vault-line regex** in doctor.test.js matches `/\[fail\].*vault/`
  (single-line) rather than anchoring the exact message, to stay robust to
  wording.

- **`tests/unit/codex-adapter.test.js` was NOT in the Deliverables table but
  its line-220 integration test broke** under the intended behavior change.
  The spec's Current-state note claimed the integration tests "keep passing,"
  but that test runs plain `init --yes`, drops identity files into the vault,
  then runs `sync` — which now throws on `vault: null`. I made the minimal
  one-line fix (switch its setup to `init --fresh-vault --yes`, matching the
  new reality) and flag it here as a spec gap. No behavior in the test's
  assertions changed. The gws-* tests that use plain `init` do not touch the
  vault and pass unchanged.

## WP-027 lessons

- WP-027: The `vault:` comment in init.js `renderConfig` sits inside a JS
  template literal — a spec-supplied string containing backticks must be
  escaped (`\``) or Node throws a SyntaxError at require time; the whole CLI
  crashed until escaped.
- WP-027: The spec's Current-state test census missed
  `tests/unit/codex-adapter.test.js`, which does `init` → `sync` end-to-end and
  depended on eager vault creation. When deferring an install step, grep for
  every test that runs the command and then a downstream command, not just the
  tests named in the spec.

---
Generated-by: claude-opus-4-8

https://claude.ai/code/session_014EA5BsHCfHRViJFXdovC86

### DIFF
--- a/docs/specs/ROADMAP.md
+++ b/docs/specs/ROADMAP.md
@@ -45,7 +45,7 @@ Milestone acceptance criteria are binding; WPs are the unit of implementation. S
 | [WP-024](done/WP-024-layout-aware-dream.md) | Layout-aware dream write path (validate tiers, brain prompt, skill) | M3 | opus | Done | WP-022 |
 | [WP-025](done/WP-025-guided-import.md) | Guided import from an existing vault (setup skill step 3) | M2 | sonnet | Done | WP-022 |
 | [WP-026](done/WP-026-full-adoption-flow.md) | Full vault adoption — `wienerdog adopt` CLI, prerequisites, layout mapping | M2/M3 | opus | Done | WP-024, WP-025 |
-| [WP-027](WP-027-defer-vault-creation.md) | Defer vault creation until the vault path is chosen (init `--fresh-vault`) | M2/M3 | opus | Ready | WP-026 |
+| [WP-027](WP-027-defer-vault-creation.md) | Defer vault creation until the vault path is chosen (init `--fresh-vault`) | M2/M3 | opus | In-Review | WP-026 |
 
 ## Dependency graph
 
--- a/docs/specs/WP-027-defer-vault-creation.md
+++ b/docs/specs/WP-027-defer-vault-creation.md
@@ -102,6 +102,9 @@ Tests that touch this area:
   `tests/integration/dream.test.js` — none assert the default vault exists after
   `init`; they build their own vaults or only check `config.yaml`. They keep
   passing.
+- `tests/unit/codex-adapter.test.js` — its integration test does
+  `init` → drop identity → `sync`, which now exits 1 on `vault: null`. Its init
+  call gains `--fresh-vault` (in Deliverables above). No assertions change.
 - `tests/unit/setup-skill-structure.test.js` — string-presence checks on the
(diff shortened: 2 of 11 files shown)


thegyula:
wd-reviewer verdict: **REQUEST-CHANGES** resolved by spec amendment (not code) — the implementer's codex-adapter.test.js fix was minimal and correct (only the init call gained --fresh-vault, zero assertions changed) but sat outside the Deliverables table, so the live CI boundary job rejected it. Root cause was a spec test-census gap; amended the table + census claim (the 'spec is the bug' path). Boundary check now exits 0. All else verified green: 284/284, three manual sequences exact (no phantom vault, doctor warn/exit-0, sync friendly exit-1, --fresh-vault 1-commit idempotent, adopt notice suppressed), goldens byte-unchanged. Merging.

Generated-by: claude-fable-5

PR #104, opened 2026-07-13
feat(doctor): flag a connected Google account with a missing client library (WP-103)

thegyula:
> **Stacked on #103 (WP-102..104 specs) — the docs/specs files in this diff land there. Soft-dep on WP-102 (remedy wording only); merge after the WP-102 PR.**

Spec: docs/specs/WP-103-doctor-gws-deps-probe.md

## Deliverables checklist

## Codex design-review deltas (rounds 1-4, applied on top of the initial implementation)

- **Load probe (Finding 1a).** The library check is now a containment-guarded LOAD probe — `deps.loadGoogleapis(paths)` in a try/catch — instead of the resolve-only `deps.isInstalled`, so a corrupt/partial install that resolves but fails to `require` warns instead of falsely reading `[ok]`.
- **Token validation (Finding 4 + round-2 Finding 3).** The probe JSON-parses the token (read-only) and requires `typeof refresh_token === 'string' && refresh_token.trim() !== ''`. Zero-byte / malformed / missing / wrong-type (`{"refresh_token":true}`) / whitespace-only tokens get a separate `[warn] Google sign-in file looks damaged — reconnect with /wienerdog-google-setup` and never `[ok]`. No token file at all still emits nothing (normal state).
- **Absent/broken warn split (round-2 Finding 2).** On load failure the probe splits on `deps.isInstalled`: **absent** (false) keeps the WP-102 self-heal promise ("the next `wienerdog gws` command will offer to install it, or run: <npm>"); **broken** (true — resolvable but unloadable) makes no offer claim, since WP-102's self-heal no-ops there.
- **Delete-then-reinstall remedy (round-4 Finding).** The broken warn prescribes "delete the folder <depsDir>, then reinstall it: <npm>" — a bare `npm install` can no-op over a corrupt-but-resolvable tree. Parity with WP-102's `loadGoogleapis` broken message.
- **No `gws auth` mention (Finding 3).** Both warn wordings dropped the `wienerdog gws auth` suggestion; a negative `!/gws auth/` assertion guards it.
- **New fixtures.** `plantCorruptDeps` (entry point throws on require) and `plantDamagedToken` with five damaged variants; the broken case asserts the deps-folder path and `!/will offer to install/`.
- **PR-gate fix (P2-A, spec rev 7d01980).** Quoted the deps dir in the shown repair command (`--prefix "${dir}"`) so a home path with spaces survives copy-paste; affects both warn branches via the shared `cmd`. Per the spec revision, no test assertion pins the `--prefix` portion — verified, no test change. Also fixed an MD038 markdownlint failure the merged spec revision introduced in the WP-102 spec (escaped backticks inside a code span → standard double-backtick span) so `npm run lint` is green.
- **Closing PR-gate (P2, spec rev 947d2dc).** New shape-broken test: `plantShapelessDeps` (`index.js` → `module.exports = {}` — requires cleanly but exports no `.google`) + `plantToken` → doctor `[warn] … broken (installed but not loadable)`, never `[ok]`, exit 0. No `doctor.js` change — the fix is WP-102's `loadGoogleapis` shape check, which the probe inherits. Since this branch still carries main's `deps.js` (no shape check), the case would falsely fail standalone; it probes `deps.loadGoogleapis` behaviorally against a planted shapeless module and self-skips until the WP-102 merge lands, then runs unconditionally (verified green with WP-102's `deps.js` swapped in locally).
- **Round-6 P2 (spec rev b715bf8, commit 32ca9f4).** The broken-vs-missing split is now keyed on PHYSICAL presence — `deps.depsPresent` (new WP-102 export) instead of `deps.isInstalled`, which reads false for a package.json-present-but-mainless tree and would mis-label that BROKEN state "missing". Ordering tolerance: `depsPresent` doesn't exist in this branch's `deps.js` (it lands with WP-102), so the probe uses a `typeof` guard falling back to `isInstalled` — dead code once WP-102 merges. New test: `plantMainlessDeps` (package.json, no `index.js`) + token → `[warn] … broken (installed but not loadable)`, never `/is missing/` — self-armed like the shape-broken case (skips until `deps.depsPresent` exists and reads a mainless tree as present; verified green with a spec-shaped `depsPresent` swapped in locally). Also aligned the WP-102 spec copy to wp/102's round-6 head 8499f26 (rewrap of a wrapped line markdownlint misread as a `+` list item) so `npm run lint` stays green.

## Verification output

```
$ npm test -- --test-name-pattern doctor
✔ doctor after a plain init warns about the deferred vault and exits 0
✔ doctor after init --fresh-vault reports the vault ready and exits 0
✔ doctor prints the npx update command when npx is on PATH (no network)
✔ doctor prints `wienerdog update` when npx is NOT on PATH (no network)
✔ doctor warns (exit 0) when a registered scheduler entry probes not-loaded
✔ doctor reports [ok] when a registered scheduler entry probes loaded
✔ doctor reports [ok] Codex skills registered when Codex is present and links intact
✔ doctor warns (exit 0) when a Codex skill link is removed
✔ doctor prints no Codex-skill line when Codex is not detected
✔ doctor prints no Google-readiness line when Google is not connected
✔ doctor warns (exit 0) on a damaged Google token: malformed JSON
✔ doctor warns (exit 0) on a damaged Google token: missing refresh_token
✔ doctor warns (exit 0) on a damaged Google token: wrong-type refresh_token
✔ doctor warns (exit 0) on a damaged Google token: whitespace-only refresh_token
✔ doctor warns (exit 0) on a damaged Google token: zero-byte file
✔ doctor warns (exit 0) when Google is connected but the client library is missing
✔ doctor warns (exit 0) when the client library is broken (resolves but fails to load)
✔ doctor reports [ok] when Google is connected and the client library is installed
✔ doctor with a set-but-missing vault fails and exits 1
✔ doctorSchedulerChecks: loaded → ok, missing → warn, unknown → omitted
ℹ tests 70  pass 70  fail 0

$ npm test
ℹ tests 776
ℹ pass 772
ℹ fail 0
ℹ cancelled 0
ℹ skipped 4
ℹ todo 0
ℹ duration_ms 29522.999583

$ npm run lint
--- markdownlint ---
Linting: 194 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed
```

## Decisions made

- `package-lock.json` was stale on the branch (still `0.7.1` after the `0.8.0` release commit bumped `package.json`); `npm install` refreshed the two version fields. Included since the lockfile is always-allowed per the Deliverables note.
- Merge of `origin/docs/wp-102-104-specs` (revised specs, rounds 1-4) into this branch: the WP-103 spec body auto-merged to the docs branch's revision with `status: In-Review`; the ROADMAP conflict was resolved keeping the docs branch's row text ("missing/broken") with `In-Review` status.

- none

---
Generated-by: claude-fable-5

https://claude.ai/code/session_018MozsEQZ3G3H66VjJ3K8z7

### DIFF
--- a/docs/specs/ROADMAP.md
+++ b/docs/specs/ROADMAP.md
@@ -917,6 +917,25 @@ Milestone acceptance criteria are binding; WPs are the unit of implementation. S
 > considered), cache-immune, symlink defense preserved. Containment tests (:205–230) stay
 > byte-identical; new in-process case (a4) FAILS on the old walk and passes on the
 > rewrite. WP-103's probe inherits it (no code change). WP-104/105 untouched.
+> **Closing Codex PR pass, round-6 (2026-07-13, two findings; deps.js + prompt.js +
+> doctor.js):** **P1 (high user-visibility, stdout hygiene)** — the self-heal wrote
+> its notice, consent prompt, and npm output to STDOUT, corrupting a connected
+> user's `gws … --json | jq` (and, with stdout piped + a TTY stdin, the prompt was
+> written into the pipe, invisible, while it waited). Routed ALL chatter to stderr:
+> notice → `process.stderr.write`; npm → `defaultRunInstall` `stdio: ['inherit', 2, 2]`;
+> prompt → a new backward-compatible `opts.output` on `confirm` that
+> `ensureGoogleapis` sets to stderr (grows WP-102 by `src/core/prompt.js` + test).
+> **P2 (classification gap the §0 rewrite opened)** — a package.json-present-but-
+> unresolvable (missing-main) tree made `resolveFromDeps` throw → `isInstalled`
+> false → classified ABSENT → self-heal `npm`-over-corrupt → arborist no-op → loop
+> again. Re-keyed the absent/broken split AND the self-heal gate onto **physical
+> presence** (`depsPresent`, the §0 existence check, now exported): every present-yet-
+> unusable state (resolve-throw/require-throw/shape-fail/symlink-out) classifies
+> broken; `ensureGoogleReady` gates on `depsPresent`; `ensureGoogleapis` fails to the
+> honest delete-then-reinstall remedy for a present-but-broken tree on the auth path
+> (owner no-auto-repair); WP-103's doctor swaps its broken-vs-missing key
+> `isInstalled` → `depsPresent`. New tests a5/h/i/j + a doctor missing-main case;
+> a2/a3/a4 + containment probes stay green. `index.js` unchanged. WP-104/105 untouched.
 
 <!-- -->
 
--- a/docs/specs/ROADMAP.md
+++ b/docs/specs/ROADMAP.md
@@ -120,7 +120,7 @@ Milestone acceptance criteria are binding; WPs are the unit of implementation. S
 | [WP-099](WP-099-install-ps1-git-url-validation.md) | Validate the Git-for-Windows asset URL is HTTPS on a GitHub host before download | M7 | sonnet | Done | — |
 | [WP-100](WP-100-codex-tool-output-and-fail-closed-roles.md) | Codex transcript parser — recognize the current tool-output item type (`custom_tool_call_output` + variants) and fail closed on unknown roles | M7 | sonnet | Done | — |
 | [WP-101](WP-101-gws-oauth-state-pkce.md) | gws OAuth loopback — add `state` + PKCE (RFC 8252) | M7 | sonnet | Done | — |
-| [WP-102](WP-102-gws-deps-self-heal.md) | gws read-path self-heal + disambiguated deps error (fix the post-upgrade dead-end) | M5/M7 | sonnet | Ready | — |
+| [WP-102](WP-102-gws-deps-self-heal.md) | gws read-path self-heal + disambiguated deps error (fix the post-upgrade dead-end) | M5/M7 | sonnet | In-Review | — |
 | [WP-103](WP-103-doctor-gws-deps-probe.md) | doctor probe — connected Google account with a missing/broken client library | M5/M7 | sonnet | In-Review | WP-102 |
 | [WP-104](WP-104-gws-drive-search-friendly-query.md) | gws drive search — friendly term search by default, `--raw` for Drive query language | M5 | sonnet | Ready | — |
 | [WP-105](WP-105-sync-interactive-gws-backfill.md) | sync interactive backfill of the on-demand googleapis install (headless-only users) | M5/M7 | sonnet | Draft | WP-102 |
--- a/docs/specs/ROADMAP.md
+++ b/docs/specs/ROADMAP.md
@@ -121,7 +121,7 @@ Milestone acceptance criteria are binding; WPs are the unit of implementation. S
 | [WP-100](WP-100-codex-tool-output-and-fail-closed-roles.md) | Codex transcript parser — recognize the current tool-output item type (`custom_tool_call_output` + variants) and fail closed on unknown roles | M7 | sonnet | Done | — |
 | [WP-101](WP-101-gws-oauth-state-pkce.md) | gws OAuth loopback — add `state` + PKCE (RFC 8252) | M7 | sonnet | Done | — |
 | [WP-102](WP-102-gws-deps-self-heal.md) | gws read-path self-heal + disambiguated deps error (fix the post-upgrade dead-end) | M5/M7 | sonnet | Ready | — |
-| [WP-103](WP-103-doctor-gws-deps-probe.md) | doctor probe — connected Google account with a missing client library | M5/M7 | sonnet | Ready | WP-102 |
+| [WP-103](WP-103-doctor-gws-deps-probe.md) | doctor probe — connected Google account with a missing client library | M5/M7 | sonnet | In-Review | WP-102 |
 | [WP-104](WP-104-gws-drive-search-friendly-query.md) | gws drive search — friendly term search by default, `--raw` for Drive query language | M5 | sonnet | Ready | — |
 
 > **First-production-night incident (2026-07-04).** WP-038, WP-039 and WP-041 form
--- a/docs/specs/ROADMAP.md
+++ b/docs/specs/ROADMAP.md
@@ -903,6 +903,20 @@ Milestone acceptance criteria are binding; WPs are the unit of implementation. S
 > `WienerdogError` not a `TypeError`; WP-103 adds an empty-module → `[warn]` case.
 > Shape check is minimal (presence of `.google`), not full API-surface validation —
 > accepted residual. WP-104/105 untouched.
+> **Containment-guard rewrite (2026-07-13, owner-approved — reverses the guard's
+> DO-NOT-CHANGE status):** the closing review found a second real P2 — `resolveFromDeps`
+> resolved the **bare** `googleapis` request (ancestor walk), so on a machine with an
+> ancestor/global copy + empty deps dir, `isInstalled()` resolved+rejected the ancestor
+> but Node cached that resolution in `Module._pathCache`; the consented self-heal then
+> installed into the deps dir and same-process `loadGoogleapis()` re-resolved to the
+> cached ancestor → threw "needs a one-time install" right after consent+`npm` succeeded
+> (first-run failure in exactly the env the guard exists for). Fixed by rewriting
+> `resolveFromDeps` to **direct-path construction** (gate on the deps-dir copy's own
+> `package.json`; resolve the absolute in-dir candidate, never the bare request; retain
+> the realpath containment check). Strictly stronger containment (ancestor copies never
+> considered), cache-immune, symlink defense preserved. Containment tests (:205–230) stay
+> byte-identical; new in-process case (a4) FAILS on the old walk and passes on the
+> rewrite. WP-103's probe inherits it (no code change). WP-104/105 untouched.
 
 <!-- -->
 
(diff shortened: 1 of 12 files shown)


thegyula (commented):
## Review — WP-103 doctor gws deps probe (spec: docs/specs/WP-103-doctor-gws-deps-probe.md)

**APPROVE**

Reviewed strictly against the spec. All acceptance criteria independently re-verified (checked out the branch, ran the suite myself — did not trust pasted output).

### Findings
_None blocking. One informational note below._

1. `[quality]` (informational, not blocking) `src/cli/doctor.js:1031` — `googleReadinessChecks` is invoked unconditionally in `run`, so `require('../gws/client')` + `require('../gws/deps')` now load on every `doctor` run even for users who never connected Google. This is exactly what the spec's Exact contract specifies (lazy-require inside the function, called once per run), the modules are cheap, and it matches the existing scheduler/Codex-check pattern — so it is correct as written. Noting only for completeness; no change requested.

### Acceptance criteria — each confirmed
- [x] Connected (token present) + `googleapis` resolves from `<core>/app/deps` → prints one `[ok] Google connected and its client library is installed` and exits 0. — verified by the new `[ok]` test, passes.
- [x] Connected + library not resolvable → one `[warn] Google is connected but its client library is missing …` with the remedy, still exits 0. — verified; `check('warn', …)` does not set exit 1, test asserts `r.status === 0`.
- [x] Not connected (no token) → no Google line. — verified; helper returns `[]`, test asserts stdout has neither `Google connected` nor `Google is connected but`.
- [x] No filesystem mutation. — helper only calls `fileExists`/`isInstalled`/`depsDir`/`GOOGLEAPIS_SPEC`, all read-only.
- [x] `npm test` and `npm run lint` pass. — re-ran both myself (see below).

### Review procedure results
1. **Boundary** — changed files: `docs/specs/{ROADMAP,WP-102,WP-103,WP-104}.md`, `package-lock.json`, `src/cli/doctor.js`, `tests/unit/doctor.test.js`. The three WP-10x specs + ROADMAP belong to the stacked docs PR #103; the WP-owned set is `src/cli/doctor.js` + `tests/unit/doctor.test.js`, both in the Deliverables table. `package-lock.json` is always-allowed and its diff is **only** the two stale version fields (`0.7.1` → `0.8.0`), nothing else. Clean.
2. **Acceptance criteria** — re-ran on the branch: `node --test tests/unit/doctor.test.js` → 13 tests, 13 pass, incl. all three new Google cases; full `npm test` → 770 tests, 766 pass, 0 fail, 4 skipped (matches PR body); `npm run lint` → passed (markdownlint 0 errors, frontmatter 3 spec(s)/4 agent(s)).
3. **Contract fidelity** — `googleReadinessChecks(paths)` matches the spec's Exact contract character-for-character: lazy requires, `fileExists(tokenPath(paths))` gate returning `[]`, the `[ok]` message `Google connected and its client library is installed`, the `cmd` = `npm install --ignore-scripts --prefix ${deps.depsDir(paths)} ${deps.GOOGLEAPIS_SPEC}`, and the warn message. Wire-in is the exact `for (const c of googleReadinessChecks(paths)) check(c.status, c.msg);` placed **after** the `codexSkillChecks` loop and **before** the update-notice block. Returns only `ok|warn`, never `fail`. Verified `fileExists` and `tempEnv().core` both exist on the branch as the contract assumes.
4. **CLAUDE.md conventions** — zero new runtime deps; JSDoc-only types (no TS); read-only so trivially idempotent/reversible (doctor mutates nothing); conventional-commit title `feat(doctor): … (WP-103)`; glossary terms respected. Spec `status:` flipped to `In-Review` in the WP-103 frontmatter and the ROADMAP row.
5. **Code quality** — messages are plain-language and actionable; the containment-guard-respecting `isInstalled` means a stray `googleapis` elsewhere won't read as installed; test fixtures plant under `<core>/app/deps` per the guard. No edge-case gaps.

No spec faults found — the spec was unambiguous and fully inlined the contract.

thegyula (commented):
## wd-reviewer — delta re-review of PR #104 (WP-103), head 11677f4

**APPROVE** (conditioned on merge order — see finding 1).

Re-reviewed the full current diff against the revised spec `docs/specs/WP-103-doctor-gws-deps-probe.md` as it appears on this branch (status `In-Review`, revision log through Codex round-4). Re-ran every verification command myself in an isolated worktree after `npm install`.

### Verification (re-run, not trusted from PR body)
- `npm test -- --test-name-pattern doctor` → tests 70, pass 70, fail 0. All 5 damaged-token variants + missing + broken + ok + not-connected present and green.
- `npm test` → tests 776, pass 772, fail 0, skipped 4 (pre-existing), 0 cancelled.
- `npm run lint` → markdownlint 0 errors, shellcheck clean, frontmatter passed. `lint passed`.

### 1. [boundary] — stacked spec diffs; safe only if #103 merges first
The diff touches `docs/specs/WP-102/103/104/105-*.md`. Only the WP-103 spec, `ROADMAP.md`, and `package-lock.json` are always-allowed; **WP-102/104/105 spec files are not in WP-103's Deliverables table**. They are present because merge commit c01c3bc pulled in `origin/docs/wp-102-104-specs`. Per the PR framing these diffs land via the stacked docs PR (#103). This is **not** an implementer defect, but it is a live merge-order condition: #103 must merge before #104 so #104's diff collapses to `src/cli/doctor.js` + `tests/unit/doctor.test.js` + `package-lock.json`. If #104 were merged first, it would carry four unrelated specs into main — do not merge #104 ahead of #103. Verified the merge preserved the revised WP-103 body: `status: In-Review` (line 4), round-4 revision log present (lines 395-402), load-probe contract intact; ROADMAP row reads "missing/broken … In-Review". `package-lock.json` delta is version-only (0.7.1 → 0.8.0), legitimate.
Non-docs code changes are exactly the two deliverables — no stray source/test files.

### 2. [contract] — messages byte-match the revised "Exact contracts" (all confirmed)
- Damaged: `Google sign-in file looks damaged — reconnect with /wienerdog-google-setup` (doctor.js:99) — one message for all 5 damaged variants, never `[ok]`.
- Token validation (doctor.js:96-100): valid JSON + `typeof refresh_token === 'string' && .trim() !== ''`; malformed/missing/wrong-type/whitespace/zero-byte → damaged warn; no token file → `return []` (emits nothing).
- LOAD probe via `deps.loadGoogleapis(paths)` in try/catch (doctor.js:107), not resolve-only.
- Absent→missing warn with the `wienerdog gws` self-heal offer + npm command (doctor.js:130-136); Broken→`broken (installed but not loadable) — delete the folder <depsDir>, then reinstall it: <npm>` with **no** offer claim (doctor.js:119-127). Split on `deps.isInstalled` (doctor.js:118).
- `ok`: `Google connected and its client library is installed` (doctor.js:110).
- No `gws auth` string anywhere. All paths return `status: 'ok'|'warn'`, never `'fail'` → exit 0 preserved.
- Wired after the `codexSkillChecks` loop and before the update-notice block (doctor.js:221-223), per contract §2.

### 3. [criteria] — tests match the revised acceptance set
`plantCorruptDeps` + `plantDamagedToken` fixtures present; the five damaged variants are table-driven; broken case asserts the delete-then-reinstall wording, `includes(<core>/app/deps)`, and `!/will offer to install/`; missing case asserts the offer regex and `!/gws auth/`; not-connected excludes `/Google sign-in file/` (and `/Google connected|Google is connected but/`); every warn case asserts `r.status === 0`.

### 4. [quality] — cross-cutting merge-order safety (deps.js): CONFIRMED SAFE, no breakage either way
On this branch `src/gws/deps.js` is main's version. WP-102's deps.js changes live on the #105 branch. The doctor probe does **not** depend on WP-102's deps.js:
- doctor **discards** the thrown error (`catch { usable = false; }`), so WP-102's reworded/state-aware `loadGoogleapis` message is irrelevant.
- The only contracts doctor relies on — `loadGoogleapis` throws on absent/broken and returns on working; `isInstalled` boolean; `depsDir`; `GOOGLEAPIS_SPEC` — all hold on main's deps.js. Verified: main's `loadGoogleapis` (deps.js:76-86) requires the resolved module **inside** its own try/catch, so a resolvable-but-corrupt entry point (`plantCorruptDeps`) throws → caught → re-throws → doctor sees `usable=false`; `resolveFromDeps` (deps.js:39-54) still returns non-null for the planted package → `isInstalled` true → broken branch fires.
- Therefore the broken/missing/ok tests pass against **main's** deps.js + this doctor.js. Whether #104 merges before or after #105, `loadGoogleapis` keeps the same throw-on-broken / return-on-working shape and `isInstalled` keeps returning true for a resolvable tree, so the doctor tests remain green in both merge orders. **No flag** — the coupling is only to the throw/return contract, which is stable across both versions.

### Conventions
Read-only helper (no mutation → idempotency/reversibility N/A), zero new runtime deps, JSDoc annotations (no TS), conventional commit subject. Clean.

No blocking findings on the implementation. Approve; merge #103 before #104.

thegyula:
**Disposition of Codex PR-review P1** (out-of-scope WP-102 spec edit on this branch): **rejected — deliberate stacked-merge hygiene.** The `docs/specs/WP-102-gws-deps-self-heal.md` bytes on this branch are intentionally identical to the wp/102 branch's final version, so that when #104 merges after #105 the file diff vanishes and no conflict occurs (per-branch divergent copies of the same spec cannot merge cleanly — verified by merge simulation). At merge time the boundary diff for this PR contains only the WP-103 deliverables. The P2 (module-shape false-[ok]) is being fixed via a WP-102 spec revision + branch patches.

PR #107, opened 2026-07-13
feat(gws): friendly drive search term + --raw escape hatch (WP-104)

thegyula:
Spec: docs/specs/WP-104-gws-drive-search-friendly-query.md

## Deliverables checklist

## Verification output

```
$ npm test -- --test-name-pattern "drive"
✔ run: drive search requires the positional <query> (1.069208ms)
✔ run: drive search wraps a bare term as a full-text search by default (0.138584ms)
✔ run: drive search --raw passes the query to Drive verbatim (0.08ms)
✔ run: drive search escapes a single quote in the term (0.081708ms)
✔ run: drive read requires --id (0.112625ms)
✔ run: drive read parses the --id token (0.130917ms)
ℹ tests 59
ℹ pass 59
ℹ fail 0

$ npm test
ℹ tests 800
ℹ pass 796
ℹ fail 0
ℹ cancelled 0
ℹ skipped 4
ℹ todo 0

$ npm run lint
--- markdownlint ---
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
lint passed
```

## Decisions made

None — implemented exactly per the spec's exact contracts (buildDriveQuery, --raw parsing in run(), escaping order). buildDriveQuery is not added to module.exports since no test or contract calls it directly; it's only exercised indirectly via drive.run(), keeping the diff minimal.

None.

---
Generated-by: claude-sonnet-5

### DIFF
--- a/docs/specs/WP-104-gws-drive-search-friendly-query.md
+++ b/docs/specs/WP-104-gws-drive-search-friendly-query.md
@@ -1,7 +1,7 @@
 ---
 id: WP-104
 title: gws drive search — friendly term search by default, --raw for Drive query language
-status: Ready
+status: In-Review
 model: sonnet
 size: S
 depends_on: []
--- a/src/gws/drive.js
+++ b/src/gws/drive.js
@@ -71,6 +71,20 @@ async function read(services, opts) {
 }
 
 /**
+ * Build the Drive `q` from a user argument. By default a plain term is wrapped
+ * as a full-text search; --raw passes the argument as literal Drive query
+ * language. Drive string literals are single-quoted with `\` and `'` escaped.
+ * @param {string} term
+ * @param {{raw?:boolean}} [opts]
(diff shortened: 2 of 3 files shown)


thegyula:
Codex review follow-up: backslash-escaping test gap closed in 0271cba (new case asserts the exact escaped q string for a term containing a backslash).

Two accepted Codex nits, on record as known limitations (both follow the spec's exact contract; not fixed here):

1. A search term literally equal to `--raw` cannot be expressed — `rest.includes('--raw')` treats it as the flag. Workaround: use raw mode, e.g. `gws drive search --raw "fullText contains '--raw'"`.
2. Extra tokens after the term are silently ignored — only the first non-`--raw` token is taken, so a raw Drive query must be passed as one shell argument (quoted).

PR #121, opened 2026-07-28
docs(specs): post-merge errata — V-20 recomputed around the owner's reaffirmation, ADR-0034 vault re-measurement

thegyula:
Docs-only, architect-side. No spec implements this — it is the post-merge errata pass on the secret-fence epic after PR #120 (leg 1) landed as `72f3e46`.

## What is in here

Two commits, deliberately separate.

**`bd631b7` — the owner's line, carried, not authored.** He filled the leg-2 reaffirmation slot in `docs/specs/WP-secret-fence-ep2-redact-arm.md` on 2026-07-27 with `OWNER-REAFFIRMED 2026-07-27` (trailing space included). That commit carries his bytes across and touches nothing else. **No agent edits, rewords, re-indents, re-dates or reflows an owner line, ever (row S5)** — including to make it match the template the slot comment offered, which is longer than what he typed.

**`0abe841` — the machinery around it.** Four items below.

## 1. V-20 recomputed, and disclosed at the step

The owner's line sits inside the region V-20 checksums, so V-20 went red. This is the one digest in the file that MAY be recomputed, and its own failure text says so: it is agent-computed over prose the architect owns, unlike V-11's (the owner's signature bytes) and V-18's (a ratified review criterion).

```
was              ec53884c8fc5be494718b6ad7ae2b8439e10b2e00a2e2afd262d6d36933197cd
+owner line only a6add80ae5be6291e5408fe77d2a519c29f5d72e04b2853a74737afd6683c711
now              e3fe067fe102ba200ff729751d36f620e67e4614ee5e3dd37b22951326728b10
```

The intermediate value is recorded in the V-20 comment so the owner's edit stays separable from the architect's. What else moved inside the region: the slot comment stops calling itself empty and records who filled it and where the disclosure lives; and leg 1's path in the Provenance narrative is repaired now that its spec has moved to `done/`.

**V-11 and V-18 verified byte-untouched**, not assumed: one `OWNER-SIGNED` line in each of the ADR and the spec, one inside `## OWNER-APPROVED`, one in the ADR's header region, pinned signature digest `4ca8fe05…` unmoved, and the `OWNER-REAFFIRMED` form scores **0** against V-11's pattern. V-18 is `77a67f3f…` and byte-identical in both legs.

## 2. ADR-0034 errata — 2026-07-28, rows ER-7 / ER-8 / ER-9

Re-measured against the live vault with `scripts/measure-secret-fp.js` at `72f3e46`:

```
notes scanned                                     189
notes with ANY finding (EP2 reverts today)        109   (57.7%)
    high-entropy ONLY                             108
    a labelled rule ONLY                            0
    both                                            1
findings by rule:  high-entropy 315 occurrences  |  aws-key 1
distinct high-entropy runs                        110

notes WITHHELD (a quarantine finding)               1
notes SCRUBBED in place (redact only)               9
notes untouched (no finding)                      179
notes REVERTED by EP2, after this leg              10
```

E2 re-derived over the same corpus: 110 distinct runs, **106** containing a `/`, longest slash-free segment still **17** against the 24-character floor, distribution `7→3, 8→9, 9→25, 10→28, 11→3, 12→29, 13→4, 14→4, 17→1`.

E1, E2 and E3 are corrected **in place** and the pre-edit bytes are quoted verbatim in the errata block, using the two-space indent quoting device the round-5 structural correction already established — load-bearing here, because V-15 digests the E1 block by a content-addressed `sed` range and an unindented quotation would silently extend it.

**Every figure moved; no structural fact and no conclusion did.** One labelled-rule finding in the whole corpus, still `aws-key`, still the documented `AKIA…` placeholder, so E1's headline claim and its commentary paragraph are byte-unchanged. E3 is now measured against the **shipped, merged detector** rather than the from-spec prototype it rested on — stronger evidence, not a different claim.

**Owner's ruling, transcribed, not a signature.** The owner was asked whether this errata block needs his signature and answered that it does not and that he passes on signing it. Recorded in the block as a dated `OWNER-RULED IN SESSION — 2026-07-28 (TRANSCRIBED BY THE ARCHITECT, NOT OWNER-TYPED)` paragraph that says in terms that no gate may key on it. **Checked against the licence before relying on it**, per instruction: the errata mechanism covers *measured evidence only*; Decisions 1–7 and the owner-signature line are untouched (verified — the `## Decision` digest is `494aa6ae…` and the five-condition precedent digest `57f298f2…`, both byte-unchanged); the five-condition precedent is not invoked because nothing inside a Decision is edited. **This is errata-shaped. No Decision moved.**

Leg 2's Table D rows D1/D2 and V-17's seven literals are re-derived **from** the amended ADR, never the reverse.

## 3. The two advisory findings from round 1 of leg 1's merge review

**(a) C3-7 vs C3-2.** C3-2 needs a per-cell `today` for all 156 cells; C3-7 said "transcribed constants" and this table transcribes six. C3-7 now states both sources: the 18 cells C3-3 names take the six transcribed constants, and the other 138 are computed by `caughtByShippedDetector` in `tests/fixtures/secret-corpus.js` — the pre-WP detector re-implemented as the module's labelled findings less `high-entropy` and less the `basic-auth` rule A16 adds, OR a hand-written copy of the shipped context-free pass. **What pins it, and how far**: the six constants are asserted against that function's own output to within 0.1 point across all 18 C3-3 cells, and three of them are fractional (38.1, 91.3, 96.5), so a drifted entropy half cannot satisfy C3-3. For the other 138, C3-2's set equality pins it **in one direction only** — over-reporting `today` adds a 19th regressing cell and fails; under-reporting is not caught. That asymmetry is stated rather than glossed, and it is fail-safe in the direction that matters. The fixture module's Deliverables cell now names the emulator too, since that table is the permission boundary. **No code change asked for or made.**

**(b) A6.** Decision: **tighten the wording**, and additionally state the behaviour as a residual. Reasoning: the shipped code is byte-identical to this document's own Exact-contracts block, so the normative form already says "maximal" through `String.replace` semantics — only A6's prose was wide. Left as "any sub-run", an implementer reading A6 alone builds a sliding-window scan, which is a real behaviour change with an unmeasured false-positive cost pointing straight at the class E2 exists to protect (path-shaped runs are high-entropy *across* their segments; a window scan starts looking inside them). But wording alone would wallpaper a genuine gap, so it is also **residual 8**, measured:

```
'blob ' + 'A'.repeat(40) + 'q7PmXz4KvR9tWc2LbN8dYfGh' + ' end'
  maximal 64-char narrow run : 2.674 bits/char
  the 24-char window inside  : 4.585 bits/char
  scanAndRedact              : [] , text unchanged
'blob q7PmXz4KvR9tWc2LbN8dYfGh end'
  scanAndRedact              : [{high-entropy, redact, count 1}]
```

**Not a regression** — the pre-WP detector matched the same maximal run and tested the same 2.674 bits, so it was silent too (verified through the fixture module's own `shippedEntropyFires`). And padding a secret with a long low-entropy prefix from the same alphabet is a deliberately shaped evasion, which ADR-0034 Decision 4 puts out of scope by definition. AC-4 updated as a registered mirror. **No code change asked for or made.**

## 4. Leg 1 filed, leg 2 still Ready

`WP-secret-fence-two-tier-detector` → `status: Done`, moved to `docs/specs/done/`. Its six self-referential paths repointed; leg 2's id-allocation block repointed with a guard that fails loudly if either leg's path stops resolving (a missing file makes `mdefs`/`vdefs` return empty and every cross-leg check pass vacuously).

**M1/M2/M5 are frozen, deliberately, and V-15's five literals are retired rather than recomputed.** A completed WP is a dated record of what was built and the evidence it was accepted against; retro-fitting a corpus measured two days after it merged would make it describe evidence that did not exist at review. The live mirror obligation moves to leg 2's D1/D2 — derived rows citing ADR-0034, re-derived here, pinned on both sides by V-17 — so exactly one live mirror exists and it agrees with the canonical surface. A loud block at the head of that section says all of this and instructs nobody to "repair" the retired literals; the errata block names all five with the same disposition.

`WP-secret-fence-ep2-redact-arm` stays **Ready**. It dispatches next.

## Verification output

**V-16, leg 2's dependency gate — now PASSES where it failed before leg 1 was on disk:**

```
V-0 ok: an unstaged edit to a forbidden file IS visible to changed_paths
V-16 P5 ok: the redact producer cannot span a newline
V-16 P6 ok: Authorization: Basic is quarantine and its body is fully redacted
EXIT=0
```

(P1–P4 and P5a assert through silent `must_eq`/`must_not` helpers; exit 0 is the whole of Table P holding.)

**Digests, after:**

```
V-11 ADR whole file          : 1   (want 1)
V-11 SPEC whole file         : 1   (want 1)
V-11 SPEC in OWNER-APPROVED  : 1   (want 1)
V-11 ADR header region       : 1   (want 1)
V-11 digest got : 4ca8fe05aded57ae5bdb3113292793c8498bf0490a76a0970ef008704ec741b6
V-11 digest want: 4ca8fe05aded57ae5bdb3113292793c8498bf0490a76a0970ef008704ec741b6
OWNER-REAFFIRMED scores against V-11 pattern: 0   (want 0)
owner line still on disk, byte-exact: 1   (want 1)

V-18 leg2: 77a67f3f2d52e27ed54c1ce7ec0bc29a03280147aab0ef2813fa3f3d62503871
V-18 leg1: 77a67f3f2d52e27ed54c1ce7ec0bc29a03280147aab0ef2813fa3f3d62503871
V-18 want: 77a67f3f2d52e27ed54c1ce7ec0bc29a03280147aab0ef2813fa3f3d62503871

V-33 leg2: 4e5fc85946d6c9b7fe2cd0440fd86795f2e3d0b8b255b2664e121dee3deceeb6
V-33 leg1: 4e5fc85946d6c9b7fe2cd0440fd86795f2e3d0b8b255b2664e121dee3deceeb6
V-33 want: 4e5fc85946d6c9b7fe2cd0440fd86795f2e3d0b8b255b2664e121dee3deceeb6

V-20 got : e3fe067fe102ba200ff729751d36f620e67e4614ee5e3dd37b22951326728b10
V-20 want: e3fe067fe102ba200ff729751d36f620e67e4614ee5e3dd37b22951326728b10
V-20 former-name needle: present
V-20 OWNER-ANSWERED needle: present

V-17, both sides:
ADR notes scanned 189: ok
ADR 109 (57.7%): ok
ADR withheld 109 -> 1: ok
ADR scrubbed 0 -> 9: ok
ADR untouched 80 -> 179: ok
D1 occurrences: 1   D2: 1
D1 canonical whole-line: 1
D2 canonical whole-line: 1

V-21, must be unchanged:
DEC : 494aa6aed5203f84d0144c9ef166afb01f30c0e8be22f8f2bca495d5963c7870   want 494aa6ae…
PREC: 57f298f2b9015d16680a3c0d8dd1943164c9019d43449ad9918569c201da81e8   want 57f298f2…
PREC heading count: 1   (want 1)

ADR m1 sed-range line count: 7   (want 7 — the indented quotation did not extend it)
ER rows now: 9
```

**Shellcheck, all four extracted bash blocks, `--severity=warning`:**

```
--- shellcheck /tmp/sc-WP-secret-fence-two-tier-detector-1.sh --- clean
--- shellcheck /tmp/sc-WP-secret-fence-two-tier-detector-2.sh --- clean
--- shellcheck /tmp/sc-WP-secret-fence-ep2-redact-arm-1.sh --- clean
--- shellcheck /tmp/sc-WP-secret-fence-ep2-redact-arm-2.sh --- clean
EXIT=0
```

**Lint and tests:**

```
--- markdownlint ---   Linting: 376 file(s)   Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---   passed: 202 spec(s), 4 agent(s)
lint passed

ℹ tests 1751
ℹ pass 1746
ℹ fail 0
ℹ skipped 5
```

## Decisions made

- **The re-measurement is filed on the ADR and NOT back-fitted into the Done leg-1 spec.** Argued above and in both documents. The alternative — rewriting M1/M2/M5 — would have needed re-publishing vault-derived path fragments for M2's frequency list and would have made a merged WP describe evidence measured after it merged.
- **Leg 1's V-15 literals are retired, not recomputed.** Recomputing a digest in a verification block that can no longer run asserts an equality nobody will ever check. Both the spec and the errata block say so and forbid the "repair".
- **A6: tightened AND stated as a residual**, rather than accepted as residual only. Reasoning above.
- **The errata quotes the pre-edit E1 block indented two spaces.** Verbatim quoting at column 0 would extend V-15's `sed` range across the whole errata block. Called out in the text so it does not read as sloppy quoting.
- **ADR-0034's 2026-07-26 cross-reference-update row is left pointing at leg 1's original path.** It is a dated record of where the file was created; the forward pointer to `docs/specs/done/` is in the new dated block instead.
- **V-17's first two literals stay anchored at line start.** Now load-bearing: the errata block contains the superseded figures verbatim, and an unanchored grep would match the historical record and pass on an uncorrected document. Noted in the step.

- `docs/specs/done/WP-secret-fence-two-tier-detector.md` still carries `56%` in its title and frontmatter, and one Implementation-notes sentence still reads "9 of the maintainer's 182 vault notes". Both are inside the frozen record and are covered by the supersession block; not edited, for the same reason M1 was not.
- `scripts/measure-secret-fp.js` cannot reproduce E1 on a corpus where the `basic-auth` rule leg 1 added has matched — by design, it refuses to print anything. That is correct today (zero matches on this vault) but means E1 becomes unreproducible the first time a real `Authorization: Basic` header lands in a note. A shipped-rule-only labelled stage would close it; the script's own comment already says why it was not built.

---
Generated-by: claude-opus-5


thegyula:
**`b0fb3c7` — two overclaims of my own, corrected before review rather than after.** Both are about the strength of a claim written in `0abe841`, not about a figure; no figure moved and every digest re-verified byte-unchanged.

1. The supersession block in the leg-1 spec said *"every figure in M1, M2 and M5 moved"*. M2's frequency list was **not** re-measured — re-deriving it would mean re-publishing vault-derived path fragments, and it is the one part of these measurements ADR-0034 never mirrored. Scoped now to the figures M1/M2/M5 share with the ADR, with the omission stated rather than implied.

2. ER-8 raised E2's slashed-run count 102 → 106 while leaving E2's qualitative sentence — *"every one of those is a file path or a slash-joined prose list"* — covering four runs nobody had looked at. Re-checked over the whole new set, and **structurally, so that nothing is published**:

```
slashed runs                                : 106
  all slash-free segments <= 17 chars       : 106
  all segments are word/identifier-shaped   : 106
  every segment has a vowel (or <=3 chars)  : 105
  any run containing + or = (base64 marker) :   0
```

The claim holds. The check is recorded in ER-8 itself, so the next reader sees that it was verified rather than inherited.

CI settled all-pass on `0abe841` (boundary, pr-title, lint, smoke ×2, test ×2) before this commit; it re-runs on `b0fb3c7`.

thegyula:
**`85b360f`** — third overclaim of the same shape, and the last. The paragraph under Table D said *"V-17 went red both times and that is how both were found"*. True of 2026-07-26, false of 2026-07-28: this pass re-measured deliberately after leg 1 merged and re-derived D1/D2 in the same commit as the ADR errata, so V-17 was never observed red for it. The paragraph now distinguishes the two, because "the gate caught it" and "the architect went looking" are different claims about how much the machinery is worth. No figure moved; every digest re-verified byte-unchanged and lint passes.

thegyula:
**`af9ef5f`** — one wrong date in the errata block's opening paragraph: leg 1 merged on **2026-07-28**, not 2026-07-26. `git log -1 --format=%ad 72f3e46` is 2026-07-28; 2026-07-26 is the date the two leg specs were *added* (`7ef4c51`), which is a different event this ADR already records elsewhere. No figure, no digest and no other sentence is affected. Lint passes.

thegyula:
## Fix round — `1a079a2`

All eight findings applied. Per-finding disposition below, then the executed evidence. Two further defects were caught by this repository's own machinery during the round and are recorded at the end.

### 1 [HIGH — Codex] V-21 carried into leg 2 as V-34 — FIXED

Accepted in full. The two digests are the only thing in the repo hashing ADR-0034's owner-ratified Decisions 1–7 or its architect-only amendment licence; archiving leg 1 would have left an unauthorized edit to either passing the entire active suite.

**Both literals carried UNCHANGED — neither recomputed.** New leg-2-owned id **V-34**, placed in the preflight beside V-16, because both gate on something this WP depends on and does not own.

```
V-34 digest 1 (## Decision … ## The measured evidence)
  494aa6aed5203f84d0144c9ef166afb01f30c0e8be22f8f2bca495d5963c7870   ← identical to leg 1's V-21
V-34 digest 2 (the five-condition amendment licence, next-## terminated)
  57f298f2b9015d16680a3c0d8dd1943164c9019d43449ad9918569c201da81e8   ← identical to leg 1's V-21
```

A new id rather than reusing V-21 because the id convention says an id means one check: a check that changes owner changes id. The "V-21 is unused in this leg" paragraph is left standing and extended rather than rewritten.

**Mutation-tested on five edits of a COPY of the ADR** (nothing committed):

| probe | result |
|---|---|
| reword a Decision heading | digest 1 moves `494aa6ae…` → `04d3c2d0…` — **discriminates** |
| delete `and by nobody else` from licence condition 5 | digest 2 moves → `c399f04c…` — **discriminates** |
| demote the licence's terminator heading out of `##` form | heading count stays 1, range grows 105 → 131 lines, **digest 2 moves → `84c55805…`** — discriminates, but via the digest, not the count |
| rename the licence heading | range extracts EMPTY, empty-range guard fires — **discriminates** |
| rename the Decision terminator | digest 1 moves → `24368bae…` — **discriminates** |

No mutation left both digests green. Probe 3 is the interesting one and is now written into V-34's comment: **the heading count catches a range that absorbed another section's heading, and does NOT catch one that swelled because the next heading was demoted** — the digest is the gate, the count is the secondary guard. Stated so nobody over-reads what the count does.

Leg 2's two stale "leg 1's own V-15 greps the same lines, which is what forces the two legs to agree" claims (Derived-measurements paragraph and the D1/D2 Checklist bullet) are **deleted, not softened**, and replaced with the true reason V-17 suffices alone: it asserts both sides.

### 2 [HIGH — wd-reviewer] V-15's failure text instructed the forbidden repair — FIXED

The step now opens with a `RETIRED 2026-07-28` header stating that a red V-15 is expected on a Done spec, that the banner below it ("the architect recomputes these digests in the same pass") was written for a `Ready` spec and is superseded, and that the deciding surface is Table F row F7. **All five `FAIL V-15:` paths** — `M1_EXPECT`, `M5_EXPECT`, `E3_EXPECT`, `ER_EXPECT` and `SWEEP_EXPECT` — echo one line of it, so nobody meets any of them cold. (The reviewer named four; F7 lists five, and the sweep path gets one too.)

**The standing rule is scoped, not deleted.** "A gate with a documented exemption is a gate nobody runs" is now explicitly about a LIVE spec — a check an implementer is told to skip while work is in progress is decoration, and that is still true. Retirement is a different act: once, at the Done transition, by the architect, and only when the guarantee has a live successor or needs none. Leaving both sentences unreconciled inside one step was the defect.

### 3 [MEDIUM — Codex] C3-7's directionality — FIXED, and the gap dispositioned

Correct. The test flags a cell when `proposed < today`, so:

- **UNDER**-reporting the emulated baseline conceals a real regression → **fail-OPEN**, uncaught.
- **OVER**-reporting adds a spurious regressing cell, breaking the set equality → **fail-closed**, loud.

My text had the mechanics right and the conclusion inverted — it called the uncaught direction the safe one. Corrected, with the reviewer's worked example written in (true 100, proposed 90 is a regression; an emulated baseline of 80 reports `90 ≥ 80` and stays green).

**Dispositioned as residual 9** rather than left as a corrected sentence. Why it is narrow today, measured not argued: the emulator's **labelled half cannot under-report at all** — it reads the live module's own findings minus `high-entropy` and minus the one rule this WP adds, and A10 changed labelled rules' *severity*, never their patterns; its rot direction (a future WP adding a rule without extending the exclusion) makes it over-report, which fails closed. The **entropy half is the exposure**: a hand-written `[A-Za-z0-9+/=]{24,}` @ ≥3.5, pinned by C3-3's six constants across 18 cells *only to the extent those six fixtures exercise it*. **Closed by** an independent pre-WP oracle — which is exactly what `WP-secret-scan-baseline-oracle` was, and residual 9 records that "Why no dependency on the oracle WP" was right about the 18 cells and wrong about the other 138.

### 4 [MEDIUM — wd-reviewer] the stale 56% ahead of the banner — FIXED (widened, plus a pointer)

Both. The banner now states that **every number in the file, the H1 title and frontmatter `title` included, is a 2026-07-26 measurement**, and the Context paragraph carries a one-line superseded pointer naming its Table F row.

### 5, 6, 7 [LOW] — all FIXED

- **`:906`** — the licence-compliance sentence now reads "…a sentence that quotes one, **or the evidence-provenance paragraph ER-7(a) names**", and says in terms that this was a scoping defect in the sentence, not an undisclosed edit. (Noted: fourth instance of the overclaim shape.)
- **`:981` ER-7(b)** — "…current output verbatim **apart from the pre-existing parenthetical on the `both` row, `(a documented AKIA placeholder)`, which is editorial, predates this pass and is re-attached with the same three-space gap it always had**" — the script prints counts only and never printed it.
- **`:493`** — dispositioned in the 2026-07-28 block, not repaired in place: the 2026-07-26 row records where a file was *created*, and editing it would make a dated block assert something untrue of its date. The forward pointer to `docs/specs/done/…` is the new paragraph. Explicitly noted that the five-condition precedent does not gate the choice, since the block is appendix bookkeeping and not inside a Decision.

### 8 [INFO] — recorded, not fixed

`scripts/measure-secret-fp.js:4` cites the old leg-1 path. Out of boundary for this pass; listed under Discovered issues.

### The family observation — cheap extraction done now

Findings 2 and 4 are one family: *the surfaces that ANNOUNCE the freeze fell out of agreement with the surfaces that are frozen.* Rather than wait for a third, **Table F** now declares what is frozen in this Done spec — one row per surface with its live successor — and the banner, V-15's retirement header and ADR-0034's digest-disposition list all **cite it instead of re-enumerating it**.

| # | frozen surface | live successor |
|---|---|---|
| F1 | H1 title + frontmatter `title` ("56%") | ADR-0034 E1 / ER-7 |
| F2 | Context, "102 of 182 (56.0%)" | ADR-0034 E1 / ER-7 |
| F3 | M1 | ADR-0034 E1 / ER-7 |
| F4 | M2 (counts; **frequency list has no successor and was not re-measured**) | ADR-0034 E2 / ER-8 |
| F5 | M5 | ADR-0034 E3 / ER-9 |
| F6 | the A15 "9 of … 182 vault notes" sentence | none — illustrative, never mirrored |
| F7 | V-15's five literals | leg 2's D1/D2 + V-17 |
| F8 | **V-21's two ADR digests — NOT frozen, CARRIED** | leg 2's **V-34** |

**F8 is the row that earns the table.** It encodes the rule the whole fix round turns on: a check retires only when something else carries its guarantee — never merely because its spec reached `Done`. Also disclosed: Table F sits inside a blockquote, so V-32's registration schema does not enumerate it; not a hole today (F1–F8 are not ids of any tracked family), and the repair if that changes is to lift it out, not to widen the schema.

### Two defects this round's own checks found, which no review leg caught

- **The id-allocation step rejected my `# V-15 RETIRED …` header** as a *second definition* of a live id — Table H row H3 defines a verification by its `# V-<n> ` header. Reworded to `# RETIRED 2026-07-28 — THE STEP BELOW`, with the reason written in so it is not re-introduced.
- **leg 2's V-30 refused V-34 until registered**, and **leg 1's V-32 refused residual 9 *and residual 8*** — residual 8 was added in the previous round without registration and no review leg caught it. Both registered with their mirror sets. Both steps green.

## Executed evidence

```
V-11 ADR whole file          : 1   (want 1)
V-11 SPEC whole file         : 1   (want 1)
V-11 SPEC in OWNER-APPROVED  : 1   (want 1)
V-11 ADR header region       : 1   (want 1)
V-11 digest got : 4ca8fe05aded57ae5bdb3113292793c8498bf0490a76a0970ef008704ec741b6
V-11 digest want: 4ca8fe05aded57ae5bdb3113292793c8498bf0490a76a0970ef008704ec741b6
OWNER-REAFFIRMED scores against V-11 pattern: 0   (want 0)
owner line still on disk, byte-exact: 1   (want 1)

V-18 leg2/leg1/want : 77a67f3f… / 77a67f3f… / 77a67f3f2d52e27ed54c1ce7ec0bc29a03280147aab0ef2813fa3f3d62503871
V-33 leg2/leg1/want : 4e5fc859… / 4e5fc859… / 4e5fc85946d6c9b7fe2cd0440fd86795f2e3d0b8b255b2664e121dee3deceeb6
V-20 got/want       : e3fe067fe102ba200ff729751d36f620e67e4614ee5e3dd37b22951326728b10 (unchanged this round)
V-21 DEC            : 494aa6aed5203f84d0144c9ef166afb01f30c0e8be22f8f2bca495d5963c7870  (unchanged — carried)
V-21 PREC           : 57f298f2b9015d16680a3c0d8dd1943164c9019d43449ad9918569c201da81e8  (unchanged — carried)
V-29 A8             : 908892b2… == A8_EXPECT
V-29 M4e            : f5b90e8c… == M4E_EXPECT

V-17 both sides: five ADR literals ok; D1/D2 occurrences 1 each; both canonical rows match whole-line

V-16 + V-34 preflight:
  V-0 ok: an unstaged edit to a forbidden file IS visible to changed_paths
  V-16 P5 ok: the redact producer cannot span a newline
  V-16 P6 ok: Authorization: Basic is quarantine and its body is fully redacted
  V-34 ok: ADR-0034's Decisions and its amendment licence are unedited
  EXIT=0

V-30 ok: 203 ids defined across 20 schema-dispositioned tables … 142 registered … 61 backlogged; 0 outside both.
V-32 ok: 130 ids defined across 27 schema-dispositioned tables … 111 registered … 19 backlogged; 0 outside both.

id allocation: V defined, EP2-gate leg now includes V-34; no cross-leg M or V collision; next free M-57 / V-35.

shellcheck --severity=warning, all four extracted bash blocks: clean (EXIT=0)

lint: markdownlint 376 files / 0 errors; frontmatter 202 specs — lint passed
npm test: 1751 tests, 1746 pass, 0 fail, 5 skipped
```

**No digest was recomputed this round.** Every literal named above is byte-identical to its pre-round value; V-34's two are carried copies. The only digest that moved anywhere in this PR remains V-20, in `0abe841`, for the owner's line.

## Discovered issues (added)

- `scripts/measure-secret-fp.js:4` — the docstring cites `docs/specs/WP-secret-fence-two-tier-detector.md`, which no longer resolves. Outside this pass's boundary; a one-line fix for a future WP that touches `scripts/`.

thegyula:
## Round 3 — `b6c7638`

Both findings applied. One fixed by removing structure rather than adding it, as directed.

### 1 [HIGH] Table F was an unguarded authority — FIXED BY SUBTRACTION

Codex is right about the defect and the direction you gave is the right repair. Table F called itself *"the deciding surface for what is frozen"* while sitting outside every live check, so an edit marking F8 retired would have failed nothing — recreating the retirement weakness one level up. **We removed the authority instead of guarding it.**

**Table F's new framing, verbatim:**

> **TABLE F IS A READER'S MAP. IT DECIDES NOTHING AND CONTROLS NOTHING.** It records which of this archived spec's guarantees have live successors and where those successors are, so that somebody arriving at a red check in a `Done` file can find out in one place whether anything was lost. **Every guarantee it names is held by a live check that pins its own literals** — V-34's two ADR digests, V-17's D1/D2 literals — and those checks are the authority. Editing a row of Table F weakens nothing, because a row of Table F enforces nothing.

with the reasoning recorded:

> *Written this way deliberately, in round 3 of the gate. The previous revision called Table F "the deciding surface for what is frozen", which made a narrative table into an unguarded authority — and the recommendation that followed was to add a live check pinning it. **We removed the authority instead of guarding it.** A gate over the table that describes gates is one more turn of the same recursion: it would need its own guard next round, and a guard over a map is a map of a map.*

Two consequences are stated rather than left to be found: V-32's schema not enumerating a blockquoted table is now unremarkable rather than a hole (a map needs no registration), and **nothing checks that Table F is accurate** — if it drifts a reader is misinformed and no guarantee is lost, "which is exactly the difference between a map and a control, and the reason it was safe to stop trying to guard it."

**The promoted invariant, as it now reads inside V-34:**

```
#      THE RETIREMENT INVARIANT, stated here because this is the check that
#      performs a carry and this is where anyone about to undo one is reading:
#
#        A CHECK RETIRES ONLY WHEN A LIVE CHECK CARRIES ITS GUARANTEE, AND THE
#        CARRYING CHECK NAMES WHAT IT CARRIES.
#
#      That is the whole rule. "Its spec reached `Done`" is not a reason, and an
#      archived file is not a place a guarantee can live. This step is the second
#      half of the rule discharged for ADR-0034's two digests: it names what it
#      carries in the paragraph below. **Deleting this step retires nothing —
#      it drops the guarantee**, because after the detector leg was archived
#      there is no other holder.
```

**One correction to your framing, applied honestly rather than repeated.** You said this puts the rule "under a live digest". It does not — leg 2's verification block is digested by nothing, and asserting otherwise would be the same overclaim shape this PR has already self-corrected four times. V-34's comment now says so in terms:

```
#      NOTE WHAT THIS COMMENT IS AND IS NOT: it is an instruction to whoever
#      edits this step, not a control, and it is not itself under any digest —
#      this leg's verification block is digested by nothing. That is deliberate
#      and consistent with the paragraph above. The controls in this step are the
#      two literals below.
```

That is the subtraction applied to itself: the sentence earns its place by sitting where the people who could violate it are already reading, not by pretending to be enforced.

Rows F7/F8 restated as records with "where the guarantee actually lives" pointers; F8 now says explicitly that *"changing this cell to say 'retired' would make this table wrong and would not stop V-34 running."* The three citing surfaces (freeze banner, V-15's RETIRED header, ADR-0034's disposition list) now cite it as a map and state their own authority — for V-15 that is "leg 2's V-17 holds the same guarantee and pins its own literals for it… go and read V-17 if you want to know it is real."

### 2 [MEDIUM] Residual 9's central claim was false — FIXED, and sharpened

Confirmed. `caughtByShippedDetector`'s labelled term is a filter over `scanAndRedact`'s **own** findings, so labelled drift propagates straight into the baseline. My "cannot under-report by construction" was wrong.

**Reproduced, N = 2 000, seed `0xC0FFEE`, `sk-proj-[A-Za-z0-9_]{16,}` → `{20,}`:**

| cell | labelled `{16,}` | labelled `{20,}` | `today` | `proposed` | C3-2 flags? |
|---|---|---|---|---|---|
| `openai-proj/bare` | **77.30%** | **72.85%** | 100.00 → 100.00 | 100.00 → 100.00 | **no** |
| `openai-proj/prose` | 77.30% | 72.85% | 100.00 → 100.00 | 100.00 → 100.00 | **no** |
| `openai-proj/table` | 77.30% | 72.85% | 100.00 → 100.00 | 100.00 → 100.00 | **no** |
| `openai-proj/assign` | 100% | 100% | unchanged | unchanged | no |

Codex's figures are exact — **and they are the `labelled` column, not `today`.** `today` and `proposed` both stay pinned at 100.00% because the entropy tier fires on this fixture regardless of the labelled rule, so C3-2's inequality is *structurally* incapable of seeing the loss. C3-3 does not transcribe `openai-proj`; C3-4 covers only fixtures at 100% in all four contexts, which at 77.30% in three of them this is not.

**What the 4.45 points are, which is the part that matters:** measured on the same run, **89 of 2 000 draws go from carrying a `quarantine` finding to carrying only a `redact` one**. Under leg 2's gate that is **withheld → scrubbed and committed** — the exact disposition change this epic creates, arriving by accident, on a real credential format. Invisible to C3-2 for residual 4's reason (rate, blind to severity) and this residual is the other half: **the baseline is blind too.**

**Scope, measured:** in **47 of C3-0's 156 cells** the shipped entropy pass does not fire on some proportion of draws, so the labelled term alone decides `today` there. Highest: `gitlab-pat/assign` and the three `aws-access-key-id` non-assign cells at 100%, `authress-client-key/assign` at 61.85%, the four `google-client-secret` cells at 48.70%.

**What is NOT claimed:** no executed instance of a lockstep `today` drop. Two attempts to build one — tightening the legacy assignment rule to `{40,}` against `authress-client-key/assign` and `gitlab-pat/assign` — failed to reduce coverage at all, because those fixture bodies clear the tightened bound. The residual states the region as measured and says plainly that an instance inside it is not.

The residual now opens by recording that it has been stated twice in a direction more favourable than the measurement supported, "because the entire value of a residual is that it is true." C3-7's bound sentence carries the same correction and defers the counterexample to the residual.

## Executed evidence

```
V-11 4ca8fe05…  1/1/1/1 · OWNER-REAFFIRMED scores 0 · owner line byte-exact
V-18 leg2 == leg1 == 77a67f3f…      V-33 leg2 == leg1 == 4e5fc859…
V-20 e3fe067f…  unchanged           V-29 908892b2… / f5b90e8c…  == pins
V-21/V-34 494aa6ae… / 57f298f2…     unchanged — still carried copies
V-17 five ADR literals ok · D1/D2 one occurrence each · both canonical rows whole-line

V-16 + V-34 preflight:
  V-0 ok · V-16 P5 ok · V-16 P6 ok
  V-34 ok: ADR-0034's Decisions and its amendment licence are unedited
  EXIT=0

V-34 re-mutation-tested (5 probes on a COPY): Decision heading reworded → digest 1
  moves; `and by nobody else` deleted → digest 2 moves; licence heading renamed →
  empty-range guard fires; Decision terminator renamed → digest 1 moves; terminator
  demoted → count stays 1, digest 2 moves. No mutation left both digests green.

V-30 ok: … 0 outside both.      V-32 ok: … 0 outside both.
shellcheck --severity=warning, four extracted blocks: clean (EXIT=0)
lint: 376 files / 0 errors; frontmatter 202 specs — passed
npm test: 1751 tests, 1746 pass, 0 fail, 5 skipped
```

**No digest recomputed this round.** The only digest that has moved anywhere in this PR remains V-20, in `0abe841`, for the owner's line.

thegyula:
## Round 4 — `59b76be`

Codex is right, and the finding is sharper than a wording problem: round 3 made F7 *claim* a carrier, and the named carrier could not carry it. Verified by execution that the round-3 form is **green** on exactly the input Codex describes.

### V-17's new ADR side

Three parts, each closing a different half:

**(a) Extract the authoritative region first**, content-addressed the way retired V-15 did — E1 by its own first and last lines, E3 by its paragraph markers. This is not tidiness. Measured on the current ADR, the claim shapes occur document-wide:

```
notes untouched pattern, whole file: 4     ← E3 + the errata quote + ER-5 + ER-9
notes withheld pattern, whole file : 3
notes scrubbed pattern, whole file : 2
```

A document-wide "exactly once" rule would be **red on a correct ADR**. Inside the extracted regions each occurs exactly once.

**(b) Assert each region is unique** before extracting, so the block cannot simply be duplicated whole further down the file.

**(c) Count the claim PATTERN, not the expected literal.** Each check matches the claim's shape with its value left open, requires exactly one occurrence in the region, and only then compares that occurrence to the canonical text. Counting the expected literal is precisely what let a contradictory sibling through — the right line was still there, so the count was still one.

The whole ADR side is now one function over a file argument:

```bash
v17_adr_side() {   # <adr-file> — every ADR-side assertion, over ONE file.
  local e1 e3
  v17_region "$1" "E1 evidence-block" '^notes scanned ' '^distinct high-entropy runs '
  e1="$(mktemp)"
  sed -n '/^notes scanned /,/^distinct high-entropy runs /p' "$1" >"$e1"
  …
  v17_claim "notes scanned" "$e1" \
    '^notes scanned +[0-9]+$' 'notes scanned 189'
  v17_claim "notes with ANY finding" "$e1" \
    '^notes with ANY finding \(EP2 reverts today\) +[0-9]+ +\([0-9.]+%\)$' \
    'notes with ANY finding (EP2 reverts today) 109 (57.7%)'
  …
  v17_region "$1" "E3 end-state" '^\*\*E3 — ' '^\*\*E4 — '
  e3="$(mktemp)"
  awk '/^\*\*E3 — /{f=1} /^\*\*E4 — /{f=0} f' "$1" >"$e3"
  …
  v17_claim "notes withheld" "$e3" \
    'notes withheld \*\*[0-9]+ → [0-9]+\*\*' 'notes withheld **109 → 1**'
  v17_claim "notes scrubbed in place" "$e3" \
    'notes scrubbed in place \*\*[0-9]+ → [0-9]+\*\*' 'notes scrubbed in place **0 → 9**'
  v17_claim "notes untouched" "$e3" \
    'notes untouched \*\*[0-9]+ → [0-9]+\*\*' 'notes untouched **80 → 179**'
}
v17_adr_side "$ADR"
```

`v17_claim` fails on a count ≠ 1 **before** it compares values, and the probes call the same function in a subshell — a probe suite that re-implements what it probes is testing itself.

### The three probe results, executed

```
=== PROBE 1: duplicated 'notes scanned' line carrying a different figure ===
FAIL V-17: ADR-0034's E1 evidence-block region starts 2 times, not once.
           A duplicated evidence block means the ADR states the same
           measured fact twice and the D-table agrees with whichever
           copy it was derived from. STOP AND REPORT; do not delete a
           copy to make this pass — an errata amendment decides which
           figure is live.
[REJECTED]

=== PROBE 2: contradictory end state beside the canonical one ===
FAIL V-17: the claim 'notes untouched' occurs 2 times inside ADR-0034's own
           evidence region, not once.
           MORE THAN ONE means the ADR makes the same measured claim
           twice — and if the two disagree, D1/D2 are derived from one
           of them and silently contradict the other. This is the check
           the round-3 existence-only form did not have.
           …
[REJECTED]

=== PROBE 3 (positive control): the current, unmodified ADR-0034 ===
[exited zero: PASSES, as it must]

=== For comparison: the ROUND-3 existence-only form on probe 1 ===
ROUND-3 FORM: GREEN on a document asserting both 189 and 190 — the defect.
```

Both probes now ship inside V-17 and run every time; in-suite output:

```
V-17 ok: ADR-0034's E1 and E3 each state their claims exactly once, and
         D1/D2 are derived from those values
V-17 probe ok: rejected — a duplicated 'notes scanned' line carrying a different figure
V-17 probe ok: rejected — a contradictory end state beside the canonical one
EXIT=0
```

### F7 as it now reads — split, because the single sentence was rounding up

| # | frozen surface | what it stated | live successor |
|---|---|---|---|
| **F7** | V-15's five literals | that this spec and ADR-0034 agree, **and that ADR-0034's evidence bytes do not move except by disclosed errata** | **PARTLY CARRIED, AND THE REST IS CARRIED BY NOBODY.** See the two rows below — this cell used to say "leg 2's V-17 carries the same guarantee", which was rounded up twice over and is corrected here in round 4. |
| **F7a** | the part **leg 2's V-17 does carry** | that ADR-0034's E1 and E3 state the five claims D1/D2 are derived from, each **exactly once inside its own extracted region**, each equal to the value D1/D2 carry, on both sides | **leg 2's V-17**, and it runs. *Round 4 is what made this true.* |
| **F7b** | the part **nobody carries** | byte-immutability of E1's other five lines, E3's prose outside its three claims, `ER_EXPECT`'s pin on the errata rows, and `SWEEP_EXPECT`'s pin on both documents' prose | **none.** *Two of V-15's guarantees are MOOT rather than lost — `M1_EXPECT`'s spec↔ADR byte equality and `M5_EXPECT` both had this frozen file as one operand, and a superseded copy has nothing to stay equal to. The four above are genuinely unheld.* |

F7b also records why it is not closed here (leg 2 mirrors none of those bytes; pinning them would redden it for changes that cannot affect it), what **is** still held (V-34 digests Decisions 1–7 and the amendment licence; V-11 the owner's signature — so the *ratified* content is covered and what is unheld is measured-evidence bytes outside D1/D2's five claims), and what closing it would take.

**The same rounded-up sentence appeared in two other places** — V-15's RETIRED header and ADR-0034's digest-disposition bullet — and both are corrected identically rather than left to be found. That shape has now been caught six times in this PR; each instance is recorded where it was made.

### Checks

```
V-11 4ca8fe05… 1/1/1/1 · OWNER-REAFFIRMED scores 0 · owner line byte-exact
V-18 leg2==leg1==77a67f3f…      V-33 leg2==leg1==4e5fc859…
V-20 e3fe067f… unchanged        V-29 908892b2… / f5b90e8c… == pins
V-34 carried literals 494aa6ae… / 57f298f2… unchanged
V-16 + V-34   EXIT=0
V-17 + probes EXIT=0
V-30 ok — 0 outside both        V-32 ok — 0 outside both
shellcheck --severity=warning, four extracted blocks: clean (EXIT=0)
lint: 376 files / 0 errors; frontmatter 202 specs — passed
npm test: 1751 tests, 1746 pass, 0 fail, 5 skipped
```

**No digest recomputed and no digested region edited this round.** The only digest that has moved anywhere in this PR remains V-20, in `0abe841`, for the owner's line.

thegyula:
# FINAL ROUND — `f72f92a`

**This is the last round on this PR.** The boundary is set deliberately: two fixes below, everything else recorded as a residual, and we ship. The reasoning for drawing the line here is at the end.

## FIX 1 — F7 contradicted itself, and the rule was the thing that was wrong

Round 4 split F7 into "carried" and "carried by nobody", then left the paragraph below it saying F7 retired because V-17 *"carries the same guarantee"*. Both cannot be true. Neither matched the retirement rule in V-34, which admitted exactly one disposition.

**Retiring V-15 was correct. The rule could not describe why.** So the rule changed, not the practice. It is now **per guarantee, not per check** — one check holds several and they need not end the same way:

```
#        A CHECK MAY RETIRE ONLY WHEN EVERY GUARANTEE IT HELD HAS ONE OF THESE
#        THREE DISPOSITIONS, WRITTEN DOWN:
#          (i)   CARRIED — a live check holds it, AND THAT CHECK NAMES WHAT IT
#                CARRIES, so the claim can be checked rather than believed;
#          (ii)  MOOT — one of the guarantee's operands no longer exists, so
#                there is nothing left for it to be about;
#          (iii) DROPPED — nothing holds it, and the drop is RECORDED, naming
#                exactly what is no longer held.
#        (iii) IS A DECISION, NEVER A DEFAULT. Dropping without naming is the
#        failure this rule exists to prevent, and it is the only one of the three
#        that can happen by inattention.
```

Worked in the rule itself, and identically at F7/F7a/F7b and in V-15's RETIRED header:

| guarantee V-15 held | disposition |
|---|---|
| the D1/D2 agreement guarantee | **CARRIED** by repaired V-17, which names it |
| `M1_EXPECT`'s spec↔ADR byte equality, and `M5_EXPECT` | **MOOT** — both had the now-frozen detector spec as one operand |
| byte-immutability of ADR-0034's other evidence lines, E3's other prose, `ER_EXPECT`, `SWEEP_EXPECT` | **DROPPED**, recorded at F7b, each one named |

The contradicting sentence is deleted rather than softened, and the deletion is recorded where it was made:

> *Round 5 removed the sentence that used to sit here — "F7 retired because leg 2's V-17 carries the same guarantee". It contradicted F7b two lines above it and it contradicted the rule itself, which at the time had no clause admitting a dropped guarantee. **The rule was the thing that was wrong**, not the practice… This is the seventh instance in this PR of a claim written stronger than what was true, and like the other six it is recorded where it was made.*

## FIX 2 — two hardenings against failures that need no attacker

**(a) Ordered boundaries and exact width.** Move E1's end marker above its start: `sed` finds no end after the start, runs to EOF, and extracts **817 lines instead of 7** — while both markers still occur exactly once, so the uniqueness check is silent and every claim is satisfied by the real block sitting inside the over-wide region. V-17 now asserts the markers are **in order** and that the extraction is **exactly as long as the marker line numbers imply**. No bound is hard-coded — the expected length is computed from the markers, so it cannot go stale.

**(b) Label counts, independent of punctuation and spelling.** `notes scanned: 190` beside the canonical line is not the canonical spelling, so the shape pattern never matched it, the count stayed 1, and the value still compared equal. Each claim's stable prose label must now occur **exactly once** in its region, in any spelling.

### All four probes, executed

```
V-17 ok: ADR-0034's E1 and E3 each state their claims exactly once, and
         D1/D2 are derived from those values
V-17 probe ok: rejected — a duplicated 'notes scanned' line carrying a different figure
V-17 probe ok: rejected — a contradictory end state beside the canonical one
V-17 probe ok: rejected — a second figure in a NON-canonical spelling ('notes scanned: 190')
V-17 probe ok: rejected — E1's end marker moved above its start (over-wide extraction)
EXIT=0
```

Each new probe fails for the right reason:

```
PROBE 3  FAIL V-17: the LABEL 'notes scanned' occurs 2 times inside ADR-0034's own
                    evidence region, not once.
PROBE 4  FAIL V-17: ADR-0034's E1 evidence-block markers are OUT OF ORDER — the region
                    starts at line 244 and ends at line 238.
```

And measured **without** the hardening, so the gap each one closes is on the record:

```
PROBE 3 without the label count:
  canonical SHAPE matches '^notes scanned +[0-9]+$' : 1   ← round-4 form GREEN
  LABEL occurrences 'notes scanned'                 : 2   ← round-5 check fires

PROBE 4 without the bounds check:
  start marker count : 1    end marker count : 1          ← both still unique
  extracted lines    : 817                                ← vs 7 expected
```

*(The gate reported 822 lines against a slightly earlier ADR revision; 817 is the figure reproduced at the head that ships. Same point, my measurement.)*

## The declared boundary

A threat model now sits beside V-17, in the shape the fence specs already use:

**IN SCOPE — accidental disagreement** between ADR-0034's measured evidence and the D-table: a stale figure left by a re-measurement; a half-applied errata that corrected E1 and forgot E3; a duplicated evidence block; a restatement of a quantity in different words; a boundary edit that makes an extraction mean something other than it says. All five are things an architect or reviewer does by inattention. All five are probed. All five now fail loudly.

**OUT OF SCOPE — a deliberately evasive edit to the ADR**, and V-17 cannot defend against it. Anything still passing after the four probes requires someone with commit access to a ratified, owner-signed ADR to craft a claim that is contradictory in substance, novel in spelling **and** shaped to miss a label count. That is not drift; it is an insider editing the authority his own work rests on.

**The structural defences for that class are elsewhere and are named:** V-34's two digests over the ratified Decisions 1–7 and the amendment licence; V-11 over the owner's signature lines; and review of the diff — the only one of the three that reads meaning.

**Further findings of the form "here is another spelling that slips through" are out of scope for V-17 by declaration.** That class has no fixed point, and this epic has already paid for chasing one twice: ADR-0034 Decision 4 exists because five consecutive fail-open criticals were constructed evasions, and the #117 verification loop was the same shape. A reviewer who constructs one is producing an observation, not a blocker; the answer is a diff review, not a sixth pattern in this step.

## What ships as residual

| # | recorded where | what is not held |
|---|---|---|
| F7b | detector leg's Table F | byte-immutability of ADR-0034's E1 lines outside D1's two claims, E3's prose outside D2's three, the errata rows' pin, and both documents' prose-residue sweep. Nothing holds these. Closing it means a live document taking a whole-block digest over the ADR's evidence — a decision for whoever next needs that guarantee. |
| residual 8 | detector leg | a high-entropy 24-char window inside a longer low-entropy narrow run is not scrubbed. Pre-existing; the pre-WP detector behaved identically. |
| residual 9 | detector leg | the FN matrix's `today` baseline is self-measured and fail-open outside C3-3's 18 cells — 89 of 2 000 draws measured going withheld → scrubbed-and-committed without any check firing. Closed only by an immutable pre-WP oracle. |
| V-17's threat model | beside the step | deliberately evasive ADR edits, as above. |
| discovered | this PR | `scripts/measure-secret-fp.js:4` cites the old leg-1 path (out of boundary); and that script cannot reproduce E1 on a corpus where the `basic-auth` rule has matched. |

## Why the line is drawn here

Five rounds have each found something real and each fix has been smaller than the last: a missing carry (HIGH), a false claim in a residual (MEDIUM), an unguarded authority removed by subtraction, an incomplete successor, and now a self-contradiction plus two accidental-failure hardenings. The remaining class is deliberate evasion of a check by someone who can already edit the thing being checked — and that class is where this epic has twice before spent rounds without converging. The honest move is a declared scope with the evidence attached, which is what shipped. Everything not fixed is named above, in the document, where the next reader inherits it.

## Checks

```
V-11 4ca8fe05… 1/1/1/1 · OWNER-REAFFIRMED scores 0 · owner line byte-exact (trailing space intact)
V-18 leg2==leg1==77a67f3f…      V-33 leg2==leg1==4e5fc859…
V-20 e3fe067f… unchanged        V-29 908892b2… / f5b90e8c… == pins
V-34 carried literals 494aa6ae… / 57f298f2… unchanged
V-16 + V-34   EXIT=0
V-17 + four probes EXIT=0
V-30 ok — 0 outside both        V-32 ok — 0 outside both
shellcheck --severity=warning, four extracted blocks: clean (EXIT=0)
lint: 376 files / 0 errors; frontmatter 202 specs — passed
npm test: 1751 tests, 1746 pass, 0 fail, 5 skipped
```

**No digest recomputed and no digested region edited this round.** Across the whole PR, exactly one digest moved: **V-20**, in `0abe841`, for the owner's line — `ec53884c…` → `e3fe067f…`, with the owner-only intermediate `a6add80a…` recorded so his edit stays separable from the architect's.

PR #122, opened 2026-07-28
feat(dream): EP2 redacts in place on a redact-severity finding (WP-secret-fence-ep2-redact-arm)

thegyula:
Spec: docs/specs/WP-secret-fence-ep2-redact-arm.md

Leg 2 of the `secret-lifecycle` split. Leg 1 made the detector two-tier; EP2 still
withheld a whole note on **any** finding of either severity. This WP makes the
staged-output gate consult severity: on a findings set with no
`quarantine`-severity finding it preserves the unredacted original into
`state/quarantine/redacted/`, rewrites **only the lines this run added** to their
sanitized form, stages the scrub in the git index **before** the working tree,
commits the note, counts it as `secretRedactions` (so transcripts are not
deferred), and announces it in the dream report. Any `quarantine` finding keeps
today's behaviour byte for byte.

Both new delete paths are guarded by **byte identity**, not copy existence:

* the fall-through deletes its `redacted/` copy only when a byte-identical
  withheld copy demonstrably exists (`Buffer.compare` of the two copies), and
* the gate **aborts without reverting, removing, or clearing the index entry**
  when no durable artefact holds the bytes that are on disk *right now* — Table B
  row **B3b**, reached from **R0** (both preserves failed) and **R0b** (a copy
  exists but is of different bytes, or the identity read could not be performed).

## Deliverables checklist

Twelve changed paths = the eleven Deliverables rows plus the spec file (exempt).
`memory/lessons/inbox.md` is **not** in the changed set (V-25's explicit check).

## Verification output

**V-16 (Table P) and V-34 ran FIRST, before a line of code was written** — both
are stops, not reports:

```
V-0 ok: an unstaged edit to a forbidden file IS visible to changed_paths
V-16 P5 ok: the redact producer cannot span a newline
V-16 P6 ok: Authorization: Basic is quarantine and its body is fully redacted
V-34 ok: ADR-0034's Decisions and its amendment licence are unedited
```

P1–P4 are silent `must_eq`/`must` assertions and passed; the block exits non-zero
on the first failure.

**The complete verification block, run as a script** (`set -euo pipefail`, exit 0;
`shellcheck -s bash` on the extracted block is clean):

```
V-0 ok: an unstaged edit to a forbidden file IS visible to changed_paths
V-16 P5 ok: the redact producer cannot span a newline
V-16 P6 ok: Authorization: Basic is quarantine and its body is fully redacted
V-34 ok: ADR-0034's Decisions and its amendment licence are unedited

# V-1 full suite (node tests/run.js)
tests 1805 | pass 1800 | fail 0 | skipped 5

# V-4  src/cli/dream.js and src/core/secret-scan.js untouched   — silent, passed
# V-5  digest.js banner head byte-identical, Q1+Q14 replaced, PERM5 bound — silent, passed

# V-9 targeted suites (dream-validate, digest, private-fs)
tests 218 | pass 217 | fail 0 | skipped 1

# V-10 lint
--- markdownlint ---   Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 202 spec(s), 4 agent(s)
lint passed

# V-13 runbook: all three Q edits present, PERM13 removed-lines bound — silent, passed
V-17 ok: ADR-0034's E1 and E3 each state their claims exactly once, and
         D1/D2 are derived from those values
V-17 probe ok: rejected — a duplicated 'notes scanned' line carrying a different figure
V-17 probe ok: rejected — a contradictory end state beside the canonical one
V-17 probe ok: rejected — a second figure in a NON-canonical spelling ('notes scanned: 190')
V-17 probe ok: rejected — E1's end marker moved above its start (over-wide extraction)
# V-18 ratified threat-model digest   — silent, passed
V-33 ok: the shared check contracts match the cross-leg digest
# V-20 split-provenance digest        — silent, passed
# V-22 glossary extracted permitted-removals bound — silent, passed
# V-24 ledger banner: one removed line per file, dream.test.js unchanged — silent, passed
# V-25 permission boundary            — see below
V-23 ok: quarantine/redacted/ is inside the A5 private tree
# V-11 ADR Accepted + both OWNER-SIGNED lines: count, placement, pinned digest — silent, passed
V-26 ok: the Table Q family is exactly the registered file set
V-27 ok: the threat model states the shipped contract and nothing else moved
V-28 ok: validate.js's own comment states the shipped contract (Q17)
V-30 ok: 203 ids defined across 20 schema-dispositioned tables (8 corpus/measurement,
         registered by table) plus the acceptance-criterion, verification-step and
         accepted-residual families; 142 registered in the Checklist under H6's
         boundary form; 61 on the dated backlog and therefore NOT registered;
         0 outside both.
V-31 vocabulary, read from Table J: safety={safe|safety|safely|safest|unsafe|authoriz}
     subjects={revert|withhold|destr|delet|remov|K3|K4}
V-31 ok: view = whole file minus this step and Table J's rows, fenced blocks
         INCLUDED, read through a 3-line sliding window; vocabulary read from
         Table J; J6 collisions excised; allowlist subtracted; all four
         published bypasses fired as self-tests; 0 unregistered claims.

EXIT: 0
```

**V-25 — the permission boundary**, run explicitly so the result is visible
(`scripts/boundary-check.js` is silent on success):

```
$ git diff --name-only origin/main...HEAD
docs/GLOSSARY.md
docs/THREAT-MODEL.md
docs/runbooks/secret-incident.md
docs/specs/WP-secret-fence-ep2-redact-arm.md
src/core/digest.js
src/core/dream/ledger.js
src/core/dream/validate.js
src/core/private-fs.js
tests/unit/digest.test.js
tests/unit/dream-validate.test.js
tests/unit/ledger.test.js
tests/unit/private-fs.test.js

$ node scripts/boundary-check.js docs/specs/WP-secret-fence-ep2-redact-arm.md <changed…>
V-25 boundary-check exit: 0
```

**V-26's membership sweep** returned exactly the ten files its `QREG` heredoc
registers (`diff -u` empty) — no `+` line, no `-` line. `QSWEEP` was **not**
touched.

### Fault injections and Table R coverage

Every row of Table R has its own test (thirteen rows), plus the two
consequence-2 keep-combinations, the derivation harness, both invariants and the
residual pin. All seams live in the test file; nothing under `src/` was edited to
make a fault reachable.

| id | row / property | mechanism |
|----|----------------|-----------|
| FI-1 | R1 | path-specific `writeFileSync` EACCES under `redacted/` only |
| FI-2 | R2 | stateful `readFileSync` throw on the **2nd Buffer read** of the target |
| FI-3 | R3 | direct `scrubAddedLines(…, [99])` / `[0]` — helper only |
| FI-4 | R4 | `require.cache` stub: always `text+'!'` plus a finding |
| FI-5a / FI-5b | R5 | dir `0500` (open fails) / truncated write then ENOSPC |
| FI-6 | R6 | a **different** stub: finding on call 1, none after — order-independent |
| FI-7 | R7 | `spawnPinnedSync` wrapper: one-shot fail of `update-index`; also `hash-object`; also empty `ls-files` stdout. Helper **and** gate |
| FI-7b | R7b | `renameSync` EIO for `.wienerdog-scrub.` sources |
| FI-16 | R7c | different bytes written **before** the comparison read; rename asserted never invoked |
| FI-8 / FI-9 | R9 | tracked `checkout` / untracked `add -A -- rel` fail after an R7 |
| FI-12/13/14 | R0 | no `stateDir` / ENOSPC over the whole quarantine tree / `stateDir` `0500` with `quarantine/` absent — **each ×2, tracked and untracked** |
| FI-17 / FI-18 | R0b | the cross-product: a save at K2 (→ R7c) **and** the withheld preserve fails |
| FI-19 | R0b via K4's throw | R7 trigger (leaves the target byte-unchanged) + withheld-preserve write fails + K4 read throws |
| FI-10 | consequence 2, keep #1 | withheld preserve fails, redact copy survives, Q8 suffix appended |
| FI-11 | consequence 2, keep #2 | K1 delegates, K2 throws, K3 returns **different** bytes → both copies kept and they differ |
| FI-15 | rule 1 | three-method ordered event log; asserts **exactly two** target reads before the rename, the first a Buffer read, the second between the temp write and the rename |
| BU | byte-unchanged, R2–R7b | helper level: read before, call directly, assert `false` **and** byte equality |
| I1 / I2 | AC-24 | `appendFileSync` snapshot at Step 4 (R8 + tracked R7 + untracked R7); `renameSync` snapshot immediately before the scrub's rename |
| RP-1 | residual 11 | a save written **inside** the patched `checkout` / `rmSync`, immediately before delegating — the loss is asserted, and that no artifact claims otherwise |

The abort message is asserted **field by field on all three arms** (Table Q row
Q18): the path in its exact `JSON.stringify` form; which preserves failed
(discriminated between R0 and R0b, with the other arm's wording asserted
**absent**); the identity disposition (three distinct values, each with the other
two asserted absent); and the surviving basename (present on R0b, asserted as an
**absence** on R0). Each arm is additionally driven with a newline-bearing and an
ESC-bearing path, asserting the complete expected rendering, no raw newline and
no raw ESC — and `04-Atomic/a\nb.md` vs `04-Atomic/ab.md` are asserted to render
**differently**, which is what discriminates JSON-string encoding from any
dropping or replacing scheme.

### Mutation checks — all 36 rows go red

Each mutation was applied to `src/`, its named check run, and the file restored.
Working tree verified clean afterwards; targeted suites re-run green (248 tests,
0 fail).

```
M-7     RED (ok)       redact arm never runs
M-8     RED (ok)       scrub-then-preserve
M-9     RED (ok)       redact arm increments secretReverts
M-10    RED (ok)       drop withFileTypes
M-11    RED (ok)       scrub rewrites the whole file
M-12    RED (ok)       proceed when the preserve failed
M-16    RED (ok)       ignore the scrub return value
M-17    RED (ok)       drop the re-scan verification
M-18    RED (ok)       raise the cap so the prune never fires
M-19    RED (ok)       drop the A5 element
M-20    RED (ok)       drop the index-first stage
M-21    RED (ok)       single writeFileSync onto the target
M-22    RED (ok)       a completed B4 enters reverted[]
M-23    RED (ok)       prune by filename, no exclusion
M-23b   RED (ok)       prune by filename sort
M-24    RED (ok)       require ,b in the hunk pattern
M-24b   RED (ok)       a missing ,d contributes zero lines
M-25    RED (ok)       drop '; not committed' from B3's reason
M-26    RED (ok)       revert the ledger banner sentence
M-27    RED (ok)       delete B3a's git add
M-28    RED (ok)       delete the copy on every fall-through
M-29    RED (ok)       revert the digest banner sentence
M-30    RED (ok)       write-then-stage
M-31    RED (ok)       drop the Buffer.compare delete guard
M-32    RED (ok)       leave the digest.js comment at its old text
M-33    RED (ok)       leave the validate.js Step-3 comment at its old text
M-45    RED (ok)       delete row B3b
M-46    RED (ok)       restore the second content read
M-47    RED (ok)       drop the pre-rename compare
M-48    RED (ok)       prune per call instead of per run
M-49    RED (ok)       weaken B3b to a copy-EXISTENCE test
M-50    RED (ok)       K4's catch path becomes an ordinary revert
M-51    RED (ok)       omit the basename from the abort message
M-53    RED (ok)       omit the path from the abort message
M-54    RED (ok)       omit WHICH preserve failed
M-55    RED (ok)       collapse the identity disposition into one value
M-56    RED (ok)       render the path RAW instead of JSON-escaped

ALL MUTATIONS CAUGHT
```

Sample failing output — M-49, the shipped design of every revision before round 5:

```
✖ EP2 redact arm R0b (FI-18 untracked): a copy exists but is of the WRONG bytes
✖ EP2 redact arm R0b (FI-17 tracked): a copy exists but is of the WRONG bytes
✖ EP2 redact arm R0b (FI-19, untracked): the identity read cannot be performed
```

M-50, which every other mandated arm survives:

```
✖ EP2 redact arm R0b (FI-19, untracked): the identity read cannot be performed
✖ EP2 redact arm R0b (FI-19, tracked): the identity read cannot be performed
✖ EP2 abort message R0b: the hostile path is escaped on the read-error arm too
```

M-46, re-keyed to the read count rather than to a perturbation:

```
✖ EP2 redact arm FI-15: exactly the two reads the contract names, in that order
```

## Decisions made

1. **FI-10 needed a fall-through trigger its Table T row does not name.** Table T
   gives FI-10 only the `writeFileSync` patch that fails writes under
   `quarantine/` outside `redacted/`. On its own that patch leaves the arm
   completing at **R8** — the redact preserve writes inside `redacted/` and the
   scrub's temp is in the vault, so neither is touched, B3 never runs, and
   consequence 2 (the cell FI-10 exists to produce) is never reached. This is the
   same defect round 6 recorded for FI-19. I used **FI-7's `spawnPinnedSync`
   trigger**, exactly as FI-19's row prescribes and for the reason it gives: it is
   the fall-through that leaves the target byte-unchanged, so the identity read
   still shows the copy is of the bytes on disk and the withhold proceeds into the
   keep-combination.

2. **FI-19's K4 is identified structurally, not by an absolute read count.**
   FI-19's mandated trigger is an R7 (the staging failure), and on an R7 the
   pre-rename comparison **K2 never executes** — the arm returns before it. The
   reads on that path are K1, K3, K4, so the row's round-8 note ("a counter that
   lets two reads through and throws on the third lands the EACCES on K3") does
   not describe the sequence its own trigger produces. Rather than hard-code
   either count, the injection throws on the first target Buffer read **after the
   withheld preserve's write has failed** — which is K4 by construction, on both
   arms, immune to the count question. M-50 confirms it reddens.

3. **`M-23` was split into `M-23` and `M-23b`.** The spec's M-23 states two
   independent mutations in one cell ("prune by filename sort instead of
   `(mtimeMs, name)`, **and** stop excluding the just-created copy"). Applied
   together, either half alone satisfies the row, so a single run cannot show that
   both clauses are held. They are run separately; both go red, against different
   tests.

4. **The abort message's wording is mine, within Q18's four decided fields.** Q18
   decides *what* the message states, not the sentence. It is plain language for a
   knowledge worker, one line, and each field's wording is asserted — and
   mutated — independently, so no phrase covers two arms.

5. **`quarantinePreserve`'s `kind` validation sits outside the try**, so it throws
   as Table R consequence 5's one deliberate exception while the rest of the body
   stays total over `{shape, null}`.

6. **Retention: the cap counts every regular file in `redacted/`**, while the
   candidate set is the date-prefixed ones minus the run's own (N3). The spec
   fixes the candidate set precisely and leaves the denominator implicit;
   counting all regular files is the simpler reading and is what makes "at most
   N1 remain" true of what the user sees in the folder.

7. **`scrubAddedLines`'s line array excludes the trailing empty element** produced
   by a final newline, so `1 … lines.length` is the file's real line count and the
   join round-trips byte-exactly.

1. **Table T's FI-10 row names a mechanism that cannot reach the cell it
   produces** — see Decision 1. The row needs a fall-through trigger clause, as
   FI-19's row already carries. Not fixed here: it is a spec edit.

2. **Table T's FI-19 row describes a read sequence its own trigger does not
   produce** — see Decision 2. Its round-8 correction assumes K1, K2, K3, K4, but
   the trigger it mandates (the `update-index` failure, chosen precisely because
   it leaves the target byte-unchanged) returns before K2 ever runs, so the
   sequence is K1, K3, K4. The row's *outcome* is right and reachable; only its
   let-through list is arithmetically wrong. Again a spec edit.

3. **`docs/specs/README.md:11`'s ≤ 8-file sizing heuristic is exceeded at
   eleven**, which the spec discloses and argues for. Noted, not touched.

## Dogfooding lessons

- **WP-secret-fence-ep2-redact-arm:** A fault injection that names a *mechanism*
  but not a *trigger* is unproducible in exactly the way the spec's own round-6
  finding describes — and it fails silently green, because the arm completes
  successfully and the assertion about the failure path is simply never reached.
  Before writing an injection, trace which rows its patches actually touch; two of
  this spec's rows (FI-10, FI-19) needed a trigger their cells did not name.
- **WP-secret-fence-ep2-redact-arm:** Read-counting seams must be anchored to a
  *structural* event, not to an ordinal. FI-19's "throw on the Nth read" broke
  because its own mandated trigger short-circuits before one of the reads it was
  counting. Anchoring on "the first read after this other thing failed" is immune
  to that, and to any future reordering.
- **WP-secret-fence-ep2-redact-arm:** A permitted-removals heredoc bounds *removed*
  lines, so rewrapping a JSDoc block near an edit fails the bound even when the
  edit itself is in scope. V-5 caught a rewrap of `listSecretQuarantine`'s doc
  comment that was pure tidying; the fix was to leave the original lines byte-exact
  and *append* the new sentence. The same trap bites single-line grep sentinels:
  my first `digest.js` comment rewrap split `CONTENT is never read or rendered`
  across two lines and V-5's positive `grep -qF` went red.
- **WP-secret-fence-ep2-redact-arm:** Mutation rows stating two independent changes
  in one cell cannot demonstrate that both clauses are held. Split them and run
  each; the spec's M-23 is the worked instance.
- **WP-secret-fence-ep2-redact-arm:** A mutation driver that edits `src/` in place
  makes the working tree lie for the duration. Commit first, and do not read source
  files while it runs — I briefly read a mutated `validate.js` and had to re-check
  it against the commit.
- **WP-secret-fence-ep2-redact-arm:** The retention fixtures are the slow ones (a
  51-note run is ~4s of git spawns and there are four such cases), which makes the
  full mutation sweep a ~25-minute job. Worth budgeting for, and worth *not*
  shortening the fixtures: the 51-note boundary is the only place the cap-yields
  precedence is observable at all.

---
Generated-by: claude-opus-5


thegyula:
## Codex finding [P2] addressed — `d7fd6ba`

**Whose defect: mine, not the spec's.** The spec's `scrubAddedLines` contract says
*"Each added line number L is replaced by `scanAndRedact(lines[L-1]).text`"*, which
implies a decode — you cannot index a `lines` array or call `scanAndRedact` (a
string function) without one. But it **never names an encoding, never says how the
file is reconstituted from `lines`, and says nothing about invalid UTF-8**. Grepped
at the head that ships: `lossless`, `round.?trip`, `latin`, `invalid utf`, `U+FFFD`,
`non-utf` return **zero hits** anywhere in the spec. Table K's `encoding` column
governs the four *reads* (all Buffer), not the decode inside the scrub. So
`captured.toString('utf8')` + re-encode was my realization of an under-specified
contract, and failing to verify losslessness is mine to fix. No spec amendment is
owed for the decode itself; two smaller gaps are recorded below.

**Reproduced first, not assumed** (`git init` + a Latin-1 note + one added ASCII
line):

```
numstat: "1\t0\t04-Atomic/l1.md\0"     ← git calls it TEXT, so the binary
binary? false                             fail-closed branch never fires
diff -U0: @@ -2,0 +3 @@ second line
          +ref q7PmXz4KvR9tWc2LbN8dYfGh in prose
lossless round trip? false
re-encoded length 71 vs captured 69
U+FFFD present in re-encode? true
```

Two bytes of the user's own text, on a line this run never added, replaced by a
replacement character — and the pre-rename compare does **not** catch it, because
it compares the target against `captured`, and the target is unchanged.

### The guard as shipped — fail-closed, per the coordinator's ruling

```js
function isLosslessUtf8(buf) {
  return Buffer.compare(Buffer.from(buf.toString('utf8'), 'utf8'), buf) === 0;
}
```

Held in **two** places, one predicate, no second copy:

* **`scrubAddedLines`**, as its first act — `if (!isLosslessUtf8(captured)) return false;`
  so the **exported** helper is safe for every caller, not only for the gate that
  checks before calling it;
* **the gate**, so the withhold reason names the cause. The arm declines outright,
  never calls the scrub, writes no temp and stages nothing.

The note is then **withheld** — the behaviour this gate had before the redact arm
existed, which is the right default for a file that cannot be processed without
risking its bytes. The reason gains one suffix:

```
(not rewritten: this note is not valid UTF-8 text, so the secret could not be
 replaced without changing the rest of it)
```

Splicing into the original buffer (Codex's other option) was **not** taken: it is
the more elegant fix and preserves the captured-buffer discipline end to end, but
it rewrites the scrub path late in a heavily-gated WP and changes the shape of
every assertion that reads the rebuilt text. Fail-closed is small, lands the note
in a known-good behaviour, and is what the coordinator ruled.

### Does any Table R / Table B row cover the outcome?

**Table B: yes — row B5a** covers the transition exactly ("B4, the preserve
succeeded, and the arm did not complete → fall through to B3", `secretRedactions`
not incremented, `secretReverts` incremented). Nothing invented.

**Table R: no row covers the CONDITION.** The thirteen rows enumerate: the redact
preserve failing (R1/R0), K2 throwing (R2), an out-of-range line number (R3), the
re-scan still firing (R4), the temp write failing (R5), a no-op rewrite (R6), the
stage failing (R7), the rename failing (R7b), K2's bytes differing (R7c), success
(R8), the two aborts (R0/R0b) and the fallback failing (R9). "The capture cannot be
decoded losslessly" is none of them. The new outcome **shares R2–R7c's entire cell
set** — `P` = the copy, `S` = not called, working tree byte-unchanged, `redacted/`
copy deleted by consequence 2, index cleared by B3/B3a, next row B5a → B3, no
redaction, +1 revert, in `reverted[]` — but no row names the condition. Per the
instruction I did **not** invent a row; it is filed under Discovered issues.

### New tests

1. **`a NOT-lossless-UTF-8 note is withheld, and not one byte of it is rewritten`** —
   a tracked Latin-1 note (`0xE9`) whose HEAD version is committed, plus one added
   ASCII line carrying a redact-severity finding. Asserts, with git's own
   classification pinned as a precondition (`numstat` is not `-\t-`, so the binary
   branch is provably not what catches it): `secretRedactions === 0`,
   `secretReverts === 1`, not committed, the working tree **byte-identical** to the
   committed original, **no `EF BF BD` anywhere**, the withheld copy byte-identical
   to the pre-revert bytes with `0xE9` intact, `redacted/` empty, the reason
   carrying the new suffix, and **no `## Redacted in place` section** in the report.
2. **`scrubAddedLines itself refuses a not-lossless buffer (helper)`** — direct
   call returns `false`, target byte-unchanged, **plus a valid-UTF-8 control on the
   same fixture shape that still returns `true`**, so the guard is not simply
   refusing everything.

### New mutation — M-57

Id allocated by **running** the spec's allocation procedure with both leg files
final (not by continuing a range):

```
M defined, detector leg : M-1 M-13 M-14 M-15 M-2 M-3 M-34 … M-44 M-5 M-52 M-6
M defined, EP2-gate leg : M-7 … M-56
next free M: M-57
```
No cross-leg M collision; the shared V set is exactly `V-1 V-10 V-11 V-18 V-33`.

**M-57 — remove the lossless-UTF-8 round-trip check** (`return true;`). Reddens
both new tests, and reddens them in the corrupting direction:

```
M-57 RED (ok)
   ✖ EP2 redact arm: a NOT-lossless-UTF-8 note is withheld, and not one byte of it is rewritten
       actual: 1, expected: 0        ← secretRedactions: the note WAS scrubbed
   ✖ EP2 redact arm: scrubAddedLines itself refuses a not-lossless buffer (helper)
       actual: true, expected: false
```

### Full re-verification at `d7fd6ba`

* **Verification block** (extracted, `shellcheck -s bash` clean) — **exit 0**. All
  steps green: V-0, V-1, V-4, V-5, V-9, V-10, V-11, V-13, V-16, V-17 (+ 4 probes),
  V-18, V-20, V-22, V-23, V-24, V-25, V-26, V-27, V-28, V-30, V-31, V-33, V-34.
* **`npm test`** — `tests 1807 | pass 1802 | fail 0 | skipped 5`.
* **`npm run lint`** — `lint passed`, markdownlint `0 error(s)`, frontmatter
  `202 spec(s)`.
* **Mutation sweep — all 38 rows RED**, M-7 … M-57 (`ALL MUTATIONS CAUGHT`). Three
  anchors (M-8, M-12, M-16) were re-pinned to the new branch shape and verified by a
  dry-run pass before the sweep; tree verified clean afterwards and the targeted
  suites re-run green (250 tests, 0 fail).

### Owner surfaces — re-checked byte for byte after the change

```
## OWNER-APPROVED block   origin/main 4d80cd33…  HEAD 4d80cd33…   identical
OWNER-REAFFIRMED line     …2026-07-27 20 0a       ← trailing space intact
spec diff                 1 file changed, 1 insertion(+), 1 deletion(-)  (status only)
```
V-11 (both `OWNER-SIGNED` lines, pinned digest), V-18, V-20, V-33 and V-34 all
green, so every pinned literal is intact.

### Added to Discovered issues

4. **Table R has no row for a note the arm cannot process.** The lossless-UTF-8
   refusal produces R2–R7c's exact cell set through B5a, but no row states the
   condition. A spec amendment is owed — a row, not an implementer's invention.
5. **The new reason suffix is an unregistered Table Q member.** Q8 is the
   registered precedent for a redact-arm reason suffix; this one has no row. V-26
   stays green because it pins the file set and `validate.js` is already a member,
   so nothing mechanized catches it — which is itself the point.

### Added to Decisions made

8. **The guard is held in both the gate and the helper, from one predicate.** The
   gate's copy exists only so the withhold reason can name the cause; the helper's
   is the correctness boundary, because `scrubAddedLines` is exported and a future
   caller would otherwise inherit the corruption.

### Added to Dogfooding lessons

* **WP-secret-fence-ep2-redact-arm:** A captured buffer only protects the bytes you
  never decode. `Buffer.toString('utf8')` never throws — it substitutes U+FFFD —
  so a lossy decode makes "I only touched the added lines" false for the *whole
  file*, silently, on a file git happily calls text. Any per-line rewrite of a
  user's file needs a round-trip assertion before the first re-encode, and the
  pre-rename compare cannot stand in for it: it compares the target against the
  capture, and on this path the target never changed.

PR #144, opened 2026-08-02
docs(specs): pre-dispatch reconciliation of WP-scheduler-node-path-durability against the landed sibling (WP-scheduler-node-path-durability)

thegyula:
Pre-dispatch revision of `docs/specs/WP-scheduler-node-path-durability.md` only. No ADR, no `src/`, no `tests/`, no other spec. Status stays **Ready**.

**Tested SHA: `1093e51`** (main). All Current-state claims were re-executed there on 2026-08-02.

## 1. Gates satisfied — dated records added, history retained

Both dispatch gates were open when this spec was last revised. Both are now closed, and each carries a dated satisfaction record placed beside (never replacing) the text it supersedes.

- **DISPATCH PRECONDITION** (banner + new Definition of done item **0d**): `WP-scheduler-register-replaces-loaded-record` is **`Done`** — implemented and merged as **PR #140** (merge **`7b22d71`**: +266 `src/cli/schedule.js`, +132 `src/scheduler/generators.js`, +1197 `tests/unit/scheduler-schedule.test.js`), Done-flipped by **PR #143**, now at `docs/specs/done/`. Post-merge suite on `main` @ `1093e51`: `ℹ tests 1897 / ℹ pass 1888 / ℹ fail 0 / ℹ skipped 9`.
- **Definition of done item 8 (ADR gate)**: the dated amendment is **owner-signed**. `docs/adr/0028-…:1213` is the amendment heading; **`:1215`** reads byte-exact `Status: **ACCEPTED - OWNER-SIGNED 2026-08-02**`. Landed via **PR #138**. `:83` is correctly left unedited per the ADR's own preamble convention.

**No owner marker was written by this pass.** Item 8 is discharged by *citing* Gyula's existing signature; the item now says so explicitly, and the implementer still touches no file under `docs/adr/`.

## 2. Every anchor re-verified at `1093e51`

Each Current-state subsection is now tagged **HOLDS** / **MOVED** / **REWRITTEN**, with `5f0ffc0`'s numbers retained in parentheses.

| | |
|---|---|
| **HOLDS** | `generators.js:16-22` (`nodePath`), all four renderer interpolations (`:364 :407 :482 :606`), `descriptor.js:215`, `status.js:80` + `:125-133` + `:283` + `reloadMissing :354-392`, `darwinReplaceEntry :51-55`, `doctor.js:317`, `launcher.js:348`, `exec-identity.js:203`, `generators.js:101-122` + `:178-180`, `exec-identity.test.js:26`, `scheduler-entry-identity.test.js:423-430`, `scheduler-generators.test.js:421`, `ci.yml:32-34` |
| **MOVED** | the six ENTRY sites `:303 :342 :417 :638 :667 :752` → **`:477 :536 :611 :856 :885 :970`**; `run-job.js :408 :529` → `:439 :579`; `generators.js:997` → `:1125`; `sync.js:240` → `:241`; `GLOSSARY.md:25` → `:26`; the notice `:584` → `:802`; the reload warning `:464` → `:678`; `ensureWindowsTaskRegistered :240-245` → `:412-417`; `run-job.js:1236` → `:1318`; Table E row 8's mutation target `:417` → `:611` |
| **REWRITTEN** | **Current state §8** — the bare-`bootstrap` sites (`:315`, `:431`) it described **no longer exist** |

Re-executed and byte-identical: §4's Homebrew layout (all four commands), §5's live entry + empty `launchctl getenv PATH`, and V3/V4/V5/V5b. **Re-baselined:** V1 `82/79/0/3` → **`127/120/0/7`**, so V2's threshold moves from "> 79" to **"> 120"**.

## 3. Design reconciliations, with reasoning

**(a) What "the entry's node path" means now.** The canonical `node` feeds both the renderer *and* the readback expectation. `darwinLoadedVerdict` (`schedule.js:134-139`) compares it **twice at FATAL tier** — as `arguments[0]` and as `program`. So a node-path change yields **`'mismatch-fatal'`**, which sibling Table A1 makes the *one* verdict authorizing teardown-and-replace. The sibling's Table A2b tiers `program` FATAL and names this WP by slug as its justification: the interlock is deliberate on both sides.

**(b) Rows the sibling's landing changed.**

| Row | Was (`5f0ffc0`) | Now (`1093e51`) | Why |
|---|---|---|---|
| C-5 macOS, healthy loaded record | **NO**, silently reported as success | **YES — converges on the first `sync`** | `'mismatch-fatal'` → `bootout` + `bootstrap` + `verifyLoaded()` |
| C-4 linux, degraded reload | **NO**, silently reported as success | **NO, but loud and retried every `sync`** | `loaded = reloadOk && enableOk` (`:681`); reload hoisted out of `if (changed)` (`:671`) |

**The old "sync refused by launchd then silent forever" hazard is handled by the replace path — stated explicitly in the spec.** Table D's false-success rows are **gone on both platforms**, by two independent mechanisms: macOS re-reads the loaded record on *every* register (ungated by `changed`), and Linux re-runs and re-warns on *every* register. That was the entire substance of the dispatch blocker, and it is discharged.

**Residual transition hazards, re-derived (not assumed away):**
- **Table C row 7 (NEW)** — macOS `'indeterminate'` readback: `'indeterminate'` is the absence of evidence and never authorizes teardown (`:208`), so the entry stays pinned. Loud on every `sync`, never falsely converged. New sub-table **D-c** reports it.
- **Table C row 4** still does not converge; it is merely visible now.
- Row 6 (step-8b heal) is demoted to a **backstop**, since row 5 converges immediately.

**(c) Test-index / AC items keyed on old shapes.** AC9 and DoD-7 both asserted "rows 4 and 5 do not converge" — now false for row 5, so both were restated (repeating them would be an inaccurate *under*claim). AC6's non-claim was re-pointed at rows 4 and 7. Table E row 8 re-anchored. `tests/unit/scheduler-schedule.test.js` stays a **non-deliverable**: this WP changes the node *value*, not the register path, and the sibling's `fakeLaunchd` suite already covers the mismatch-fatal→replace path for arbitrary `argv[0]`.

**New implementer rule (Implementation notes §D2).** At `:477` and `:611` the hoisted `const node` feeds **both** the renderer and `expect.argv[0]`. The single-token swap keeps them consistent automatically — but splitting them (durable path in the renderer, pinned in `expect`) would make every register grade `'mismatch-fatal'`, churning a real `bootout`+`bootstrap` on every `sync` with a permanent false failure report. **No test in this WP's Deliverables would catch that**, since T1/T2/T4/T5 test `generators.js` in isolation. Spelled out, with the instruction to leave it hoisted.

## 4. Open design question — Table G (NEW canonical table, gated by DoD item 9)

**Not settled, not guessed.** Table C row 5 and all of Table D-a rest on an unverified premise: that `launchctl print` echoes `ProgramArguments[0]` **verbatim** into `program` rather than realpath-resolving it. **This WP is the first change that can falsify it**, because it is the first to register a *symlinked* program path.

- **Evidence for:** `properties = inferred program` in the sibling's executed dump; and the incident signature itself (after the Cellar dir is deleted the record still prints clean), which is store-verbatim/resolve-at-spawn behavior.
- **Why unresolved:** the live entry's `ProgramArguments[0]` is already its own realpath, so the dump cannot distinguish "echoed" from "resolved and equal". Settling it needs a symlinked-program job or a throwaway `bootstrap` — a mutation of the user's launchd state, which is out of bounds for a spec-authoring pass.
- **If false:** permanent `'mismatch-fatal'` ⇒ teardown churn + a permanent false failure on every `sync`. Loud, never silent, and the schedule keeps working — but **not shippable**. The fix would be on the comparison side (`darwinLoadedVerdict`), which is the **sibling's code, outside this WP's Deliverables** ⇒ a new WP. `entryNodePath` and the six swaps are correct either way, so **the WP is not obsoleted or simplified**.
- **`npm test` cannot reach this:** `tests/run.js` sets `WIENERDOG_TEST_NO_REAL_SCHEDULER=1`, and `fakeLaunchd` renders `program: argv[0] || ''` from the plist it stored at `bootstrap` — it models the premise **true by construction**. The spec says not to cite a green suite here.
- **Gate:** Definition of done item **9** — one command on a real Homebrew macOS host after implementation, with both branches specified and a "do not merge" on the false branch.

## 5. ADR-0031 discipline

Table G added (six canonical tables now). All `(+r5)` mirrors registered in the Mirrored Surface Checklist in the same pass: Current state §2's new hoisted-`node` sub-table, §8, §9, both reconciliation preambles, DoD item 1's pass-count threshold, D2's do-not-touch list, the Out-of-scope registration bullet, AC9, and the sibling's Tables A1/A2b as out-of-spec mirrors this spec derives from and may never edit.

## Verification

```
npm run lint                     → lint passed (markdownlint 0 errors, frontmatter 209 specs / 4 agents)
node scripts/check-frontmatter.js → passed; depends_on resolves against docs/specs/done/
npm test                         → ℹ tests 1897 / ℹ pass 1888 / ℹ fail 0 / ℹ skipped 9
git diff --stat                  → 1 file changed (this spec only)
```

`status: Ready` unchanged.

### DIFF
--- a/docs/specs/WP-scheduler-node-path-durability.md
+++ b/docs/specs/WP-scheduler-node-path-durability.md
@@ -416,17 +416,59 @@ register — and `loaded` is gated on **both**:
 At `5f0ffc0` `loaded` was `enableOk` alone and the reload was best-effort and
 ungated, which is exactly what made old Table C row 4 silent. **That is fixed.**
 
-### 9. **NEW (2026-08-02) — the one thing §8's convergence rests on that is NOT executed**
+### 9. **NEW (2026-08-02) — the premise §8's convergence rests on, and the experiment that SETTLED it**
 
 §8's convergence argument assumes launchd's `launchctl print` **echoes
 `ProgramArguments[0]` verbatim** into both the `arguments` block and the `program`
 line — i.e. that it does **not** realpath-resolve the program path at load. This
 WP registers a **symlinked** node path for the first time, so it is the first
 change that can tell the difference, and the assumption was never load-bearing
-before it.
+before it. **Table G is canonical; this section records what was run.**
 
-**What IS executed.** The sibling captured a real `launchctl print` dump on this
-same host (`docs/specs/done/WP-scheduler-register-replaces-loaded-record.md:280-293`):
+> **SETTLED 2026-08-02 — P is TRUE, by direct experiment.** Executed at the
+> **orchestrator** layer on the owner's machine (macOS 26, launchd, uid 501), not
+> by the spec-authoring pass — see "what could not be run here" below for why the
+> split matters.
+>
+> **Method.** A throwaway plist with `Label ai.wienerdog.premise-check` and
+> `ProgramArguments = ["/opt/homebrew/opt/node/bin/node", "-e", "process.exit(0)"]`.
+> **The discriminating feature is the program path**: `/opt/homebrew/opt/node` is a
+> **symlinked directory** (→ `../Cellar/node/25.9.0_2`), so the literal path
+> `/opt/homebrew/opt/node/bin/node` and its realpath
+> `/opt/homebrew/Cellar/node/25.9.0_2/bin/node` **differ**. A verbatim echo and a
+> resolving echo predict different strings; the live `dream` record could not
+> distinguish them because its `ProgramArguments[0]` is already its own realpath.
+> `plutil -lint` OK; `launchctl bootstrap gui/501 <plist>` exit 0.
+>
+> **Readback** — `launchctl print gui/501/ai.wienerdog.premise-check`:
+>
+> ```
+> path = <the scratch plist path>
+> program = /opt/homebrew/opt/node/bin/node
+> arguments = { /opt/homebrew/opt/node/bin/node, -e, process.exit(0) }
+> ```
+>
+> **Result: BOTH `program` and the `arguments` block echo the alias VERBATIM.**
+> launchd does **not** realpath-resolve the program path at load. **P is TRUE**,
+> and Table C row 5 / Table D-a hold exactly as written.
+>
+> **Teardown — zero residue.** `launchctl bootout` exit 0; post-`bootout`
+> `launchctl print` exit **113** (absent); the scratch plist deleted. **The real
+> `ai.wienerdog.dream` label was untouched throughout** — the experiment used its
+> own label and never went near it.
+
+**What could NOT be run here, and why that is still the right boundary.** A
+spec-authoring pass may read the machine but not mutate it: bootstrapping even a
+throwaway label changes the user's launchd state, and an architect does not do that
+unilaterally. That constraint is unchanged and is why this section originally
+recorded an *unverified premise* instead of guessing. It was lifted by escalating
+the experiment to the layer that could authorize it, with a scratch label and full
+teardown — **not** by relaxing the rule.
+
+**The corroborating evidence that was already executed**, and which the experiment
+promoted from suggestive to confirmed. The sibling captured a real `launchctl print`
+dump on this same host
+(`docs/specs/done/WP-scheduler-register-replaces-loaded-record.md:280-293`):
 
 ```
 gui/501/ai.wienerdog.dream = {
@@ -444,18 +486,20 @@ the loaded record *"stays correct"* and `launchctl print` *"still exits 0"*, whi
 the fire dies in `posix_spawn` — the evidence points strongly at **store-verbatim,
 resolve-at-spawn**.
 
-**What is NOT executed, and why it could not be here.** The live entry's
+**Why this dump alone was not enough** *(the reasoning that motivated the
+experiment, retained)*. The live entry's
 `ProgramArguments[0]` is already the realpath, so this dump cannot distinguish
 "echoed verbatim" from "resolved, and the two happen to be equal". Settling it
-requires either a loaded job whose program path is a symlink, or bootstrapping a
-throwaway label — a **mutation of the user's launchd state**, which is out of
-bounds for a spec-authoring pass and is not something an architect may do
-unilaterally. It is therefore recorded as an **unverified premise**, in the same
-form and for the same reason as this spec's Windows layout claim (Implementation
-notes §"Windows", WP-114 Residual 2 precedent): **specified, not observed**.
+required either a loaded job whose program path is a symlink, or bootstrapping a
+throwaway label. **The scratch-label experiment above supplied exactly that missing
+discrimination**, which is why the premise moved from *specified, not observed* to
+**observed and true** rather than staying a residual like the Windows layout claim
+(Implementation notes §"Windows", WP-114 Residual 2 precedent — **that one is still
+unverified; only this premise was settled**).
 
-**Table G is the canonical statement of the premise, its consequence if false, and
-the one command that settles it.** Definition of done item 9 is the gate.
+**Table G is canonical** for the premise, its executed result, and the retained
+history of the branch that did not occur. Definition of done item 9 is now a
+**live smoke test**, not a premise gate.
 
 ## Deliverables (permission boundary — touch ONLY these)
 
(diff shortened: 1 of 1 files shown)


thegyula:
## Update — Table G's premise is SETTLED (P is TRUE). Pushed as `3816cee`.

The open question this PR flagged is closed. The coordinator ran the distinguishing experiment the spec-authoring pass could not (orchestrator-run, owner's machine, scratch label, full teardown), **pre-implementation**.

**Why the earlier evidence was inconclusive and this is not.** The live `dream` record's `ProgramArguments[0]` is already its own realpath, so it could not distinguish "echoed verbatim" from "resolved, and equal by coincidence". The experiment supplied exactly that discrimination: a scratch label `ai.wienerdog.premise-check` whose `ProgramArguments[0]` is `/opt/homebrew/opt/node/bin/node` — a path **through a symlinked directory** (`/opt/homebrew/opt/node` → `../Cellar/node/25.9.0_2`) whose realpath is `/opt/homebrew/Cellar/node/25.9.0_2/bin/node`. **Literal ≠ resolved**, so a verbatim echo and a resolving echo predict different strings.

**Readback** (`launchctl print gui/501/ai.wienerdog.premise-check`):

```
path = <the scratch plist path>
program = /opt/homebrew/opt/node/bin/node
arguments = { /opt/homebrew/opt/node/bin/node, -e, process.exit(0) }
```

**Both `program` and the `arguments` block echo the alias verbatim — no realpath resolution at load. P is TRUE.** Teardown: `bootout` exit 0, post-`bootout` `print` exit 113 (absent), plist deleted; the real `ai.wienerdog.dream` label untouched throughout.

### What changed in the spec

- **Table G** — Status `NOT EXECUTED` → **`EXECUTED 2026-08-02 — P is TRUE`**, with the method, the result, and a new row naming what makes the evidence conclusive. The **"If P is FALSE"** and **"Consequence if P is false"** rows are **retained as history**, marked *DID NOT OCCUR* / *MOOT* — the routed `darwinLoadedVerdict` follow-up stays on the record as the correct response should launchd behavior ever change, but **no such WP is needed**.
- **Current state §9** — records the run, and keeps the boundary note: a spec-authoring pass still may not mutate launchd state. The constraint was lifted by **escalating** the experiment to the layer that could authorize it, not by relaxing the rule.
- **Table C preamble** — row 5's convergence is now an **observed outcome, not a conditional one**.
- **Definition of done item 9** — reframed from **premise gate** to **live smoke test**. The superseded "Cellar path printed ⇒ do not merge" branch is retained. The post-`sync` check is still worth running: it confirms end-to-end wiring (`entryNodePath` → renderer → `expect.argv[0]` → verified skip on the second `sync`), not the premise.
- **Mirrors updated in the same pass** per the checklist: AC6's non-claim bullet, AC9 sub-item (iv), DoD item 7, the Convergence heading paragraph, and three Mirrored Surface Checklist entries — all of which previously said "unverified".

Every superseded sentence is retained inline per the history-retained convention; nothing was rewritten in place.

**One unchanged caveat:** `npm test` still cannot substitute for the smoke test. `fakeLaunchd` renders `program: argv[0] || ''` from the plist it stored, so it models P true **by construction**. P is now known true **from the real system** — the spec says to cite Current state §9, never the suite.

**No open design questions remain in this spec.** `npm run lint` green; `status: Ready`; one file touched.

thegyula:
## Light-gate blocker fixed — `d651c43`

Both cells verified wrong in code before touching them, and both corrected under Table D's "byte-exact from source at `1093e51`" standard.

### (a) D-c's `plutil` claim was false

`schedule.js:194-211` confirms the ordering: the preflight is at **`:201-203`**, the teardown guard at **`:208`**. So the `'indeterminate'` path runs `print` → `bootstrap` (refused) → **`/usr/bin/plutil -lint`** → `return false`. Only the `bootout` half of the old cell was right.

New cell:

> `print`; `bootstrap` (refused); `/usr/bin/plutil -lint` (**runs** — the preflight at `:201-203` precedes the teardown guard; skipped only when `/usr/bin/plutil` is absent, since it is gated on `fs.existsSync(PLUTIL)`). **No `bootout`** — `:208` returns before it

As the reviewer noted, this also reconciles D-c with D-a's own first-`sync` row, which sequenced the lint correctly all along.

### (b) D-a's steady-state cell under-counted

Confirmed at `schedule.js:646-647`: `const cu = ensureCatchup(...)` then `loaded: loaded && cu.loaded`, and `ensureCatchup` does its own readback at `:495`.

I took the **"assert per helper"** option and made the scope explicit rather than just swapping the number, because D-a's other rows list *call sequences* that are genuinely per-helper — renumbering only the steady-state row would have left the table internally inconsistent. D-a now carries a call-count scoping paragraph quoting the sibling's canonical rule verbatim (`done/WP-scheduler-register-replaces-loaded-record.md:1881-1883`), the OS-calls column header is scoped to the per-job helper, and the steady-state cell gives both figures: **one `print` per helper, two through the full `sync` path, zero mutations either way**. D-c inherits the same scoping.

### Two more mirrors the gate did not flag

Swept for the same under-count and found it twice more — both corrected in this commit:

- Table D's **RECONCILED preamble** bullet ("A later `sync` makes one read-only `print`…")
- **Table G's "If P is TRUE"** row ("Steady state after one `sync`: … one read-only `print`…")

### Mirror discipline

**AC9 now requires the scoping paragraph and both dated correction notes to travel with the tables** — reproducing the call lists without them restates the exact *"'exactly one' through the full path"* claim the sibling forbids, turning a correct table into an incorrect one in the merge artifact. Registered as a new **(+r6)** Mirrored Surface Checklist entry covering all six surfaces, with a note that the full-path count is the **sibling's** canonical property and must be derived from it, never restated independently.

Both dated correction notes are inline per the history-retained convention; the superseded text is quoted, not deleted.

`npm run lint` green; `status: Ready`; one file touched.

PR #148, opened 2026-08-02
docs(specs): two Ready follow-up WPs — symlink row-3 narrowing + gws-broker/threat-model destale

thegyula:
## What

Two routed follow-ups from the 2026-08-02 wave, both `Ready`.

**`WP-symlink-lexical-fallback-removal`** (S) — drops row 3's link-text sub-test from `reverseSymlink` in `src/core/manifest.js`. Not dead-code removal: it is an intentional, strictly-narrowing behaviour change. Ships three tests (T4a/T4b/T4c) and twelve byte-exact amendments to the Done `WP-153`, all in one commit.

**`WP-gws-broker-runbook-destale`** (S) — three prose regions across two files: the broker runbook's gate-status banner and its live-proof harness claim, plus one clause of `docs/THREAT-MODEL.md:146` that carried the same defect.

## Design gate

Seven Codex design-gate rounds, terminating in APPROVE with no material findings. Arc: **R1** recharacterized the change after a reachable counterexample (relative recorded target, `realpath` resolves it against cwd) disproved the "dead branch" premise; **R2** pulled four unregistered `T4` mirrors into lockstep; **R3–R5** hit three successive lexical evasions of the verification gate (raw grep → block comments → `//` inside a string literal plus `if (false)` nesting), which tripped the ADR-0031 loop circuit-breaker and forced an **extraction pass**: every gate is now a whole-artifact reconstruction diff rather than a token check; **R5** then found the reconstruction's own trust root could move with the implementation commit, closed by **V0** (spec reconstructed from branch base, licensing only `Ready` → `In-Review`); **R6** applied the same V0 to the sibling spec and corrected AC6 to count the mandatory status flip. Every gate was proven red-and-green against real fixtures, with the spec's verbatim text.

## Declared residuals (all recorded in-spec)

- **Relative-target preserve ruling** — an untrusted manifest field may only narrow deletion. The governing rule is in no ADR and is flagged for the owner (see "Owner items" below).
- **Upstream-realpath TOCTOU race** — pre-existing, ADR-0028/WP-159 class; outcome moves delete → preserve.
- **Exported-helper divergence** for direct callers of `reverseSymlink`.
- **Comment text inside `reverseSymlink` is now contract** — the accepted price of a byte-exact gate.

## Owner items

- The rule *"an untrusted manifest field may only narrow deletion"* is load-bearing in `WP-symlink-lexical-fallback-removal` (and in the parallel Part A/B provenance specs) but written in no ADR. Architect recommends promoting it to an ADR; flagged four times across the gate rounds, unchallenged in seven. Decision is Gyula's.

## Verification

- `npm run lint` exit 0 at `c1b6a62` (388 files, 0 errors; frontmatter 211 specs).
- `scripts/boundary-check.js` exit 0 for both specs' deliverable sets.
- wd-reviewer leg to follow in PR comments before merge.

Generated-by: Claude Fable 5 (claude-fable-5)

### DIFF
--- a/docs/specs/WP-gws-broker-runbook-destale.md
+++ b/docs/specs/WP-gws-broker-runbook-destale.md
@@ -46,10 +46,16 @@ poisoned-email test harness — a claim about the harness's **current execution
 status**, which a `Ready` WP (`WP-broker-e2e-terminal-auth`) and the harness's own
 `AUTH-BLOCKED` short-circuit contradict today (Current state §3). Both defects are
 the same failure mode — **prose asserting a status a later change can falsify** —
-so both are fixed here, in the one file, by the same rule.
+so both are fixed here, by the same rule.
 
-This WP changes prose in one file. It adds no behavior, no file, no flag and no
-process (**IRON RULE, ADR-0004: Wienerdog is just files** — nothing here starts
+**And the same defect turned out to be in the document the runbook defers to.**
+Round 2 of the design gate found `docs/THREAT-MODEL.md:146` asserting that the
+harness "is being re-fitted" and naming a WP that is already `Done`. That clause
+is **E3**. The scope is therefore **three regions across two files** — resolve it
+from **Table S**, never from prose.
+
+This WP changes prose only. It adds no behavior, no file, no flag and no process
+(**IRON RULE, ADR-0004: Wienerdog is just files** — nothing here starts
 anything). It is user-facing text for knowledge workers, not developers: plain
 language, no jargon without explanation.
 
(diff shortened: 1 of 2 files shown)


thegyula:
wd-reviewer leg (round 1): **REQUEST-CHANGES** at c1b6a62. 1 blocks-Ready: the WP-153 post-merge-note anchor census is wrong (:641 has no link; :806 — R7's own replaced line — is missed; after the mandated edits only four of the claimed five links survive, so AC6 is unsatisfiable). 7 should-fix (stale scan results, a five-not-four reference count, manifest.json vs GLOSSARY-canonical install-manifest.json, a split Deliverables table, bare-git-diff gates that are empty post-commit, a T4-census contradiction with R9, AC6-vs-R9 bare-T4 conflict). 9 advisory. Reviewer recommends an ADR-0031 canonical-extraction for the WP-153 mirror census (anchor line / opening words / live-or-dated / carries-link / R-item) — same contract family Codex rounds 1 and 4 reworked. All base facts, V0 both directions, R8-vs-live-function byte checks, and spec 2's full dry-run (E1+E2+E3 applied, all gates green, reverted) were executed and confirmed. Full verdict retained in session record; findings routed to wd-architect. Codex design leg previously terminated APPROVE at 935e6c2.

thegyula:
Double gate CLOSED at 0ea4f70. **Codex: approve** (final delta round — occurrence-level V3a2 verified against all five fixtures, V0/R8 heredoc unchanged; earlier design leg terminated approve at 935e6c2 after seven rounds). **wd-reviewer: APPROVE stands** (re-ran the shipped V3a2 verbatim on all five fixtures incl. the same-line adversarial case on macOS/BSD grep; confirmed the old form was green on that fixture — the hole was real; spec 2 blob-hash-identical since 445b3d8; V0 both specs green; lint + boundary green). Reviewer lesson for the logbook: a gate carrying a whitelist needs a fixture where the whitelisted and forbidden tokens co-occur in the filter's unit of operation. Merging under session authorization.

PR #154, opened 2026-08-02
fix(uninstall): prove a managed-block separator's position before stripping it (WP-managed-block-insertion-anchor)

thegyula:
Spec: docs/specs/WP-managed-block-insertion-anchor.md

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table
- [x] Spec status flipped to In-Review

Changed files vs base `17a2bc5`: `src/core/manifest.js` (D1, D2, D6a),
`src/adapters/shared.js` (D7a, D8, D9), `tests/unit/manifest.test.js`
(A-T1…A-T11, incl. the Table F edit to WP-147 T9), plus the spec's own
status flip (always allowed; only the `status:` line changed). Nothing else —
`git diff --stat 17a2bc5` shows exactly these four files.

Worktree base verified before starting: `git rev-parse HEAD` = `17a2bc5` (the
dispatch SHA).

## Verification output

**Red baseline first (tests applied, src unmodified)** — per the Test index's
"Red against" column, exactly the predicted set went red; the baseline rows
A-T2/A-T4/A-T9/A-T10 stayed green against base (they pin base-equal behaviour):

```
✖ WP-147 T9 (Table M bound): … ('foo\n\n' !== 'foo' — Table F's measured flip, A-T5's edit)
✖ WP-managed-block-insertion-anchor A-T1 (Q1)
✖ WP-managed-block-insertion-anchor A-T3 (Q5, R2)
✖ WP-managed-block-insertion-anchor A-T6 (Q10)
✖ WP-managed-block-insertion-anchor A-T7 (Q12)   (the relocation arms)
✖ WP-managed-block-insertion-anchor A-T8 (Table B :179/:197)
✖ WP-managed-block-insertion-anchor A-T11 (Q16, R2b)
ℹ tests 91  ℹ pass 83  ℹ fail 7
```

**V1 — full suite (post-implementation):** baseline at `17a2bc5` was
`tests 1903 / pass 1894 / fail 0 / skipped 9`; after the WP:

```
ℹ tests 1913
ℹ suites 0
ℹ pass 1904
ℹ fail 0
ℹ cancelled 0
ℹ skipped 9
ℹ todo 0
```

(+10 tests / +10 pass — the ten new A-T blocks; A-T5 is an edit, not a new test.)

**V2 — targeted:**

```
node tests/run.js tests/unit/manifest.test.js
ℹ tests 91  ℹ pass 90  ℹ fail 0
```

[...]

PR #156, opened 2026-08-03
docs(specs): pre-dispatch reconciliation of WP-symlink-authorship-identity against the landed Part A (WP-symlink-authorship-identity)

thegyula:
## What

Part B's dispatch-time re-verification blocked dispatch: Part A's merge (#154) moved `src/core/manifest.js` (1062 → 1122 lines) and `src/adapters/shared.js` under the spec. Reconciliation only — no design, ledger row, or ruled disposition changed (41 of 74 changed lines differ only by a number or SHA; the rest are the four restated sections + provenance).

- Precondition proven before re-anchoring: `reverseSymlink`'s body is byte-identical at `9188a1c` ≡ `17a2bc5` ≡ `8515eb1` (cmp). Every anchor re-derived by content at `8515eb1`.
- Shared hunks restated as half-shipped: the typedef/doc-comment already carry `anchorBefore` and `shared.js:5` already imports `insertionAnchor` — D6b/D7b extend shipped bytes; V8's base direction re-aimed accordingly (all four names kept so Part A's hunk retains a guard).
- V4 re-pinned to `8515eb1` and re-run from the spec: self-check green, red arms fire, V4z 31/14.
- Dispatch precondition updated: Part A is Done (#154/#155). Suite baseline re-anchored (1913/1904/0), Part A's already-landed T9 flip called out so it isn't double-counted.
- One over-replacement self-caught: `shared.js:406` inside the WP-153 stale-claim note is a quotation of the wrong text, not an assertion — restored.

Precedent: PR #144 (node-path pre-dispatch reconciliation).

## Verification

Lint green (213 specs); suite 1913/1904/0; boundary clean (one spec file).

Generated-by: Claude Fable 5 (claude-fable-5)

### DIFF
--- a/docs/specs/WP-symlink-authorship-identity.md
+++ b/docs/specs/WP-symlink-authorship-identity.md
@@ -1496,120 +1496,188 @@ node /tmp/wd-docfields.js src/core/manifest.js origin dev ino anchorBefore && ec
 #   **Re-run this instead of sweeping.** When it fails, fix the citation — or, if
 #   the code genuinely moved, fix the anchor and let it re-resolve.
 cat > /tmp/wd-citescan.js <<'CS'
-// V11 — CITATION SCAN. Resolves every src/ line number this spec is allowed to
-// cite BY CONTENT against the tree, then flags any src/ citation in the spec that
-// is not on that list. Ships in the spec so the next reconciliation re-runs it
-// instead of sweeping by hand.
+// V11 — CITATION SCAN. Resolves every src/ line number this spec may cite BY
+// CONTENT, per file, and fails on any citation that does not resolve against the
+// file it is attributed to.
+//
+// THREE DESIGN ROUNDS, each closing a demonstrated false negative:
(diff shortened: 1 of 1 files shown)


thegyula:
Double gate CLOSED at 3a13066. **wd-reviewer: APPROVE** (V11 table reproduced; four controls + an independent inverse-direction per-file probe; ambiguity failure verified live; zero design change confirmed hunk-by-hunk). **Codex: approve** (qualified slash-group first-matched, its shared.js:439/:216/:496 case now fails at the right coordinate; five controls reproduce; no precedence regression). Arc: the reconciliation itself took four gate rounds — a line-scoped sweep missed 11 mirrors + 3 fence-comment coordinates, and the resulting V11 citation gate went through four measured design rounds (global-number sets, inert per-file map caught only by a control, grouped-tail qualification) before reaching per-file content-anchored resolution with five permanent red controls and a declared unqualified residual. Part B is now DISPATCHABLE. Merging under session authorization.

PR #175, opened 2026-08-30
fix(launcher): bound the refusal alert log (WP-launcher-alert-bound)

thegyula:
Spec: docs/specs/WP-launcher-alert-bound.md

## Deliverables checklist

Files changed (6 deliverables + the allowed spec status flip):

| File | What changed |
|------|--------------|
| `src/core/alerts.js` | new exported `MAX_COUNT`; `sanitizeAlert` returns a fail-closed `count`; `Alert` typedef |
| `src/core/digest.js` | `formatAlerts` sums `count` (fail-closed coercion + clamped sum) |
| `src/scheduler/launcher.js` | `appendRefuseAlert` collapse + bound; new `readAlertRecords` helper; the three duplicated literals; `appendRefuseAlert` exported for tests |
| `tests/unit/alerts.test.js` | `count` coercion/clamp/round-trip/backward-compat; two pre-existing `deepEqual` shape assertions updated for the new field |
| `tests/unit/digest.test.js` | `formatAlerts` summing, single-record wording, clamp, per-job sums, golden byte-identity |
| `tests/unit/launcher.test.js` | collapse, non-collapse, bounds, durability, failure swallowing, builtin-only requires |
| `docs/specs/WP-launcher-alert-bound.md` | `status: Draft` → `In-Review` |

`tests/golden/digest-default.md` is untouched (AC-6).

## Verification output

```
$ npm test -- --test-name-pattern alerts
ℹ tests 132
ℹ suites 0
ℹ pass 132
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1912.753667

$ npm test -- --test-name-pattern digest
ℹ tests 152
ℹ suites 0
ℹ pass 152
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1952.759958

$ npm test -- --test-name-pattern launcher
ℹ tests 156
ℹ suites 0
ℹ pass 156
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1907.550208

$ npm test
ℹ tests 1969
ℹ suites 0
ℹ pass 1960
ℹ fail 0
ℹ cancelled 0
ℹ skipped 9
ℹ todo 0
ℹ duration_ms 45948.513417

$ npm run lint

> wienerdog@0.13.0 lint
> node scripts/lint.js

--- markdownlint ---
Linting: 402 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 221 spec(s), 4 agent(s)

lint passed

# C11 — the launcher still requires no app-tree code (expect NO output):
$ grep -n "require(['\"]\.\./src\|require(['\"]\.\./\.\./src" src/scheduler/launcher.js
(no output)

# C3/C4/C5 — the duplicated bounds are present in both places:
$ grep -n "MAX_ALERTS\|MAX_FILE_BYTES\|MAX_COUNT" src/core/alerts.js
28:const MAX_ALERTS = 200; // keep only the most-recent N records
30:const MAX_FILE_BYTES = 512 * 1024; // hard byte bound on the log file / the read
33:// it reaches the digest banner's arithmetic. MAX_ALERTS, MAX_FILE_BYTES and
34:// MAX_COUNT are duplicated as literals in src/scheduler/launcher.js, which cannot
36:const MAX_COUNT = 1000000;
57: *  MAX_COUNT, so a hand-edited alerts.jsonl cannot push a nonsense or memory-hostile
68:  const count = Number.isSafeInteger(n) && n >= 1 ? Math.min(n, MAX_COUNT) : 1;
73: *  Compacts to `MAX_ALERTS` records / `MAX_FILE_BYTES` bytes when either bound
126:  if (all.length > MAX_ALERTS || size > MAX_FILE_BYTES) {
128:    let kept = all.slice(Math.max(0, all.length - MAX_ALERTS)); // newest N (append order = chronological)
132:    // well under MAX_FILE_BYTES (4 fields × MAX_FIELD_CHARS ≈ 8 KiB + JSON overhead).
133:    while (kept.length > 1 && Buffer.byteLength(text) > MAX_FILE_BYTES) {
145: *  Byte-bounds its read (tail window of `MAX_FILE_BYTES` for oversized files) and
163:    if (st.size > MAX_FILE_BYTES) {
164:      // Read the trailing MAX_FILE_BYTES (newest records) PLUS one preceding byte, so
167:      const readStart = st.size - MAX_FILE_BYTES - 1; // >= 0 since st.size > MAX_FILE_BYTES
236:  MAX_ALERTS,
238:  MAX_FILE_BYTES,
239:  MAX_COUNT,

$ grep -n "200\|512 \* 1024\|1000000\|1_000_000" src/scheduler/launcher.js
173:const ALERT_MAX_RECORDS = 200; // twin: MAX_ALERTS in src/core/alerts.js
174:const ALERT_MAX_FILE_BYTES = 512 * 1024; // twin: MAX_FILE_BYTES in src/core/alerts.js
175:const ALERT_MAX_COUNT = 1000000; // twin: MAX_COUNT in src/core/alerts.js
243: *  newest-200 compaction. Two repairs, together because either alone is lossy:

# AC-6 — the frozen golden is untouched (expect NO output):
$ git diff --stat -- tests/golden/digest-default.md
(no output)
```

## Known open finding (C8 race — flagged by the Codex design review, NOT fixed here)

The Codex review of the spec set flagged Table C row C8 (append-then-rewrite
compaction) as racy across two concurrent launcher processes:

> P1 appends + reads snapshot A, P2 appends + reads snapshot B and renames,
> then P1 renames its older snapshot A over the file and B is lost.

What this branch's write path does in that interleaving, stated plainly:
`appendRefuseAlert` appends its own record atomically first (so P1's record is
in both snapshots by the time either rename lands), reads the file back, and
then renames its snapshot over the log with **no** compare-and-retry and no
cross-process lock. So in the interleaving above P2's record is indeed lost —
P1's older snapshot A wins, and P2's alert is gone even though P2's own
`process.exit(1)` and zero-spawn refusal still stand.

This is the same residual `appendAlert` has carried since it was written (its
own comment: *"A compaction by one run-job can still drop a record a DIFFERENT
run-job appended in the same window; that residual is accepted — full
cross-process locking is out of scope per ADR-0004"*), so this WP does not
widen it — but it does not narrow it either, and the launcher's failure mode
(scheduled jobs that can fire concurrently) makes the window more reachable
than the app side's. The spec may be revised (cross-process lock, or
compare-and-retry before the rename) before this merges; I implemented C8
exactly as it stands today and did not redesign. **Draft PR for that reason.**

## Decisions made

1. **`appendRefuseAlert` is now exported** from `src/scheduler/launcher.js`.
   The bound and collapse behaviour (AC-7 … AC-11, AC-13) cannot be exercised
   through `main()` at reasonable cost — AC-10 alone needs 225 refusals, each
   of which would otherwise require a full drifted `setupProd` install. The
   export is test-only surface and changes no verification logic.
2. **The collapse is implemented as "compare the just-appended last record
   against its immediate predecessor, then drop the appended row and increment
   the predecessor's `count`."** C7 is written from the pre-append point of
   view ("increment that last record's count"), but C8 mandates appending
   first; the two are reconciled the only way that preserves both — the
   predecessor keeps its original `at` (C7) and the file never loses a record
   it did not just write (C8).
3. **The launcher's read-back drops the first line whenever its tail window
   starts mid-file**, rather than duplicating `readAlerts`' preceding-byte
   check to tell a boundary-aligned window from a split one. The simpler rule
   can discard at most one *oldest* record, which the byte bound (C10) was
   about to drop anyway. The reader stays a ~40-line self-contained twin
   instead of a ~60-line one.
4. **`formatAlerts` is more than the "exactly one line" the spec's Exact
   contracts predicted.** The security checklist and AC-5 require the summed
   total to be clamped, and callers legitimately pass records that never went
   through `sanitizeAlert` (every existing `digest.test.js` alert literal does
   — none of them has a `count`). A local fail-closed `countOf` mirroring C2
   is therefore applied before the arithmetic; without it those records would
   sum to `NaN`. Wording, `first`/`lastReason`/`hint` handling and the
   single-record output are all unchanged.
5. **Two pre-existing assertions in `tests/unit/alerts.test.js` were updated**
   (`deepEqual(got[0], {job, at, reason, log_hint})` → the same object plus
   `count: 1`). They pin `sanitizeAlert`'s exact output shape, which C1
   changes by design; the file is in the Deliverables table.
6. **The launcher's constants are named `ALERT_MAX_RECORDS` /
   `ALERT_MAX_FILE_BYTES` / `ALERT_MAX_COUNT`**, not the app-side names, so a
   reader cannot mistake them for imports. Each carries a `// twin: <name> in
   src/core/alerts.js` comment as C3–C5 require.
7. **`appendAlert` is unchanged apart from its JSDoc.** It writes `count: 1`
   for free, because `sanitizeAlert` supplies it — no collapse was added to
   the app-side writer (explicitly out of scope).

1. **`clearAlerts` does not re-`chmod` after its rename.** `appendAlert` calls
   `chmodAlerts` after its compaction rename and `clearAlerts` writes its temp
   file with the default mode and no `{ mode: 0o600 }`, so a `clearAlerts`
   rewrite can leave `alerts.jsonl` at the umask default rather than 0600.
   Pre-existing, unrelated to this WP, not touched.
2. **The byte-budget loop in both writers is O(n²)** — it re-serializes every
   surviving record on each drop. Harmless at `MAX_ALERTS = 200`; noted only
   because this WP copies the shape into a second place.
3. **`appendAlert`'s comment says "4 fields × MAX_FIELD_CHARS ≈ 8 KiB"**,
   which is still true of the string fields after this WP (`count` adds ~12
   bytes), so it was left alone rather than churned.

## Lessons

- WP-launcher-alert-bound: the spec's "changes exactly one line" prediction for
  `formatAlerts` conflicted with its own security checklist (clamp the sum) and
  with the existing tests, which pass alert literals that never see
  `sanitizeAlert`. Trust the acceptance criteria and the checklist over a
  prose estimate of diff size — and check what the *existing* tests feed a
  function before assuming every caller is the sanitized path.
- WP-launcher-alert-bound: C7 ("increment the last record's count") and C8
  ("append first, then rewrite") describe the same operation from opposite
  sides of the append. Reading either alone yields a different implementation;
  the reconciliation (drop the appended row, increment the predecessor) is
  worth stating in a spec that mandates both.
- WP-launcher-alert-bound: a duplicated-constant rule (C3–C5, ADR-0031) is much
  easier to keep honest when the copies are *deliberately named differently*
  and each carries a `// twin:` comment — the grep in the spec's verification
  block then reads as a real cross-check rather than a coincidence of literals.
- WP-launcher-alert-bound: testing a best-effort writer's failure swallowing
  needs two different unwritable shapes — a `0o500` state dir (the append never
  lands) and a `0o200` log file (the append lands but the read-back fails, the
  C9 guard). Only the second one proves the empty-read guard is doing anything.

---
Generated-by: claude-opus-5

https://claude.ai/code/session_01U9LjCEYVM481sJirm1HDY1

### DIFF
--- a/docs/specs/WP-launcher-alert-bound.md
+++ b/docs/specs/WP-launcher-alert-bound.md
@@ -1,7 +1,7 @@
 ---
 id: WP-launcher-alert-bound
 title: Bound the launcher's alert writer without understating a real failure streak — collapse consecutive identical refusals into a count field
-status: Draft
+status: In-Review
 model: opus
 size: M
 depends_on: []
--- a/src/core/alerts.js
+++ b/src/core/alerts.js
@@ -28,32 +28,52 @@ const ALERTS_FILE = 'alerts.jsonl';
 const MAX_ALERTS = 200; // keep only the most-recent N records
 const MAX_FIELD_CHARS = 2000; // cap each string field (control-plane text, not prose)
 const MAX_FILE_BYTES = 512 * 1024; // hard byte bound on the log file / the read
+// Ceiling on a record's `count` — the streak length a collapsed launcher refusal
+// carries (WP-launcher-alert-bound, Table C3). Bounds a parsed-input number before
+// it reaches the digest banner's arithmetic. MAX_ALERTS, MAX_FILE_BYTES and
+// MAX_COUNT are duplicated as literals in src/scheduler/launcher.js, which cannot
+// require this module (Table C11) — update the two copies together.
+const MAX_COUNT = 1000000;
 
 /** @param {import('./paths').WienerdogPaths} paths @returns {string} */
 function alertsPath(paths) {
   return path.join(paths.state, ALERTS_FILE);
 }
 
-/** Coerce a record to the known string fields, each length-capped and then
- *  secret-scrubbed (EP3, audit A5 / ADR-0024 / WP-124): the cap bounds the scan
- *  input, then `redactOnly` guarantees no secret persists to alerts.jsonl or
- *  reaches the digest — `at`/`job`/`log_hint` are code-owned no-ops, but
- *  scanning uniformly is the fail-closed choice. Requires a non-null,
+/** The alert record.
+ *  @typedef {{job:string, at:string, reason:string, log_hint:string, count:number}} Alert
+ *  `count` — how many consecutive identical (job, reason) refusals this row
+ *  represents (Table C1). ALWAYS present after sanitizeAlert; >= 1; an
+ *  absent/invalid input value becomes 1. Only the launcher's writer ever
+ *  collapses, so appendAlert always records 1. */
+
+/** Coerce a record to the known fields: the four string fields, each length-capped
+ *  and then secret-scrubbed (EP3, audit A5 / ADR-0024 / WP-124) — the cap bounds the
+ *  scan input, then `redactOnly` guarantees no secret persists to alerts.jsonl or
+ *  reaches the digest (`at`/`job`/`log_hint` are code-owned no-ops, but scanning
+ *  uniformly is the fail-closed choice) — plus the numeric `count`, coerced FAIL
+ *  CLOSED to 1 (Table C2): anything that is not a safe integer >= 1 (absent, 0, -1,
+ *  1.5, NaN, Infinity, 1e309, an object) becomes 1, and anything larger is clamped to
+ *  MAX_COUNT, so a hand-edited alerts.jsonl cannot push a nonsense or memory-hostile
+ *  number into the digest banner's arithmetic. A pre-WP record with no `count` reads
+ *  as 1, so no migration is needed (Table C13). Requires a non-null,
  *  non-array OBJECT — any other value (null, number, string, array) is
  *  treated as an empty object, so a valid-JSON primitive can't crash the deref.
- *  Drops unknown keys; missing fields become ''.
- *  @param {*} r @returns {{job:string, at:string, reason:string, log_hint:string}} */
+ *  Drops unknown keys; missing string fields become ''.
+ *  @param {*} r @returns {Alert} */
 function sanitizeAlert(r) {
   const o = r && typeof r === 'object' && !Array.isArray(r) ? r : {};
   const scrub = (v) => redactOnly(String(v == null ? '' : v).slice(0, MAX_FIELD_CHARS));
-  return { job: scrub(o.job), at: scrub(o.at), reason: scrub(o.reason), log_hint: scrub(o.log_hint) };
+  const n = Number(o.count);
+  const count = Number.isSafeInteger(n) && n >= 1 ? Math.min(n, MAX_COUNT) : 1;
+  return { job: scrub(o.job), at: scrub(o.at), reason: scrub(o.reason), log_hint: scrub(o.log_hint), count };
 }
 
 /** Append one unresolved failure alert (atomic append; creates state/ if needed).
  *  Compacts to `MAX_ALERTS` records / `MAX_FILE_BYTES` bytes when either bound
  *  is exceeded after the append.
  *  @param {import('./paths').WienerdogPaths} paths
- *  @param {{job:string, at:string, reason:string, log_hint:string}} record */
+ *  @param {{job:string, at:string, reason:string, log_hint:string, count?:number}} record */
 function appendAlert(paths, record) {
   mkdirPrivate(paths.state); // 0700 independent of umask (audit A5, WP-126)
   const file = alertsPath(paths);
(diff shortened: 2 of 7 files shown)


thegyula:
Closed by owner ruling 2026-08-31 (see `docs/specs/logbook/2026-08-31-two-adr-0039s-and-the-chain-stands-down.md`, merged in #179): the digest-delivery chain stands down with ADR-0040's withdrawal, and this draft additionally collided with the fork's `formatAlerts` rewrite now on main. The underlying concern (bounding the launcher's alert writer) survives in the alert-bound spec archived as Superseded — re-derive against the current tree if picked up.

PR #185, opened 2026-09-01
fix(run-job): keep the durable record when a state write fails (WP-failloud-survives-state-write-failure)

thegyula:
Spec: docs/specs/WP-failloud-survives-state-write-failure.md

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table (`src/cli/run-job.js`, `tests/unit/scheduler-runjob.test.js`)

## Verification output

Round 1 (commit 1273288):

```
npm test → 2384 tests, 2372 pass, 0 fail, 12 skipped (18 new tests)
npm run lint → lint passed (markdownlint 0 issues; frontmatter 264 specs)
boundary-check → exit 0
HEAD-red / branch-green pair: run-job.js reverted → 16 of 18 new tests FAIL
  (the 2 still-green test failLoud's default/unknown-outcome cases, independent
  of the call-site guards — correct); restored, full suite green reconfirmed.
Mechanism probe: writeScheduleState still throws on unwritable state/, and
  writeFilePrivate still refuses a symlinked leaf — red by construction (this
  WP guards the call sites, not the primitives).
```

Round 2 (commit 5a2c35d, on top of the Table E spec amendment 9b788ea — F1-F5
fixes from round-1 gates: wd-reviewer APPROVE-conditional, plugin 1 P2, shadow
2 findings, converging on the amended Table E derivation/isolation contract):

```
npm test -- --test-name-pattern "run-job|runjob" → 213 tests, 213 pass, 0 fail
  (26 new tests added this round: poison-reason x3, poison-outcome x4,
  GWS-unconfigured B2 x1, plus F2/F4/F5 hardening of existing both-fail/
  single-fault tests)
npm test (full) → 2392 tests, 2380 pass, 0 fail, 12 skipped
npm run lint → lint passed (markdownlint 0 issues; frontmatter 264 specs)
boundary-check → exit 0
RED-side proof for the F1 fix specifically: reverted src/cli/run-job.js to
  commit 1273288 (pre-F1) while keeping the round-2 test file → 6 of 6 new
  poison-reason/poison-outcome tests FAIL (the 2 poison-reason tests on the
  wrong-caller-reason path, and all 4 poison-outcome tests on :797/:872/
  :1096/catch-up — each shows the poisoned alert-cleanup-refused outcome
  actually reaching the ordinary site: e.g. subject
  "job digest succeeded, but its alert cleanup failed" instead of the
  expected "job digest failed"). Restored (diffed byte-identical against the
  pre-revert copy); full targeted suite (213/213) and full suite (2380 pass)
  reconfirmed green.
```

## What changed

All five state-write sites (`:797`, `:872`, `:1096`, `:1060`, `:1061`) guarded inline; a watermark/clearAlerts throw is surfaced once on stderr and never suppresses `failLoud`. `failLoud` gains the Table E closed outcome enum (`job-failure` = byte-identical HEAD default, `success-marker-refused` = B1, `alert-cleanup-refused` = B2 with the Table C append skip); unknown discriminant throws; `persisted` honors an explicit `false` from `appendAlert` (`undefined` keeps HEAD semantics — the F10 producer lands in the mode-pin WP).

## Decisions made

1. Canonical B1/B2 reason/subject strings live in two local template functions — single source for both `failLoud`'s derivation and the call sites' thrown `WienerdogError`; the subject/append-policy seam stays strictly inside `failLoud`.
2. The non-alert stderr diagnostic is emitted on any watermark/clearAlerts throw (required unconditionally for B2; applied uniformly — no criterion conflicts).
3. `opts.appendAlert` test-only seam added, mirroring the existing `opts.sendAlert` idiom (named as acceptable by the spec).
4. (Round 2, F1b) Outcome isolation is a new `ordinaryFailLoudOpts(opts)` helper (`{ ...opts, outcome: 'job-failure' }`) applied at all seven ordinary `failLoud` call sites, rather than an inline spread at each — reduces the risk of missing one of the seven and silently reopening the suppression seam.
5. (Round 2, F2) The four both-writes-fail tests now inject a call-counting wrapper around the real `appendAlert` (delegates to it, so the directory-fixture throw still fires) rather than asserting only on the absence of a record, closing the "removing the failLoud call would still pass" gap shadow flagged.
6. (Round 2, F3) The GWS-unconfigured B2 test uses a pre-existing EMPTY DIRECTORY at `alerts.jsonl` (not the pid-tmp-dir + other-job-alert fixture used elsewhere) — `readAlerts()`'s own EISDIR catch resiliently returns `[]`, so `clearAlerts` takes its `rmSync` branch, which also throws without `recursive:true`. This gives a literal `readAlerts(paths) === []` as the acceptance criterion asked for.

- `mkdirPrivate`'s `chmodIfNeeded` auto-heals a `chmod 0500` state/ before `appendAlert` writes, defeating that technique for both-writes-fail simulation (tests use pre-created directories instead — EISDIR). Relevant to the mode-pin WP's tests.
- Pre-existing: `failLoud`'s append and email share one `try`, so an appendAlert *throw* (vs return false) skips the email too — Table A's both-writes-fail case gets zero email attempts. HEAD behavior, out of scope here.
- Stale comment in `catchUp` (`// doRun already recorded the error watermark and failed loud.`, originally at commit 1273288's `:1383` on the per-job `catch` block): its claim that the watermark is always "recorded" is no longer exact after this WP — the watermark write can throw and NOT persist while `failLoud` is still attempted. Flagged by wd-reviewer C in round 1; dispositioned as out of scope for this WP (no spec instruction to touch it), noted here rather than fixed.

## Review gates

Runs recorded in comments below (wd-reviewer + Codex plugin on the same tip).

## Lessons (for memory/lessons/inbox.md)

- WP-failloud-survives-state-write-failure: `chmod(state/, 0o500)` is not a durable both-writes-fail simulation — `mkdirPrivate` heals it before the append; pre-created directories at the target paths (EISDIR) are portable and unhealed.
- WP-failloud-survives-state-write-failure: `renameSync(tmp, dest)` replaces a symlinked dest rather than following it, so a symlinked `alerts.jsonl` alone doesn't make the pre-mode-pin rewrite throw; a directory planted at the deterministic `${file}.${pid}.tmp` name does.
- WP-failloud-survives-state-write-failure: `failLoud`'s shared try means an appendAlert throw skips the email send — easy to assume email is always attempted; it isn't.
- WP-failloud-survives-state-write-failure: a spec that says a value is "derived inside" a function but also requires the caller to throw the SAME string is unsatisfiable as written — the fix is a shared module-level template called from both sides, not a choice of which side "owns" it. Worth stating explicitly in a future spec rather than leaving it for review to catch.
- WP-failloud-survives-state-write-failure: a generic `opts` object threaded through a whole call chain (test seams, forwarded caller opts) is also a channel an attacker-shaped test/caller can use to inject an unintended key (here, `outcome`) into an unrelated call site — worth an explicit "which sites forward opts verbatim vs. construct their own" note in specs that introduce a new discriminant riding an existing opts bag.
- WP-failloud-survives-state-write-failure: a pre-existing EMPTY DIRECTORY at a target path is a cleaner way to force `readAlerts()` to literally return `[]` (its own read fails EISDIR and is caught) than pre-populating a file and blocking a rename — useful whenever a test needs "the reader sees nothing" rather than "the reader sees something it can't act on."

---
Generated-by: Claude Sonnet 5 (implementer) / Claude Fable 5 (orchestrator)

https://claude.ai/code/session_01LCXat1SiCGPBaijqCCkAjG

### DIFF
--- a/docs/specs/WP-failloud-survives-state-write-failure.md
+++ b/docs/specs/WP-failloud-survives-state-write-failure.md
@@ -243,7 +243,10 @@ values, and `failLoud` derives everything from it internally.
 | Fact / rule | Value |
 |-------------|-------|
 | Closed set | those three values and nothing else. The caller passes only the discriminant |
-| Derived inside `failLoud` | the canonical reason wording, the email subject, and the append policy. Callers may pass the job-specific detail the reason interpolates, but **never** a subject and **never** a skip boolean — there is no caller-supplied way to suppress an append |
+| Derived inside `failLoud` | the email subject and the append policy. Callers may pass the job-specific detail the reason interpolates, but **never** a subject and **never** a skip boolean — there is no caller-supplied way to suppress an append |
+| **One source for each canonical string** (amended 2026-09-01) | Table B `:173` requires `runJob` to **throw** a `WienerdogError` carrying the site's distinct reason, and this table requires `failLoud` to persist and email that same wording. A reason derived *only* inside `failLoud` cannot reach the caller's throw, so the earlier "derived inside `failLoud`" phrasing made the two rows unsatisfiable together. The resolution: each outcome's reason and subject come from **one shared module-level template**, called by `runJob` for the throw and by `failLoud` for what it records. The template is the string's single source; neither caller nor callee may compose its own variant, and the verbatim-string assertions test the template's output on both paths |
+| **`failLoud` self-derives the reason for the non-default outcomes** | for `success-marker-refused` and `alert-cleanup-refused`, `failLoud` derives the reason it persists and emails **from the outcome**, via that template — it does **not** trust the `reason` argument it was handed. A stale or wrong caller string would otherwise misreport replay: measured on this branch, `failLoud(paths, name, 'x', { outcome: 'alert-cleanup-refused' })` skips the append and emails body `x`, so the operator is told nothing true about which state operation was refused. **The outcome decides the reason, the subject and the append policy — all three.** The `reason` parameter is honored **only** for `job-failure`, which keeps HEAD semantics for the three failure sites and the catch-up refusals |
+| **The outcome channel is separate from the forwarded `opts`** | the discriminant must **not** be inheritable from the generic `opts` object `runJob` forwards to its ordinary `failLoud` calls. Those sites pass their caller's `opts` straight through, so an `outcome` riding that object reaches them too — and `alert-cleanup-refused` sets the append-skip, which is precisely the "suppression capability any caller could aim at the primary reporting channel" this table's preamble exists to deny. Contract: the three failure sites (`:797`, `:872`, `:1096`) and the catch-up refusals pass **no** outcome, or explicitly force `job-failure`; **only** `:1060` and `:1061` carry one, and it cannot arrive from caller-supplied `opts`. **The observable is the contract, the mechanism is the implementer's:** an outcome injected into `runJob`'s `opts` leaves every ordinary failure path's reporting — append attempted, subject, reason, return value — unchanged |
 | Unknown value | **refuse the call loudly** — throw rather than fall back to any behavior. Silently degrading an unrecognized discriminant to `job-failure` would append through a refused artifact; silently degrading it to a skip would suppress the record. Neither is acceptable, so an unknown value is a programming error surfaced at once. Note this is the ONE place `failLoud` may throw; it is unreachable from the five sites, which pass literals |
 | Absent input | byte-identical to HEAD on every observable: the append is attempted, the subject is `` `job ${name} failed` ``, and the return is the existing `persisted` boolean |
 | Canonical strings | each of the three values has exactly one reason template and exactly one subject template, asserted verbatim by the tests. "Must say" prose is not a contract; the strings are |
@@ -281,6 +284,12 @@ values, and `failLoud` derives everything from it internally.
 - [ ] **Table E's enum** — its three values appear in the Deliverables row, the
       Exact contracts paragraph, the acceptance criteria and the Security
       checklist; adding or renaming a value moves all four
+- [ ] **The outcome→reason derivation (amended 2026-09-01)** — Table E's three
+      amended rows (one shared template; `failLoud` self-derives for the
+      non-default outcomes; the outcome channel is separate from the forwarded
+      `opts`) are mirrored by Table B `:173`'s throw requirement and by the
+      poison-reason / poison-outcome / one-template acceptance criteria. A change
+      to which side owns the string moves all of them together
 - [ ] **Reachability of every acceptance precondition** — B1 both-unwritable is
       reachable, B2 both-unwritable is not and is explicitly excluded; any new
       criterion must be walked against the call graph before it is added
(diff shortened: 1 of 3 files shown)


thegyula:
## Review gates — round 1 (tip `1273288`)

**wd-reviewer (merge gate): APPROVE conditional** on the PR-title fix (done — title now matches the DoD literal). Independently reproduced the red/green pair (76/16 on reverted run-job.js, the 2 still-green being exactly the HEAD-preservation assertions — non-vacuous). Four C-level findings (GWS-unconfigured B2 untested; pidfile clause covered only by an adjacent test; two tests leak stderr; raw fs messages unredacted on stderr — noted, no spec instruction) + one discovered issue (stale catchUp comment at :1383) + a spec-fault routed to wd-architect.

**Codex plugin: 1 finding [P2]** — failLoud persists the caller-supplied reason on success outcomes where Table E requires internal derivation.

**Herdr shadow (comparison): needs-attention, 2 findings** — the same derivation-locus defect, sharpened: the outcome rides the generic forwarded opts, so an injected valid B2 outcome at an ordinary failure site suppresses healthy reporting; and the four both-writes-fail tests are vacuous about the failLoud attempt (removing the call would still pass them).

**Dispositions:** the spec tension was the architect's to fix — Table E amended on this branch (`9b788ea`: outcome decides reason+subject+append-policy, caller reason honored only for job-failure; the outcome channel separated from forwarded opts; three new acceptance criteria — poison-reason, poison-outcome, one-template-both-paths). Implementer applying the code/test fixes now; fresh gate round on the fixed tip.

thegyula:
## Review gates — round 2 (tip `5a2c35d`; docs touch-up `73ce219`)

**wd-reviewer (merge gate): APPROVE.** All round-1 findings verified genuinely fixed against the amended Table E: failLoud self-derives B1/B2 reasons (poison tests 6/6 RED on the pre-fix tree, independently reproduced); all 9 failLoud call sites enumerated — 7 ordinary sites force `job-failure` via `ordinaryFailLoudOpts` (a poisoned key cannot survive the spread-then-override), only `:1060`/`:1061` carry literal outcomes; countingAppendAlert delegates and counts in all four both-fail tests; GWS-unconfigured B2, pidfile retention via real reap fixtures, and captureStderr wraps all confirmed. Full suite 2380/0 fail; lint clean; boundary exit 0; worktree byte-identical after review. Four non-blocking quality notes recorded in the raw report; three stale spec mirrors routed to wd-architect and fixed in `73ce219` (docs-only, per the reviewer's explicit no-re-review-round disposition); the `opts.appendAlert` seam observation filed as a spec residual for a future architect pass (production-unreachable, test-only by contract).

**Codex plugin: clean** — "guards all five specified state-write sites, preserves original failure reasons, and implements the closed failLoud outcome behavior without altering successful paths. No actionable correctness defects."

**Herdr shadow (comparison): approve** — "Both findings genuinely closed... No material regression found."

**Both gates clean → WP-failloud-survives-state-write-failure review is complete.** Ready for maintainer merge. The mode-pin WP (`WP-private-state-writers-mode-pin`) dispatches once this is merged and its spec flips Done.

Issue #168, opened 2026-08-20
schedule.json is rewritten world-readable on every job run — writeScheduleState drops the 0600 mode


felho:
## Observed (0.13.0)

`wienerdog doctor` warns:

```
[warn] ~/.wienerdog/state/schedule.json has wrong permissions (expected 0700 for folders, 0600 for files) — run 'wienerdog sync' to repair it
```

and the file is indeed `-rw-r--r--` (0644). The same warning also lands in the rendered digest, so it is injected into every session until repaired.

`wienerdog sync` does harden it to 0600 — but the repair does not stick: the next scheduled job run rewrites the file back to 0644. On a machine with a nightly dream this means the file spends essentially all of its life world-readable, and the doctor/digest warning reappears the morning after every sync.

## Root cause

`writeScheduleState` in `src/scheduler/jobs.js` writes the watermark atomically via temp+rename, but the temp file is created with the process default mode and the rename carries that mode onto the final file:

```js
function writeScheduleState(paths, name, patch) {
  const state = readScheduleState(paths);
  state[name] = { ...state[name], ...patch };
  const file = scheduleStatePath(paths);
  const tmp = `${file}.${process.pid}.tmp`;
  fs.mkdirSync(path.dirname(file), { recursive: true });
  fs.writeFileSync(tmp, `${JSON.stringify(state, null, 2)}\n`);   // default mode → 0644
  fs.renameSync(tmp, file);                                       // 0644 survives the rename
}
```

Meanwhile `src/core/private-fs.js` explicitly lists `schedule.json` among the private (0600) files — so the writer and the policy disagree, and the doctor correctly flags the result.

## Suggested fix

Create the temp file private and make it umask-proof:

```js
fs.writeFileSync(tmp, `${JSON.stringify(state, null, 2)}\n`, { mode: 0o600 });
fs.chmodSync(tmp, 0o600); // mode is masked by umask at creation; chmod pins it
fs.renameSync(tmp, file);
```

The same pattern may be worth a quick audit at the other state writers that go through temp+rename (`watermarks.json` etc.) — anything on the private-fs list that is rewritten at job end rather than only at init/sync time.

## Repro

1. Fresh install, any scheduled job configured (the nightly dream is enough).
2. Let the job run once.
3. `ls -l ~/.wienerdog/state/schedule.json` → `-rw-r--r--`; `wienerdog doctor` warns.
4. `wienerdog sync` → 0600; after the next job run it is 0644 again.

[...]

PR #190, opened 2026-09-01
fix(state): write every private-listed state file through the private writer (WP-private-state-writers-mode-pin)

thegyula:
Spec: docs/specs/WP-private-state-writers-mode-pin.md

Closes issue #168's defect: `writeScheduleState`, `writeWatermarks` and `clearAlerts` move onto `writeFilePrivate` (exact 0600 under any umask, both destination states, no loose-mode window, no stale/symlinked-temp write-through); `appendAlert`'s compaction branch migrates inside a non-throwing local try/catch (Table D); `private-fs.js`'s F10 post-rename throw gains `err.code = 'WD_F10_POST_RENAME'` (Table D2) — completing the producer side of the persisted-false wire whose consumer landed with the failloud WP; the two stale waiver-comment mirrors corrected per the lifted 2026-09-01 owner decision.

## Deliverables checklist

- [x] Every changed file listed in the (amended) Deliverables table — incl. `tests/unit/scheduler-runjob.test.js`, added by wd-architect for exactly one fixture: the `:2661` failure-injection used the predictable-temp hazard this WP retires; re-armed against the post-mode-pin refusal shape (F16 destination symlink, mirroring `:2739`)

## Verification output

```
Table C gate GREEN: 10/10 applicable mode cells exactly 0600 (umask 000 + 0777 ×
  absent/present; clearAlerts present-only with its delete branch asserted)
Table C gate RED on unfixed HEAD (exit 1): 0666 under umask 000, 0000 under 0777, all three writers
Table C gate RED on mode-argument-only fix (exit 1): fails exactly at umask 0777 — the cell a
  single-umask gate cannot catch
M2 red side: writeScheduleState temporarily re-pointed at a hand-rolled temp+rename → its
  delegation spy fails (0 !== 1, "writeFilePrivate called exactly once"); restored, spy green again
New unit tests: 16/16 (mode cells, delegation spies × 3, stale/symlinked temp, F16 refusal + no
  side effects, idempotence, D2 tag, Table D cases + both discrimination directions)
npm test → 2408 tests, 2396 pass, 0 fail, 12 skipped
npm run lint → clean; boundary-check → exit 0
```

## Decisions made

1. Table D Case 1/2 failure injection via the codebase's `stubCollaborators` require-cache-swap idiom — no test-only parameter added to `appendAlert`'s production signature. Round 1 generalized this into `stubWriteFilePrivateFor(moduleId, stubFn)` and reused it for the three new delegation-spy tests (M2).
2. Compaction-refusal diagnostic on `process.stderr` per the file-wide convention.
3. The `:2739` sibling test's vestigial `mkdirSync` left untouched per the amended Deliverables scope.
4. Round 1, M1(b): checked how `:2739` (the reference sibling for the re-armed `:2661` fixture) handles win32 — it carries **no** explicit platform skip guard, and neither does the rest of its Table A/B/C section; those tests are implicitly POSIX-only because `writeScript` writes a `#!/bin/sh`-shebang fake job executable, unusable on win32 regardless of the symlink. Matching that existing, unguarded convention was the smaller change, so `:2661`'s re-armed fixture also carries no new explicit guard — see Discovered issues.

- Table A rows 13–14 (`config.yaml`, `install-manifest.json`) creation-path residual — named in the spec.
- `appendAlert`'s own append at `:89-90` (create window, destination-symlink write-through, growth) — named follow-up owner in the spec.
- `src/cli/run-job.js`'s `failLoud` JSDoc (`:655-658`, the G2 paragraph, and `:681-683`) is now stale: it still describes "a throw from the post-append compaction rewrite also yields `false`" and "`appendAlert`... returns `undefined` on every path today" — both were true before this WP and are superseded by Table D, which wraps compaction in a non-throwing local try/catch and adds the explicit `false` return on Case 2. `run-job.js` is outside this WP's Deliverables boundary; flagging for whoever next touches that file's JSDoc.
- Table A/B/C section of `tests/unit/scheduler-runjob.test.js` (including the re-armed `:2661` and its `:2739` sibling) has no explicit win32 skip guard; it is implicitly POSIX-only via `writeScript`'s shebang fake-job mechanism. Not introduced by this WP, not fixed here (outside Deliverables boundary beyond the one permitted `:2661` fixture edit).

## Review gates

Runs recorded in comments below (wd-reviewer + Codex plugin on the same tip). Round 1 findings applied in this round: M1 (plugin P2 + shadow #2, Windows guards on the repeat-run mode assertions and a check of the re-armed `:2661` fixture's platform handling), M2 (shadow #1, delegation spies for all three writers), M3 (wd-reviewer finding 3, stderr-leak guard on the EXDEV discrimination test).

## Lessons (for memory/lessons/inbox.md)

- WP-private-state-writers-mode-pin: a WP that closes a "predictable temp path" hazard breaks any earlier test that used that predictability as its failure-injection fixture — check sibling WPs' suites before assuming a red test is your own defect.
- WP-private-state-writers-mode-pin: the `stubCollaborators` require-cache-swap idiom is the established way to unit-test an unexported failure branch without widening a production signature; it generalizes cleanly across multiple modules sharing one collaborator (`stubWriteFilePrivateFor`).
- WP-private-state-writers-mode-pin: pre-writing an oversized malformed line directly to `alerts.jsonl` is the cheap established trigger for the compaction branch — no 200-call seeding.
- WP-private-state-writers-mode-pin: a test asserting only a writer's *final on-disk mode* (Table C) cannot distinguish a correct fix from a regression that reintroduces a hand-rolled writer which happens to land the same mode by luck — pair mode assertions with a delegation spy on the shared primitive.
- WP-private-state-writers-mode-pin: before adding a platform skip guard to a new/modified test, check whether the surrounding test family already relies on an implicit platform restriction (here: shebang fake-job scripts) — matching the existing convention is smaller and more honest than a locally-added guard that implies the rest of the family is portable when it isn't.

---
Generated-by: Claude Sonnet 5 (implementer) / Claude Fable 5 (orchestrator)

https://claude.ai/code/session_01LCXat1SiCGPBaijqCCkAjG

### DIFF
--- a/docs/specs/WP-private-state-writers-mode-pin.md
+++ b/docs/specs/WP-private-state-writers-mode-pin.md
@@ -186,6 +186,7 @@ repair cannot hold a file that a scheduled job rewrites nightly.
 | modify | src/core/alerts.js | two functions: (a) `clearAlerts` — replace lines 206-208 per Table B, thread `{ core: paths.core }`; (b) `appendAlert`'s **compaction branch** — replace lines 117-120 per Table D (migrate + local try/catch, must not throw). Do NOT touch the atomic append at `:89-90`, the empty-read guard at `:105`, the compaction trigger at `:106`, or `chmodAlerts` itself |
 | modify | src/core/private-fs.js | **comment text PLUS one tagged throw** (this row was "comment-only" in earlier rounds; it is not any more). (a) Correct the two now-false mirrors of the 2026-07-19 decision at lines 20-22 and 129-134 (the `A9_PRIVATE_STATE_FILES` JSDoc) so they no longer claim these writers are unchanged. The array literal it documents, lines 135-141, is CODE and is NOT edited — its membership is correct and 3 of its 5 entries were never part of that decision. The `A9_PRIVATE_CORE_FILES` comment at lines 143-148 also stays TRUE and is NOT edited. (b) Add `code = 'WD_F10_POST_RENAME'` to the single throw at lines 361-365, per Table D2 — the ONLY code change permitted in this file |
 | create | tests/unit/private-writer-modes.test.js | cover the acceptance criteria below; the implementer designs the cases |
+| modify | tests/unit/scheduler-runjob.test.js | **ONLY the `:2661` fixture**, whose failure-injection technique — a directory pre-created at the predictable temp path `${alertsFile}.${process.pid}.tmp` (`:2674`) — this WP's own fix retires, so it no longer intercepts anything and `clearAlerts` succeeds. Replace it with a post-mode-pin refusal trigger (e.g. symlink `alertsFile` itself, per the `:2739` sibling). **No other change to the file** — see "A test that the fix retired" below |
 
 ### Exact contracts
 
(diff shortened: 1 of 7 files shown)


thegyula:
## Review gates — final record (tip `2938fd1`)

- **R1** (tip 33d20c1): wd-reviewer **APPROVE, merge-as-is** (Table C gate green + both red sides: unfixed tree 0666/0000; mode-argument-only fails exactly at umask 0777; live schedule.json still 0644 on the installed 0.13.0 — the field defect confirmed untouched by review probes). Plugin P2 (POSIX assertions unguarded on win32) + shadow (delegation spies; :2661 fixture platform posture) → fixed in `c5f73af` (guards; 3 delegation-spy tests, red-proven; EXDEV stderr count). wd-architect reconciled the spec in `2938fd1` (Table D cell ↔ Table B core-threading; residual count 5 with the 3-subset guarded; run-job.js JSDoc registered KNOWN-STALE).
- **R2**: wd-reviewer **APPROVE** (all dispositions verified; the :2661 no-guard disposition upheld on sibling-parity grounds with the shebang rationale corrected for the record); plugin **clean**.

**Both gates clean.** Final: 2408 tests / 0 fail, lint clean, boundary green. Follow-ups: comment-only run-job.js JSDoc refresh; the two unguarded symlink siblings (:2679/:2754) guarded together.

PR #194, opened 2026-09-01
test(dream): close the show option-position slot and re-sync the guard's mirrors (WP-show-slot-own-value-kind)

thegyula:
Spec: docs/specs/WP-show-slot-own-value-kind.md

Shape (4)'s argument stops being a FREE slot and becomes the literal
`HEAD:reports/warnings.md`, **retyped rather than imported** from `WARNINGS_REL`.
`git show --output=<user index>` is a two-token call that overwrites the user's
index and carries no `GIT_INDEX_FILE`, so it matched `['show', ANY]` on arity
*and* satisfied the `unset` disposition. Because the literal and its interpolated
form are identical at run time, the pinned set moves into
`tests/unit/dream-pipeline.known-calls.js` and the guard file hashes that WHOLE
file against `KNOWN_CALLS_SOURCE_DIGEST` — the only thing that can hold the
choice. C2 restates the own-value membership invariant so every member satisfies
it (MINTED is retired); C3 records the residual's closure at `b19121bb` and drops
the bare source count from both prose surfaces.

Full verification transcripts, including every RED, are in a PR comment.

## Deliverables checklist

Changed files (`git diff --name-only main...HEAD`):

```
docs/specs/WP-show-slot-own-value-kind.md
docs/specs/done/WP-dream-promote-in-workspace.md
docs/specs/logbook/2026-08-31-index-refresh-dropped-with-its-cause.md
tests/unit/dream-pipeline.known-calls.js
tests/unit/dream-pipeline.test.js
```

## Verification output

```
$ npm test -- tests/unit/dream-pipeline.test.js
✔ dream-pipeline: the pinned call set module is its canonical SOURCE FORM (Table W row W1(c)) (1.70ms)
… 44 tests
ℹ tests 44 / pass 44 / fail 0
$ echo "guard-file suite exit=$?"
guard-file suite exit=0

$ npm test
ℹ tests 2444 / pass 2432 / fail 0 / skipped 12

$ npm run lint
markdownlint: 621 file(s), 0 error(s)
shellcheck / PSScriptAnalyzer: clean
frontmatter check passed: 264 spec(s), 4 agent(s)
lint passed

$ git cat-file -e c853245b^{commit} && echo "c853245b resolves"
c853245b resolves
$ git cat-file -e b19121bb^{commit} && echo "b19121bb resolves"
b19121bb resolves
$ git merge-base --is-ancestor b19121bb 53b1519b && echo ok
b19121bb is an ancestor of 53b1519b

$ node scripts/boundary-check.js docs/specs/WP-show-slot-own-value-kind.md $(git diff --name-only main...HEAD)
boundary-check exit=0
```

### The module IS the spec's extraction, byte for byte (Deliverables, criterion 1)

```
$ git show HEAD:docs/specs/WP-show-slot-own-value-kind.md | awk '…BEGIN/END CANONICAL MODULE…' > /tmp/wd-module.js
$ node -e "…normalize, sha256…"
5887 b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f5c
# expected: 5887 b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f5c
$ diff /tmp/wd-module.js tests/unit/dream-pipeline.known-calls.js && echo same
committed module == spec extraction, byte for byte
```

The file was **created by running the spec's own extraction command**, not retyped.

### Criteria 4, 5 and 12 — four APPLIED mutations, each diff shown before its red

Diffs and full output are in the PR comment. Summary — every mutation was applied,
observed, and then reverted (a clean 44/44 is shown after each revert):

| mutation (applied where) | runtime arm — the three `row G8` index tests | source arm — the committed source-form test | suite |
|---|---|---|---|
| **(a)** slot reverted to `ANY` (module) | **RED** — all three FAIL: `classify(['show','--output=<index>'],{})` returned `null` where `UNKNOWN SHAPE …` is required | RED | 40 pass / **4 fail** |
| **(b)** slot interpolated `` `HEAD:${WARNINGS_REL}` `` (module) | **GREEN** — all three PASS, `violations` empty, both canaries classify as criterion 2 requires | **RED** with the re-pin message | 43 pass / **1 fail** |
| **(c)** slot concatenated `'HEAD:' + WARNINGS_REL`, split across two lines (module) | **GREEN** — all three PASS, same as (b) | **RED** with the re-pin message | 43 pass / **1 fail** |
| **(d)** digest constant tampered, module canonical (guard file) | GREEN | **RED**, naming both hexes | 43 pass / **1 fail** |

**(b) and (c) are the point, and the full `npm test` REDS on both — that is the
tripwire working, not a contradiction.** A green suite on either would mean the
relocation tripwire is dead: the guard's own decision is unchanged (the runtime
arm is green by construction, because the two forms produce an identical token),
so the source-form digest is the only surface that can tell them apart.
**(a) and (d) are the two directions of the canary discipline** — (a) proves the
runtime pin is alive against a two-token vector that matches the shape's arity,
(d) proves the durable arm itself can fail.

### Criterion 5 — the state table RE-RUN, not cited

All eleven states applied to the real files and run against the **committed**
source-form test (`node --test tests/unit/dream-pipeline.test.js`):

```
PASS       | canonical module
PASS       | pure re-indentation of the module
FAIL       | show slot interpolated (WARNINGS_REL)
FAIL       | show slot concatenated ('HEAD:' + WARNINGS_REL)
FAIL       | a tenth shape added
FAIL       | the nine shapes reordered
FAIL       | a canonical decoy in a comment inside the module
FAIL       | a block-scoped declaration added to the module
FAIL       | one word of a comment changed (a typo fix)
FAIL       | KNOWN_CALLS_SOURCE_DIGEST tampered, module canonical
FAIL (suite could not run: Cannot find module) | the module file absent
```

Identical to the table the spec records at `5d31a7dc`.

### Criterion 8 — the claim sweep, one pattern per pass, every hit adjudicated

```
docs/specs/WP-index-guard-residuals.md :: read back out of :: 1
docs/specs/WP-index-guard-residuals.md :: values THIS RUN PRODUCED :: 1
docs/specs/done/WP-dream-promote-in-workspace.md :: MINTED :: 2
docs/specs/done/WP-dream-promote-in-workspace.md :: read back out of :: 1
docs/specs/done/WP-dream-promote-in-workspace.md :: values THIS RUN PRODUCED :: 1
docs/specs/done/WP-dream-promote-in-workspace.md :: 5c5d082 :: 3
docs/specs/done/WP-dream-promote-in-workspace.md :: KNOWN_CALLS_SOURCE_DIGEST :: 1
docs/specs/done/WP-dream-promote-in-workspace.md :: known-calls.js :: 1
docs/specs/logbook/2026-08-31-index-refresh-dropped-with-its-cause.md :: values THIS RUN PRODUCED :: 1
memory/lessons/inbox.md :: MINTED :: 1
tests/unit/dream-pipeline.known-calls.js :: read back out of :: 1
tests/unit/dream-pipeline.test.js :: read back out of :: 1
tests/unit/dream-pipeline.test.js :: KNOWN_CALLS_SOURCE_DIGEST :: 3
tests/unit/dream-pipeline.test.js :: b770de2e1030 :: 1
tests/unit/dream-pipeline.test.js :: known-calls.js :: 1
```

**Eight of the fifteen patterns now report ZERO hits tree-wide** — `four sources`,
`admits six`, `single-line output of`, `stays free`, `merely carries`,
`` stay `ANY` ``, `'show', ANY`, `show «HEAD` — which is the drift closing.

Adjudication, hit by hit:

- `WP-index-guard-residuals.md` (both patterns) — **registered NON-move**: its item 2
  quotes C2's sharpened invariant in the hand-off this WP gives it. Draft stub, correct as written.
- Done spec `MINTED :: 2` — **corrected text**. Both occurrences are inside the new
  closure record: the quoted subject of `b19121bb` (provenance) and the sentence
  immediately after it stating that the word in that subject is itself retired,
  because `rev-parse HEAD` reads the head back from the user's ref.
- Done spec `read back out of :: 1` — **corrected text**, inside C1's rewritten
  clause (the surviving, still-true half of the retired "stays free" reasoning).
- Done spec `values THIS RUN PRODUCED :: 1` — **registered NON-move**: W1(c)'s
  *"THE REPAIR DOES NOT INSPECT THE TOKEN"* clause, which the C2 checklist decides
  is `WP-index-guard-residuals` item 2's to sharpen.
- Done spec `5c5d082 :: 3` — **all three on `:545`, row W5**, exactly as the spec
  predicts. Registered Discovered issue below; out of this WP's boundary.
  `grep -n -o '5c5d082'` → `545:`, `545:`, `545:` and nothing else.
- Done spec `KNOWN_CALLS_SOURCE_DIGEST :: 1` / `known-calls.js :: 1` — **expected
  hits**: W1(c) names the constant and names where the executable copy lives. The
  digest **prefix** `b770de2e1030` reports the **guard file only**, as required —
  W1(c) registers the constant by name and does not quote its value, so the bytes
  live in exactly one place.
- Logbook `values THIS RUN PRODUCED :: 1` — **registered NON-move** (`:356-366`);
  only its dead SHA moved, `5c5d082` → `c853245b`.
- `memory/lessons/inbox.md :: MINTED :: 1` — **registered NON-move, deliberately not
  edited** (also: not editable on a WP branch).
- `known-calls.js` / `test.js :: read back out of` — **corrected text**, C2's
  restated invariant in the module JSDoc and in the `produces` recorder comment.
- `test.js :: KNOWN_CALLS_SOURCE_DIGEST :: 3` / `b770de2e1030 :: 1` /
  `known-calls.js :: 1` — **expected**: the constant's declaration, its assertion
  and its failure message; the pinned value once; the require's single path constant.

### Criterion 6 — MINTED is gone from the guard file

```
$ grep -c MINTED tests/unit/dream-pipeline.test.js
0
```

### The W1(c) whole-cell re-read — done, and confirmed

W1(c) is one markdown table cell of ~46 KB on line `541`. After the five in-place
rewrites the cell was extracted whole (`awk 'NR==541'`), folded to 431 readable
lines, and **read end to end**. Confirmed:

- the two-way partition sentence (*"A slot holding a value the run COMPUTED is `own`;
  … merely carries … is FREE"*) is **gone**, replaced by the three clauses — no
  surviving copy anywhere in the cell (`merely carries` sweeps to zero);
- the *"`show «HEAD:path»` stays free"* ruling is **gone** (`stays free` and
  `show «HEAD` both sweep to zero) and its still-true half (no observation pins the
  token) survives only inside the new clause;
- the *"five shapes the rule reaches"* sentence is **untouched** and still true —
  the amendment leaves the own-value rule alone, and a sentence saying so now
  precedes it so "the rule" reads unambiguously;
- enumeration item **(4)** is the literal, and no second spelling of shape (4)
  survives in the cell;
- the residual clause is **replaced** by the closure record — no stale
  four-versus-six sentence survives beside it;
- `5c5d082` is gone from row W1(c) (it survives only in row W5, out of boundary);
- the three new sentences (module location, digest constant by name, same-commit
  re-pin) are present exactly once each.

## Decisions made

1. **Where the three new W1(c) sentences land.** The spec fixes their content, not
   their position. Placed immediately after the *"In the executable copy the two
   kinds are the symbols `RUN_VALUE` and `ANY`"* sentence — the one place the row
   already talks about the executable copy and registers the pair — rather than
   opening a new clause. Simpler option: extend the existing registration.
2. **The module path is named once in W1(c), not twice.** A first draft also named
   `tests/unit/dream-pipeline.known-calls.js` in the closure record when attributing
   the source count to the `produces` markers; changed to "the `produces: true`
   markers beside the set" so the sweep's `known-calls.js` pattern stays
   single-homed inside the row.
3. **`b19121bb`'s subject is quoted, and the MINTED inside it is dispositioned in
   the same breath.** Naming the closing commit is criterion 7; quoting its subject
   is the provenance discipline `c853245b` was verified by. Since C2 retires the
   word, the quotation carries an explicit note that the word in that subject is the
   retired one and the membership predicate stands in its place — rather than
   silently dropping the subject.
4. **The logbook's retired slot-kind sentence is recorded as retired, not merely
   deleted.** The *"Slots holding data the run did not compute — modes, paths,
   messages — stay `ANY`"* sentence is gone (it violated that entry's own deference
   sentence and C1 makes it false); one sentence in its place says it stood until
   2026-09-01 and why, per Table W's retire-with-cause discipline.
5. **The show canary sits beside the two-token `read-tree` canary**, inside the
   existing non-vacuity block, and carries both directions (the `--output=` vector
   rejected, the run's own `show HEAD:reports/warnings.md` still admitted) — the
   canary discipline the spec makes mandatory.

Nothing else was ambiguous. No scope was expanded; `src/` is untouched.

- **Row W5 of `docs/specs/done/WP-dream-promote-in-workspace.md` (`:545`) carries
  THREE dead `5c5d082` references.** Measured contexts: *"Both forms are measured
  RED at"*, *"ACCEPTED by the set of that date — see the … cells below"*, *"ONE CELL
  WAS ADDED AT"*. **The correct SHA is `c853245b`** — *"test(dream): pin the
  object-name slots to values the run itself computed"*, verified by CONTENT (its
  diff introduces `RUN_VALUE` and the five own-value slots), not by name alone.
  `git cat-file -t 5c5d082` → `fatal: Not a valid object name`, almost certainly a
  fork-port history rewrite. **Not fixed here: this WP's Deliverables scope it to row
  W1(c), and W5 is a different row** — fixing them would breach the permission
  boundary this spec exists to define. The sweep pattern `5c5d082` will keep
  reporting `:545` until a package takes them.
- **`src/core/dream/warnings.js:63-66`'s JSDoc is known-falsified by design.** It
  says `WARNINGS_REL` is *"the ONE place it is decided"* and that anything naming the
  path imports it rather than retyping it — and C1 deliberately retypes it in the
  guard, because an importing pin would follow a relocation silently. The claim
  narrows to PRODUCTION consumers; the guard is not one. `src/` is frozen in this WP,
  so nothing was edited; the one-line JSDoc narrowing rides the queued architect
  follow-up batch (owner ruling 2026-09-01, per the spec's registered non-move).
- **This spec's canonical module block (`:506-613`) is an unregistered THIRD copy of
  the pinned set — and the sweep is structurally blind to it.** *(Raised by the
  wd-reviewer gate; routed to wd-architect, not fixed here.)* W1(c)'s Mirrored
  Surface Checklist states the set has exactly two copies and that no third may be
  written. The sweep excludes this spec file on the rationale that it *"quotes the
  pre-change text on purpose"* — which is true of its other quotations but **not** of
  the canonical block, which quotes the **post**-change set. The block has to be
  verbatim for the digest derivation to be reproducible, so this is a real tension
  rather than an oversight; the fix is to mark it a frozen historical snapshot on
  merge. Left alone here: editing it would change the extraction source the pin is
  derived from.
- **Clause (2) landed in W1(c), but the nine-slot adjudication that makes it
  CHECKABLE did not.** *(Raised by the wd-reviewer gate; routed to
  `WP-index-guard-residuals`.)* The per-slot walk lives only in this spec's C1 cell,
  which becomes history on merge. The specific thing lost from the canonical surface
  is the measurement that shape (3)'s `«path»` is consumed **positionally** by
  `--cacheinfo` — so it is clause (3), not clause (2). Without it, a future author
  applying clause (2) to `update-index --add --cacheinfo «mode» «own sha» «path»`
  could reasonably read that slot as an option position and conclude the shipped FREE
  marker is a defect. Not fixed here: the Mirrored Surface Checklist required the
  three clauses and nothing more, so adding the adjudication would be scope
  expansion into a cell `WP-index-guard-residuals` item 2 re-edits anyway.

## Lessons (for `memory/lessons/inbox.md`, not edited on this branch)

- `WP-show-slot-own-value-kind:` A pin that is spelled and a pin that is interpolated
  are **identical at run time**, so no behavioural test can tell them apart — the
  whole RED/GREEN matrix, the index test and every presence-grep go green under the
  interpolation. Only a check over the SOURCE FORM can hold that choice, and it has
  to be committed: a proof living in the spec's verification steps cannot bind a
  future WP that reads only its own spec.
- `WP-show-slot-own-value-kind:` **When a checker must FIND its subject inside a
  large file, it is enumerating the ways the host language can hide it, and that
  never closes** — six rounds broke the locating six different ways. The fixed point
  is to delete the locating step: give the artifact a file of its own and hash the
  whole file. The evasion class then stops existing instead of being enumerated away,
  and the machinery shrinks to read, collapse, hash, compare.
- `WP-show-slot-own-value-kind:` **A canary that differs from its exploit in ARITY
  proves nothing about the exploit** — it dies on length equality before reaching the
  slot under test. The show vector is two tokens *because* the pinned shape is two
  tokens; a three-token probe would have certified a rejection the set never made.
  Same lesson the `read-tree` gap cost, one shape over.
- `WP-show-slot-own-value-kind:` **A stated invariant that one of its own members
  falsifies cannot decide the next case.** The own-value set's "MINTED, never READ
  BACK" was false for `rev-parse HEAD` on the day it was written; the fix is to
  restate the membership rule so every member satisfies it, not to evict the member
  (evicting `head` would redden every legitimate run — it feeds three own-value slots).
- `WP-show-slot-own-value-kind:` **A docs surface can be stale on the day it lands.**
  Both prose mirrors of the own-value count were written by a commit that is a
  DESCENDANT of the commit that already fixed the code — the prose recorded the
  pre-fix state. Checking a mirror's ancestry against the fix it describes is cheap
  and catches this class.
- `WP-show-slot-own-value-kind:` **A count beside a list is a second copy that goes
  stale on its own.** Both prose surfaces stated "four sources" while the list of
  `produces` markers was the real surface; the remedy is to delete the number and
  point at the list, which is now written into W1(c) as a standing rule.
- `WP-show-slot-own-value-kind:` **Re-read a rewritten canonical cell WHOLE.** W1(c)
  is ~46 KB on one line; no mirror checklist can see inside it and the failure mode is
  a new sentence landing while the superseded one survives three clauses away. Extract
  the line, fold it, read it end to end — and sweep the retired sentences' own unique
  phrases to zero as the mechanical half of that check.
- `WP-show-slot-own-value-kind:` **A digest derived from a RECIPE cannot survive two
  readings of the recipe.** The spec records that two independent reconstructions of
  "move these lines, then apply these edits" landed 125 characters apart. Shipping the
  artifact as one canonical block, extracted mechanically by the spec's own command,
  is what made the pin reproducible — the implementer must run the extraction, never
  retype the block.

---
Generated-by: claude-opus-5 (implementer) / orchestrated by claude-fable-5


thegyula:
## Full mutation transcripts — each diff shown BEFORE its outcome is believed, and each shown reverted

### Mutation (a) — the slot reverted to `ANY` (criteria 4, 5a). RUNTIME RED.

Applied, in `tests/unit/dream-pipeline.known-calls.js`:

```diff
@@ -90,7 +90,7 @@ const KNOWN_CALLS = [
   // interpolate it. The run's own argument is a constant built at
   // `cli/dream.js:1004` from `core/dream/warnings.js:72`; retyping it here is
   // the tripwire that makes a relocation LOUD instead of silent.
-  { env: 'unset',   args: ['show', 'HEAD:reports/warnings.md'] },
+  { env: 'unset',   args: ['show', ANY] },
   { env: 'unset',   args: ['rev-parse', 'HEAD'], produces: true },
```

Observed:

```
✖ dream-pipeline: the pinned call set module is its canonical SOURCE FORM (Table W row W1(c)) (1.90ms)
✖ dream-pipeline: the run does not touch the user's git index — at all, plain vault (row G8) (545.83ms)
✖ dream-pipeline: the run does not touch the user's git index — at all, separate-git-dir vault (row G8) (515.27ms)
✖ dream-pipeline: the run does not touch the user's git index — at all, linked-worktree vault (row G8) (620.19ms)
ℹ tests 44
ℹ pass 40
ℹ fail 4

✖ dream-pipeline: the run does not touch the user's git index — at all, plain vault (row G8)
  AssertionError [ERR_ASSERTION]: the two-token show redirect must be rejected: it
  satisfies arity and the unset disposition, so only the literal slot stands between
  it and the user's index
  + actual - expected

  + null
  - "UNKNOWN SHAPE — not one of the run's pinned calls"
```

**The vector is TWO tokens**, the same arity as the pinned shape, so it reaches the
slot under test rather than dying on length equality. Reverted:

```
$ git checkout -- tests/unit/dream-pipeline.known-calls.js
$ git diff --stat -- tests/unit/dream-pipeline.known-calls.js
(no diff)
$ npm test -- tests/unit/dream-pipeline.test.js
ℹ tests 44 / pass 44 / fail 0
```

### Mutation (b) — the interpolated form of `WARNINGS_REL` (criteria 5b, 12a). RUNTIME GREEN, SOURCE RED.

This is **the future-editor scenario the whole design exists for**: the digest
constant is left UNCHANGED, and a DRY-minded rewrite is attempted. Applied:

```diff
@@ -1,4 +1,5 @@
 'use strict';
+const { WARNINGS_REL } = require('../../src/core/dream/warnings');
 /**
  * THE PINNED CALL SET, IN A FILE OF ITS OWN — and that is the whole design.
@@ -90,7 +91,7 @@ const KNOWN_CALLS = [
-  { env: 'unset',   args: ['show', 'HEAD:reports/warnings.md'] },
+  { env: 'unset',   args: ['show', `HEAD:${WARNINGS_REL}`] },
```

Observed — **the two arms captured separately**:

```
✖ dream-pipeline: the pinned call set module is its canonical SOURCE FORM (Table W row W1(c)) (2.12ms)
✔ dream-pipeline: the run does not touch the user's git index — at all, plain vault (row G8) (521.80ms)
✔ dream-pipeline: the run does not touch the user's git index — at all, separate-git-dir vault (row G8) (524.05ms)
✔ dream-pipeline: the run does not touch the user's git index — at all, linked-worktree vault (row G8) (560.20ms)
ℹ tests 44
ℹ pass 43
ℹ fail 1

✖ dream-pipeline: the pinned call set module is its canonical SOURCE FORM (Table W row W1(c))
  AssertionError [ERR_ASSERTION]: the pinned call set module is not its canonical
  source form. Any edit to it re-pins KNOWN_CALLS_SOURCE_DIGEST in the SAME commit
  as the Table W row W1(c) change.
```

**RUNTIME arm, named test by test:** all three `row G8` layouts PASS — `violations`
is empty and both canaries classify as criterion 2 requires, because the
interpolation produces a byte-identical token at run time. **SOURCE arm:** the
committed source-form test FAILS with the re-pin message.

**So the full `npm test` REDS on (b) — because of the source-form test, and that is
the tripwire working, not a contradiction. A green suite here would mean the
tripwire is dead.**

### Mutation (c) — the concatenation `'HEAD:' + WARNINGS_REL`, split across two lines (criteria 5c, 12a). RUNTIME GREEN, SOURCE RED.

```diff
@@ -1,4 +1,5 @@
 'use strict';
+const { WARNINGS_REL } = require('../../src/core/dream/warnings');
@@ -90,7 +91,8 @@ const KNOWN_CALLS = [
-  { env: 'unset',   args: ['show', 'HEAD:reports/warnings.md'] },
+  { env: 'unset',   args: ['show', 'HEAD:'
+    + WARNINGS_REL] },
```

```
✖ dream-pipeline: the pinned call set module is its canonical SOURCE FORM (Table W row W1(c)) (1.96ms)
✔ dream-pipeline: the run does not touch the user's git index — at all, plain vault (row G8) (531.39ms)
✔ dream-pipeline: the run does not touch the user's git index — at all, separate-git-dir vault (row G8) (550.79ms)
✔ dream-pipeline: the run does not touch the user's git index — at all, linked-worktree vault (row G8) (561.18ms)
ℹ tests 44
ℹ pass 43
ℹ fail 1

  AssertionError [ERR_ASSERTION]: the pinned call set module is not its canonical
  source form. Any edit to it re-pins KNOWN_CALLS_SOURCE_DIGEST in the SAME commit
  as the Table W row W1(c) change.
```

Same asymmetry as (b), and the line split makes no difference — the whole file is
the span, so there is nothing for a line-oriented evasion to hide behind.

### Mutation (d) — the digest constant tampered while the module is canonical (criterion 12b). SOURCE RED.

```diff
@@ -134,7 +134,7 @@ function realish(p) {
  *  This digest covers every byte of it; re-pin in the SAME commit as the
  *  Table W row W1(c) change. */
 const KNOWN_CALLS_SOURCE_DIGEST =
-  'b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f5c';
+  'b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f00';
```

```
✖ dream-pipeline: the pinned call set module is its canonical SOURCE FORM (Table W row W1(c)) (3.07ms)
ℹ tests 44
ℹ pass 43
ℹ fail 1

  AssertionError [ERR_ASSERTION]: the pinned call set module is not its canonical
  source form. …
  + 'b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f5c'
  - 'b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f00'
    actual: 'b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f5c',
    expected: 'b770de2e1030590e14ba92da63790d71824445cb45d2fc3e64e14a96c5157f00',
```

**A durable check nobody has watched fail is the 3-pass/0-fail guard again, one
layer out.** Both directions of the durable arm are now observed: (b)/(c) show it
reddening on a non-canonical module, (d) shows it reddening on a tampered pin.

Reverted; tree clean:

```
$ git checkout -- tests/unit/dream-pipeline.test.js
$ git status --short
(clean)
$ npm test -- tests/unit/dream-pipeline.test.js
ℹ tests 44 / pass 44 / fail 0
$ npm test -- tests/unit/dream-pipeline.test.js > /dev/null 2>&1; echo "guard-file suite exit=$?"
guard-file suite exit=0
```

---

## Notes on method

- **`--test-name-pattern` was NOT used anywhere**, per the spec. Measured on this
  tree, it filters nothing in this suite: a pattern matching no test still reports
  the full count, so a green from it attributes nothing.
- **No `wienerdog` CLI command, installer, uninstaller or scheduler operation was
  run.** This WP is tests and docs only; `src/` is untouched.
- **The module was created by running the spec's own `awk` extraction over
  `git show HEAD:docs/specs/WP-show-slot-own-value-kind.md`** and committing what it
  emitted — not retyped. Verified byte-for-byte with `diff` against a fresh
  extraction after committing, and the normalized length (5887) and SHA-256 both
  reproduce the pin.
- **The state table was re-run against the committed test**, not cited: each of the
  eleven states was written to the real files, `node --test` was run on the guard
  file, and the source-form test's own verdict was read out of the output. Every
  state was reverted before the next.

thegyula:
## Review gate: Codex PR review — CLEAN, no findings

Per `docs/runbooks/codex-review.md` "When it runs" item 2 (PR review, the
additional gate alongside wd-reviewer). Backend: Codex CLI `codex-cli 0.151.0`,
native PR review (`codex exec review --base main`, no focus text — the contract's
PR-review input). The `gptsol` agent backend was not available in this session
(`ANTHROPIC_BASE_URL` unset, so the llmp-proxied reviewer model is unreachable);
the Codex backend is the runbook's "works anywhere" path.

**Read-only check (mechanical, required by the contract):** `git status --porcelain`
in the reviewed checkout was **0 bytes before and 0 bytes after** the run —
byte-identical, so the run is valid. Process exit 0.

**What the reviewer EXECUTED** (the contract requires this be stated, not just what
it read): it ran the targeted test suite, the lint pipeline, the boundary check and
the digest verification, and it independently resolved `c853245b` and `b19121bb` by
content (`git show --stat --oneline`) and checked their ancestry.

### Output, relayed verbatim

> The pinned call set is extracted without changing its runtime semantics, the
> vulnerable `show` slot is correctly narrowed to the production literal, and the
> new digest and classifier canaries enforce the intended contract. The targeted
> test suite, lint, boundary check, and digest verification pass.

No findings were returned, so there is nothing to disposition from this gate.

### One incidental corroboration worth recording

The reviewer's own ancestry probe returned `ancestor_exit=0` for
`git merge-base --is-ancestor c853245b b19121bb` — i.e. `c853245b` (which introduces
`RUN_VALUE` and the five own-value slots) is an ancestor of `b19121bb` (which
narrows the own-value set to the four producing shapes). That is consistent with,
and independent of, this PR's own measured claim that `b19121bb` is an ancestor of
`53b1519b` — the ordering that makes the two prose mirrors provably post-dated by
the fix they described. The full chain is
`c853245b → b19121bb → 53b1519b`: ruling, then fix, then the docs that recorded the
pre-fix state.

---

`wd-reviewer` (the spec-fidelity merge gate) is running on the same diff; its
result will be posted here when it lands. Per CLAUDE.md both gates must be clean or
fully dispositioned before merge, and merging is the maintainer's act.

thegyula:
## Review gate: wd-reviewer (spec-fidelity merge gate) — APPROVE

Both gates required by `docs/runbooks/codex-review.md` "When it runs" item 2 have
now run on this diff. **Codex: clean, no findings** (comment above).
**wd-reviewer: APPROVE, no blocking findings**, five observations — all residual or
drop, none in this diff, none the implementer's to fix.

The reviewer re-ran everything independently rather than reading the PR body: the
`awk`-over-`git show` extraction (`5887 b770de2e1030…7f5c`), `diff` of that
extraction against the committed module (**byte-identical**), the check block
against the spec's Exact-contracts block (byte-identical, digest value included),
`npm test` (2432 pass / 0 fail), `npm run lint`, `boundary-check` (exit 0), the SHA
resolutions and ancestry, and the claim sweep — **whose output came back
character-for-character identical to the PR body's**, with the same eight patterns
at zero hits tree-wide.

It also did its own non-vacuity proof of the source-form check, read-only: it hashed
a scratch copy with the slot swapped to `` `HEAD:${WARNINGS_REL}` `` and got
`19cce323faa6…` ≠ the pinned `b770de2e1030…`. That is an independent confirmation
of mutation (b) obtained without touching the tree.

**On the flagged failure mode**, it extracted line 541 (46,325 bytes), folded it and
read it end to end, confirming no superseded sentence survives beside its
replacement — and verified every source citation the new prose makes
(`cli/dream.js:1004`, `:1041`, `:238-239`, `core/dream/warnings.js:72`).

### Findings and dispositions

| # | Kind | Finding | Disposition |
|---|---|---|---|
| 1 | contract | This spec's canonical module block (`:506-613`) is a **third textual copy** of the *post*-change set, against W1(c)'s "exactly two copies" rule. The sweep excludes this file on the rationale that it quotes *pre*-change text — which does not cover this block. On the next re-pin it silently becomes a stale third copy the sweep will never report. | **residual → wd-architect.** Not in this diff; the successor should mark the block a frozen historical snapshot on merge. Added to Discovered issues in the PR body. |
| 2 | contract | Clause (2) landed in W1(c), but the **nine-slot adjudication that makes it checkable** did not — it lives only in this spec's C1 cell, which becomes history on merge. Specifically lost: the measurement that shape (3)'s `«path»` is consumed positionally by `--cacheinfo` (clause (3), not clause (2)). A future author could read that slot as an option position and think the shipped FREE marker is wrong. | **residual → `WP-index-guard-residuals`.** Spec-sanctioned omission — the checklist required only the three clauses. Added to Discovered issues. |
| 3 | quality | W1(c) now states the membership predicate at two precisions in one cell: the sharp form in clause (1), the loose *"values THIS RUN PRODUCED"* in the *THE REPAIR DOES NOT INSPECT THE TOKEN* clause. | **residual, already registered.** This is the spec's explicit non-move with a named hand-off to `WP-index-guard-residuals` item 2; loose is not false. Noted so the successor need not re-derive it. |
| 4 | quality | The spec's threat-model sentence (`:1056-1062`) says the residual boundary is *"exactly one act: redirecting the test's require."* It understates: because **one constant serves both the require and the read**, a bare redirect makes the read follow the new file and the digest mismatches — the suite reds. The escaping act is redirect **plus** re-pin, which costs the same as edit plus re-pin. | **drop.** Error is in the safe direction (the check is *stronger* than the spec claims), and the sentence is in the spec, not the diff. |
| 5 | quality | The sweep's `MINTED` pattern went 0 → 2 in the Done spec. Both are correct (the quoted `b19121bb` subject as provenance, plus the sentence retiring the word), but the Mirrored Surface Checklist did not list them as expected hits, so a future gate will re-derive the adjudication. | **drop — informational.** Worth one checklist line next time that section is edited. |

Findings 1–3 are routed rather than fixed, per CLAUDE.md: *"Found something else
broken? Note it under Discovered issues; do not fix it."* None of them is reachable
inside this WP's permission boundary.

### Closed-Contract Drift Check — passed

No settled contract is silently reinterpreted. W1(c) stays the canonical decision
surface and is amended in place with an explicit date, package name and a statement
of what it replaced. The executable copy stays its registered pair, relocated with
the relocation recorded in the row itself. **No mirror is promoted to primary** — the
new module is the executable half of the existing pair, not a new authority, and the
digest constant is registered as a lockstep surface rather than as a copy of the set.

The one open drift risk is finding 1, and it is spec-shaped rather than
implementation-shaped.

**Reviewer's recommendation: merge.** Merging is the maintainer's act, not mine.

PR #203, opened 2026-09-02
test(dream): refuse to judge a relative private index and give PRODUCING its slot (WP-index-guard-residuals)

thegyula:
Spec: docs/specs/WP-index-guard-residuals.md

## Deliverables checklist

Files in the diff, and nothing else: `docs/specs/WP-index-guard-residuals.md` (the
status flip, always-allowed); `docs/specs/done/WP-dream-promote-in-workspace.md` (row
W1(c) only, all on line 541); `src/cli/dream.js` (comment text only, two sites);
`tests/unit/dream-pipeline.test.js` (six sites) — plus the round-1 gate raws under
`docs/specs/logbook/`, which `boundary-check` admits without listing.
`tests/unit/dream-pipeline.known-calls.js` is **not touched** and its byte-identity to
`main` is asserted below, so no digest re-pin is owed.

### The guard's own file, the whole suite, the lint pipeline

```
$ npm test -- tests/unit/dream-pipeline.test.js
guard-file suite exit=0
ℹ tests 44
ℹ pass 44
ℹ fail 0

$ npm test
npm test exit=0
ℹ tests 2444
ℹ pass 2432
ℹ fail 0
ℹ skipped 12

$ npm run lint
npm run lint exit=0
--- markdownlint --- Linting: 627 files. Summary: 0 issues in 0 files
--- shellcheck --- --- PSScriptAnalyzer --- --- frontmatter check ---
frontmatter check passed: 265 spec(s), 4 agent(s)
lint passed
```

`tests/integration/adopt-e2e.test.js` was **not** red here — 2432 pass / 0 fail, so the
dispatch's environmental exception was not needed.

### Criterion 3 — the module is untouched, so no re-pin is owed

```
$ git diff --quiet main -- tests/unit/dream-pipeline.known-calls.js && echo "byte-identical"
byte-identical
exit=0
$ git diff origin/main -- tests/unit/dream-pipeline.known-calls.js | wc -l
       0
```

### Criterion 4 — the loose wording is gone from Table W (GUARDED)

```
$ test -f <done-spec> && ! grep -qF -- 'values THIS RUN PRODUCED' <done-spec> \
    && echo "the loose PRODUCED wording is gone from Table W"
the loose PRODUCED wording is gone from Table W
exit=0
```

Both RED directions of that guarded grep, on scratch copies (proof the loose wording
stands in the copy under test: `grep -oF` printed `values THIS RUN PRODUCED`):

```
RED 1 — the loose sentence left standing:  exit=1 (no echo)
RED 2 — the file moved aside:              exit=1 (no echo — an absent file does not pass)
```

### Criteria 5 and 7 — the claim sweep

Baseline, re-run on this tree before any edit — 7 hits, matching the spec's `fc506110`
table hit for hit: Done spec `product code invokes git` ×1 and `values THIS RUN
PRODUCED` ×1; logbook `values THIS RUN PRODUCED` ×1; `src/cli/dream.js` `Every git
invocation this pipeline makes` ×1 and `The run's ONE git seam` ×2; test file `Every git
invocation the run` ×1 and `product code invokes git` ×1.

After (exit=0):

```
docs/specs/done/WP-dream-promote-in-workspace.md :: product code invokes git :: 1
docs/specs/logbook/2026-08-31-index-refresh-dropped-with-its-cause.md :: values THIS RUN PRODUCED :: 1
```

**What this establishes, as the spec requires:** the three registered wordings are gone
and the Table W hit of `values THIS RUN PRODUCED` is gone; `product code invokes git`
dropped from two to the Done spec's `:1114` alone (confirmed by `grep -n`), so the test
file's occurrence — the TITLE row 5 renames — is the one that moved. The logbook hit is
the registered non-move, unchanged. **It cannot certify the concept** — a rewording
scores zero hits — so criterion 5's pasted blocks and criterion 6's re-read carry that.

### The permission boundary, on the committed diff

```
$ node scripts/boundary-check.js docs/specs/WP-index-guard-residuals.md $(git diff --name-only main...HEAD)
boundary-check exit=0
```

---

## Criterion 1 — the refusal, in both directions, with its diagnostic in full

Captured by temporarily injecting `console.log` calls into the canary block, running the
guard-file suite, and restoring the committed bytes (injection proven applied:
`grep -c 'EVIDENCE-'` = 6 before, 0 after; tree clean).

```
EVIDENCE-1 classify(private shape, relative) -> Symbol(UNJUDGED — non-absolute private GIT_INDEX_FILE, refused rather than judged)
EVIDENCE-2 the raise >>>
UNJUDGED — non-absolute private GIT_INDEX_FILE, refused rather than judged:
    git write-tree
    [cwd=/var/folders/3v/02rwx2m56_b270xrhlf020080000gn/T/wd-testrun-75uajT/wd-pipe-Tj2jRf/vault, GIT_INDEX_FILE=rel.idx (RAW; no resolved path — none was computed)]
<<< EVIDENCE-2 end
EVIDENCE-3 classify(private, absolute = user index) -> "private index IS the user's index"
EVIDENCE-4 classify(private, absolute inside vault) -> "private index lies inside the user's working tree"
EVIDENCE-5 classify(private, absolute legit) -> null
EVIDENCE-6 classify(UNSET shape, relative) -> "known shape carrying an unexpected GIT_INDEX_FILE"
```

The refusal **errors** rather than returning a verdict and is a `Symbol`, so it cannot
be read as a fifth verdict string; the raise carries the **full `args`, the `cwd`, the
RAW value and the UNJUDGED state** and reports **no resolved path**; an **absolute**
value is still judged, reaching **both** private-index verdicts (3, 4) and still
admitting the legitimate one (5); and an **`unset`** shape with a non-absolute value
keeps its existing verdict (6), so no verdict becomes unreachable for a subclass.

**The required RED.** With `if (!path.isAbsolute(gif)) return UNJUDGED;` removed
(mutation proven applied by grepping the injected `// MUTATION-REMOVED-REFUSAL`
marker), the index test goes **0 pass / 3 fail** across all three layouts:

```
✖ …the run does not touch the user's git index — at all, plain vault (row G8)
✖ …separate-git-dir vault (row G8)      ✖ …linked-worktree vault (row G8)
ℹ tests 44   ℹ pass 41   ℹ fail 3
  AssertionError: a non-absolute private GIT_INDEX_FILE is REFUSED, not judged: the decision
  returns a distinguishable UNJUDGED state, never a verdict string — the four failure modes
  gain no fifth member
    actual: 'object',   expected: 'symbol',
```

`actual: 'object'` is the gap itself: without the refusal, `realish('rel.idx')` resolves
in the **Node** process's frame, lands outside the vault, and `classify` returns `null`
— **admitted**.

## Criterion 2 — the accept side stays alive

The index test is green on all three vault layouts in the 44/44 run above, and each
asserts `assert.deepEqual(watch.violations, [])` after a whole real run whose own
`GIT_INDEX_FILE` is the absolute `<stateDir>/dream-index.<pid>.tmp`. EVIDENCE-5 adds the
isolated admission. A guard that refused everything would satisfy criterion 1 alone and
fail here.

## Criterion 3 — nothing outside the edited sites moved

`classify`'s **whole body**, `main` vs `HEAD` — one added line, the signature and all
four verdict strings untouched:

```
$ diff <(git show origin/main:…test.js | sed -n '/const classify = /,/^  };$/p') \
       <(git show HEAD:…test.js        | sed -n '/const classify = /,/^  };$/p')
6a7
>       if (!path.isAbsolute(gif)) return UNJUDGED;
```

Occurrence counts on committed bytes, `main` → `head`: `function shapeMatches` 1→1,
`function realish` 1→1, `const classify = (args, env) =>` 1→1, `const
KNOWN_CALLS_MODULE` 1→1, `KNOWN_CALLS_SOURCE_DIGEST =` 1→1, `'b770de2e` 1→1, `known
shape missing its private GIT_INDEX_FILE` 1→1, `UNKNOWN SHAPE` 4→4 — all **SAME**. The
three that read 1→2 (`known shape carrying an unexpected GIT_INDEX_FILE`, `private
index IS the user's index`, `private index lies inside the user's working tree`) gained
their second occurrence as the **new canaries' expected values**; the strings
`classify` *returns* are the single occurrences inside the body above. No fifth verdict
is added.

The five `classify` call sites, each still passing an absolute or absent value:

```
1624:      watch.classify(['read-tree', '--index-output=/tmp/anywhere', 'HEAD'], {}),
1637:      watch.classify(['read-tree', `--index-output=${indexFile}`], { GIT_INDEX_FILE: '/tmp/legit-private.idx' }),
1652:      watch.classify(['show', `--output=${indexFile}`], {}),
1659:      watch.classify(['show', 'HEAD:reports/warnings.md'], {}), null,
1664:      watch.classify(['rev-parse', 'HEAD'], {}), null,
```

Every test title, `main` vs `HEAD` — 41 on each side, exactly one changed:

```
4c4
< test('dream-pipeline: claim-2b-pipeline — no product code invokes git with a cwd at or beneath the workspace root, over a whole run', async () => {
---
> test('dream-pipeline: claim-2b-pipeline — no git call the seam OBSERVES runs with a cwd at or beneath the workspace root, over a whole run', async () => {
```

And the pinned command at `WP-dream-promote-in-workspace.md:1454` still selects it:
`npm test -- --test-name-pattern "claim-2b-pipeline"` → `✔ dream-pipeline:
claim-2b-pipeline — no git call the seam OBSERVES runs with a cwd at or beneath the
workspace root, over a whole run (671.6ms)`, `ℹ fail 0`.

## Criterion 4 — the row's markers against the module, SHAPE BY SHAPE

The verifier derives **both sides independently** and carries **no expected mapping**:
side A is `require()`d from the module, each entry's args rendered into slot kinds and
literals (`RUN_VALUE`→`OWN`, `ANY`→`FREE`, strings as themselves); side B is parsed out
of the committed markdown line's physical entries. Neither side is consulted while
building the other. Run against **committed bytes**:

```
shape-by-shape (module order):
  (1) unset   —         agree  ls-tree OWN -- FREE
  (2) unset   PRODUCING agree  hash-object -w --stdin
  (3) private —         agree  update-index --add --cacheinfo FREE OWN FREE
  (4) unset   —         agree  show HEAD:reports/warnings.md
  (5) unset   PRODUCING agree  rev-parse HEAD
  (6) private —         agree  read-tree OWN
  (7) private PRODUCING agree  write-tree
  (8) unset   PRODUCING agree  -c user.name=wienerdog -c user.email=wienerdog@localhost commit-tree OWN -p OWN -m FREE
  (9) unset   —         agree  update-ref -m FREE HEAD OWN OWN

GREEN — nine entries, labels 1..9 in the module's physical order, no decoy,
every marker inside its own entry, and each shape's PRODUCING state agrees.
verifier exit=0
```

Which shapes carry the marker was **derived from the module by this work** — (2)
`hash-object`, (5) `rev-parse HEAD`, (7) `write-tree`, (8) `commit-tree` — never
predicted from the spec.

**The four required REDs**, each on a scratch copy with the mutation proven applied by
grepping the mutated bytes before running:

*(a) same-count swap* — the marker moved off `write-tree` onto `read-tree`, count still
four. Proof: ``​`read-tree «own tree-ish»` (`:234`) — **PRODUCING**.`` and
``​`write-tree` (`:258`).``

```
RED — the row and the module do NOT agree:
  * position 6 (read-tree OWN): PRODUCING differs — module produces=false, row marker=true
  * position 7 (write-tree): PRODUCING differs — module produces=true, row marker=false
exit=1
```

*(b) label swap* — `**(2)**` ↔ `**(5)**`. Proof: ``**(5)** `unset` — `hash-object`` and
``**(2)** `unset` — `rev-parse``. **exit=1** — `STRUCTURE: entry at position 2 is
labelled (5) … each once` and the mirror finding at position 5.

*(c) decoy ordinal* — a tenth entry-shaped construct. Proof: ``**(10)** `unset` —
`fsck``. **exit=1** — `STRUCTURE: 10 entries in the enumeration, expected 9` and
`position 10: present on only one side`.

*(d) module-side `produces` flip — the one that proves the module is actually read.*
`produces: true` removed from `write-tree` in a scratch copy of the module; **the row
is not changed at all**. Proof: `{ env: 'private', args: ['write-tree'] },`

```
  (7) private —         DISAGREE  write-tree
RED — the row and the module do NOT agree:
  * position 7 (write-tree): PRODUCING differs — module produces=false, row marker=true
exit=1
```

*(e) the inventory's other RED — a byte added to the module.* Proof:
`// MUTATION-BYTE-ADDED` appended, then removed and the file re-verified byte-identical
to `origin/main`. **43 pass / 1 fail** — `✖ dream-pipeline: the pinned call set module
is its canonical SOURCE FORM (Table W row W1(c))`, `AssertionError: … not its canonical
source form. Any edit to it re-pins KNOWN_CALLS_SOURCE_DIGEST in the SAME commit as the
Table W row W1(c) change.`

## Criterion 4a — no count of own-value sources in any spec or docs prose surface

Row W1(c) states none: the REPAIR clause now cites clause (1)'s predicate and the
shapes the set marks, the ORDERING paragraph states only the ordering, and the closure
clause's no-count rule stands. A repo scan for a count-plus-`own-value`/`produc*`
collocation in `*.md` returns only:

- `WP-show-slot-own-value-kind.md:106`, `:576` (inside ```js fences, quoting the
  module's JSDoc verbatim) and `:724` (quoting the test file's recorder comment) —
  **code, exemption (i)**; that spec is also the predecessor, excluded from the sweep
  because it quotes pre-change wording on purpose.
- `docs/specs/logbook/2026-09-01-index-guard-residuals-design-gate-rounds.md:67` and
  `memory/lessons/inbox.md:1276` — dated, past-tense provenance recording that the
  count *was* stated, the exempt class.

`known-calls.js:68` and `dream-pipeline.test.js:257` are code and untouchable here, so
neither is a defect — which is exactly what the criterion says.

**Correction (round-1 finding 4).** The scan missed
**`WP-show-slot-own-value-kind.md:690`** — row C3's heading *"The count of own-value
sources is FOUR"*, whose body also still names *"the `produces` markers"* as the owning
surface, which W1(c) now supersedes. That is a **prose** count, not an exempt one, so
4a is **not clean repo-wide**. It is a superseded Done spec's change-order table, out
of boundary (not a Deliverable), and its `test.js:192, :195, :197, :198-199` citations
were already dead before this diff. Recorded as **Discovered issue 4** rather than left
asserted clean. Criterion 4a's own scope does hold: row W1(c) states no count.

## Criterion 5 — Table C has landed, row 0 included

Each edited comment block, pasted whole from committed bytes, each carrying a
`row W1(c)` citation **within its own block**.

`src/cli/dream.js` — `gitIn`'s JSDoc:

```js
/**
 * THIS FILE'S git seam (CLAIM 2b): total over `src/cli/dream.js`, NOT over the dream
 * path — a LIMIT row W1(c)'s COVERAGE clause states and owns. Every invocation goes
 * through here, so a test can substitute it and observe every `cwd` — which is what
 * the pipeline-wide "no git with a cwd at or beneath the workspace root" assertion
 * needs. A source grep cannot discriminate: it is green today, green on a correct
 * implementation, and green on a broken one that passes the path through a variable.
 * @param … (the three @param lines are unchanged)
 */
```

`src/cli/dream.js:560-561`:

```js
  // THIS FILE'S git seam (CLAIM 2b) — not the run's only one; see `gitIn` and
  // row W1(c). JS-only: production passes no opts, so the pinned door always runs.
```

`tests/unit/dream-pipeline.test.js` — `watchIndexWrites`'s JSDoc:

```js
 * W1(a) is a total over the package's OWN acts. Every git invocation THE SEAM
 * OBSERVES must match one of KNOWN_CALLS exactly; an unrecognised shape is a
 * violation by default rather than something to be understood. The seam is
 * total over `src/cli/dream.js` and NOT over the dream path — row W1(c)'s
 * COVERAGE clause (`docs/specs/done/WP-dream-promote-in-workspace.md:541`)
 * states that limit, names the spawn points outside it, and owns the list; it
 * is cited here and never restated. There is no question here about which
 * repository a call reaches or which index it would write — those questions
 * were the retired direction, and both were refuted by measurement (see
 * KNOWN_CALLS).
```

`tests/unit/dream-pipeline.test.js` — the `claim-2b-pipeline` comment:

```js
  // THE SEAM, NOT A GREP. A source grep for a workspace-rooted cwd cannot
  // discriminate: it is green today, green on a correct implementation, and
  // green on a broken one that passes the path through a variable.
  // WHAT THIS EVIDENCE REACHES: the calls that arrive through the injected
  // `opts.spawnGit`, and no others — row W1(c)'s COVERAGE clause
  // (`docs/specs/done/WP-dream-promote-in-workspace.md:541`) states that limit
  // and owns it, and it is cited rather than restated here.
```

Row 0 — clause (a) SCOPE's parenthetical, in the canonical cell:

> …the run's OWN: **its own git invocations — those the seam of (c) observes and those
> it does not — and its own file writes.** **The seam is the ENFORCEMENT, never the
> boundary of the scope:** its reach is a LIMIT that (c)'s COVERAGE clause states and
> owns, cited here rather than restated, because the boundary of (a) is AUTHORSHIP, not
> visibility.

No spawn-point list is restated at any of the five sites. The renamed title states only
what the test observes and keeps its `claim-2b-pipeline` token (criterion 3). Both
Deliverables cells' unchanged lists hold.

## Criterion 6 — the whole cell and each comment block were re-read after rewriting

**What the re-read found — it was not "nothing".**

1. **A falsified citation class, caught and prevented.** The first pass of the
   `src/cli/dream.js` comment edits added seven lines. Row W1(c) carries ~15
   `cli/dream.js:NNN` citations — the nine pinned-set call sites among them — every one
   of which would have silently rotted. Both edits were **redone line-count-neutral**
   (1206 lines on both sides), and all seventeen cited anchors re-read and confirmed to
   still land (`:166 :178 :221 :230 :234 :238 :253 :256 :258 :261 :263 :562 :587 :941
   :1004 :1041 :1042`). Repairing the citations instead would have been an eighth edit,
   out of boundary.
2. **A sentence the private-arm change would otherwise have falsified.** The
   `watchIndexWrites` JSDoc's *"Both clauses are enforced"* is now qualified in place:
   *"Both clauses apply only to an ABSOLUTE value; a non-absolute one is REFUSED
   UNJUDGED and raises."*
3. **A ragged reflow** in that same JSDoc (*"There is no / question here about…"*),
   found by re-reading the committed block. Folded into the same commit.

**What the re-read confirmed still true after the rewrite** (each checked against the
new sentences rather than assumed):

- *"There is no question here about which repository a call reaches"* and *"THE TWO
  TRAPS THAT ARE GONE … asking git which repository an invocation reaches"* — **still
  true**: no locator, no out-of-band call, no environment replay. That is the point of
  refusing rather than resolving.
- *"The verdict text is unchanged and the FOUR FAILURE MODES below gain no fifth
  member"* — **still true**; the refusal is a Symbol and an error.
- *"AND EVERY VERDICT RED CARRIES ITS INVOCATION … and where it resolved"* — no longer
  falsifiable by an outcome that resolves nothing, being scoped to verdicts with the
  refusal's diagnostic beside it.
- Clause (a)'s *"the boundary is AUTHORSHIP, not visibility"* and its standing example
  *"(c)(i)'s un-seamed `validate.js` invocation"* previously sat three clauses after a
  parenthetical asserting the opposite. **That contradiction is gone.**
- The closure clause's SHA-pinned remedy record (*"SHIPPED AT `b19121bb`"*) is a
  registered NON-move and is **unchanged**.
- That spec's own Mirrored Surface Checklist (`:548-1017`) is **untouched** — the whole
  diff to the file is one hunk, `@@ -541 +541 @@`.

**Round-1 addendum — what the re-read missed, and the fix.** wd-reviewer finding 1
(band B, in boundary) caught a sentence I added and then read past: the marker
definition ended *"— the pair is registered in the Mirrored Surface Checklist exactly
as `«own …»` ↔ `RUN_VALUE` is."* It is **not** — `:548-1017` registers only the
slot-kind pair, at `:685` — and this WP may not edit that checklist to make it true.
The clause was additive (Table B's preamble row asks only for the definition, the
executable spelling and shape-by-shape agreement) and it was drift on day one in the
canonical cell, exactly what criterion 6 forbids. Fixed in `2bb3305a`: it now reads
*"— the pair moves together exactly as `«own …»` ↔ `RUN_VALUE` does."* Re-read after
that edit: the cell carries **exactly one** "Mirrored Surface Checklist" mention, the
pre-existing slot-kind one, and it is true.

Re-runs after the fix, each exit code its own statement: criterion-4 verifier on
committed bytes **exit 0** (9/9 agree — the fixed sentence sits inside the parsed
region, so this also proves the preamble still parses); same-count-swap **exit 1**
(RED, positions 6 and 7); sweep **exit 0**, same two registered non-moves;
`npm test -- tests/unit/dream-pipeline.test.js` **exit 0** (44/44); `npm run lint`
**exit 0**; `boundary-check` **exit 0**; `git diff --quiet main -- …known-calls.js`
**exit 0**. The Done spec's diff is still one hunk, `@@ -541 +541 @@`, 1 insertion /
1 deletion.

## Criteria 7, 8, 9

Sweep re-run above; every remaining hit is a registered non-move (the Done spec's
`:1114`, a NON-move by the Mirrored Surface Checklist, and the 2026-08-31 logbook
line). `npm test`, `npm run lint` and `boundary-check` all above with their exit codes.
Idempotency: **N/A — this WP ships no command and writes nothing outside the
repository.**

---

Run by the orchestrator on 2026-09-02 against the current tree (a strict descendant of
the spec's baseline `fc506110`, zero commits touching the cited files): **40 executable
claims re-run, 39 PASS, 0 STALE, 1 UNRUNNABLE** (Frame 2's symlink sub-claim,
provenance-only, no literal command). Confirmed exact: `dream.js:226-230` and the
`:156-157` / `:560` comment quotes; in `dream-pipeline.test.js` — `:128-131` `realish`,
`:167-168` JSDoc, `:179-246` `watchIndexWrites`, `:196-208` `classify`, `:202`, `:436`
the title, `:443-445`, and the five `classify` call sites `:1589` / `:1602` / `:1617` /
`:1624` / `:1629`, each passing an absolute or absent value; nine pinned shapes with 4
of 9 carrying `produces: true` (**derived here from the module, not taken from the
dispatch message** — see criterion 4); W1(c) all on line 541 (46325 bytes, now 49070),
bold `**PRODUCING**` occurring 0 times there before this change; the sweep baseline
matched hit-for-hit (pasted above); both frame falsifications reproduced on git 2.39.5
/ Node v25.9.0.

**Owner ruling in force** (`docs/specs/logbook/2026-09-02-owner-rulings-stub-queue.md`,
on the PR #201 branch): the Dispatch precondition's single ratification is **GRANTED** —
refusing to judge a non-absolute `private` `GIT_INDEX_FILE` (UNJUDGED from `classify`,
harness ERROR at the seam wrapper) is **not** a fifth failure mode. Table A is
implemented exactly as written; the WP did not return to design.

## Decisions made

1. **UNJUDGED is a `Symbol` declared inside `watchIndexWrites`, above `classify`.** The
   spec requires "a distinguishable UNJUDGED state" and leaves the mechanics to the
   implementer. A Symbol **structurally cannot** be mistaken for a fifth verdict string,
   which is the precise distinction the owner ratified; a sentinel string would sit in
   the same domain as the four verdicts and read like one. Declaring it inside the
   function keeps both halves of the pair in one scope without touching
   `watchIndexWrites`'s return shape or `classify`'s signature.
2. **Both `src/cli/dream.js` comment edits are line-count neutral**, and the second
   drops "front" from "the pinned front door always runs" to fit two lines. Reason under
   criterion 6 item 1: the alternative was rotting ~15 `cli/dream.js:NNN` citations in
   the canonical cell, and repairing those would have been an eighth, out-of-boundary
   edit.
3. **One sentence added to the `watchIndexWrites` JSDoc's second paragraph**, beyond the
   `:167-168` scope of Table C row 3: *"Both clauses are enforced"* is falsified by the
   private-arm change unless qualified, and criterion 6 forbids leaving a sentence the
   rewrite falsified. Same comment block; keeping it true in place was the simpler
   option.
4. **The canaries use the `write-tree` shape** for every private-arm vector — the one
   private shape with no slots, so they reach the disposition arm under test without
   depending on what the run put in the own-value set (the arity trap W1(c) records).
5. **The criterion-4 verifier lives in the scratchpad, not the repo.** The spec asks for
   pasted evidence from a comparison, not a shipped checker; adding one to `tests/` or
   `scripts/` would be out of boundary.
6. **Round-1 finding 1 was fixed by deletion, not by making the claim true.** Editing
   `:548-1017` to register the PRODUCING pair would have been out of boundary and a
   change to a settled checklist; dropping the clause keeps the analogy on the ground
   Table B actually gives it. Registering the pair is a wd-architect change order.

1. **`src/core/paths.js` does not validate `HOME`.** A relative `HOME` yields a relative
   `paths.state`, therefore a relative `GIT_INDEX_FILE` (measured:
   `getPaths({ HOME: 'relhome' }).state === 'relhome/.wienerdog/state'`). It is a
   product-side question about an unsupported configuration, and it is the one
   producible case on which Table A's refusal would now stop the suite rather than
   silently judging it in the wrong frame. `$WIENERDOG_HOME` is forced absolute by
   `assertSafeOverride` (`src/core/paths.js:21-31`); `HOME` is deliberately unvalidated
   (`:7-10`).
2. **`docs/specs/done/WP-dream-promote-in-workspace.md:673`** still registers the
   executable copy of the pinned set as living in `tests/unit/dream-pipeline.test.js`.
   It moved to `tests/unit/dream-pipeline.known-calls.js`, and row W1(c) already says
   so. That line is in that spec's own Mirrored Surface Checklist (`:548-1017`), which
   this WP's Deliverables cell excludes by name.
3. **Nine positional citations rotted by this diff** (round-1 finding 2, band B) —
   `WP-show-slot-own-value-kind.md:265, :276, :350, :367, :370, :393, :708, :716, :724`
   cite `tests/unit/dream-pipeline.test.js` by line, and that file grew ~89 lines above
   and inside every cited range. The growth is spec-mandated (Deliverables site (c), the
   canaries criterion 1 requires) and the citing file is a Done spec, not a Deliverable
   here, so this is **not fixable inside this WP**. Verified anchor by anchor, main vs
   HEAD: only `:30` still lands. The structural fix — positional citations into a live
   test file, which that spec's own checklist condemns at `:558-560` — is a
   **wd-architect** change.
4. **`WP-show-slot-own-value-kind.md:690`** (round-1 finding 4, band C) — row C3 still
   states *"The count of own-value sources is FOUR"* in prose and still names *"the
   `produces` markers"* as the owning surface, both superseded by W1(c). Out of boundary
   (superseded Done spec); it is why the 4a paragraph above carries a correction rather
   than a clean assertion. → **wd-architect**.
5. **`tests/unit/dream-pipeline.test.js:196-199`** (round-1 finding 3, band C) — the
   `computed` JSDoc still says *"every object name it passes was computed by an earlier
   pinned read"*, the sentence W1(c)'s ORDERING paragraph was rewritten in this very
   diff to retire (false twice over: `rev-parse HEAD` reads the head back; `hash-object
   -w` / `write-tree` / `commit-tree` are not reads). Outside the six sites and code,
   which 4a exempts, so leaving it is correct here — but it now contradicts the
   canonical cell. → **wd-architect**.
6. **Codex plugin P2, RESIDUAL with reason** — the plugin reads the narrowed (a)
   parenthetical as contradicting (c)(ii)'s *"outside this row's SCOPE by
   construction"*. (c)(ii) is byte-identical to `main` and says (a) *"never reaches"*
   `promote()`'s `merge-file` because it runs outside any repository — a vacuous case,
   not an exclusion — and the previous wording (*"every one of them through the seam"*)
   was the actually-false one Table C row 0 ordered removed. Altering (ii) would be an
   eighth edit outside the Deliverables cell. → **wd-architect**.

Issues 3–6 are one W1-wording family; the reviewer's recommended single follow-up is a
canonical-extraction pass on the pinned-set mirror registry.

---
Generated-by: claude-opus-5

## Lessons

- WP-index-guard-residuals: **a comment-only edit is not a no-op if the file is cited by
  line.** Rewriting two comments in `src/cli/dream.js` added seven lines and would have
  rotted ~15 `cli/dream.js:NNN` citations inside row W1(c) — the nine pinned-set call
  sites among them, i.e. the entries this WP was editing. The check is cheap:
  `git show origin/main:<f> | wc -l` against `wc -l <f>`, then `sed -n '<n>p'` on each
  cited anchor. **And it cuts both ways:** the *test* file's spec-mandated growth rotted
  nine citations in a Done spec that is out of boundary (Discovered issue 3), which
  line-count neutrality could not have prevented — that one needs the citations to stop
  being positional.
- WP-index-guard-residuals: **"derive both sides independently" is about where the data
  comes from, not how clever the script is.** The verifier `require()`s the module for
  one side and parses the committed markdown for the other, carrying no expected
  mapping — which is why the module-side `produces` flip reddens it. A count, an ordinal
  parser and a verifier with the mapping hard-coded were each measured to pass a wrong
  state; the mutation that catches the last of those changes only the module.
- WP-index-guard-residuals: **a Symbol is the right shape for "not a verdict".** Making
  the refusal a Symbol makes the owner's ratification structural rather than a naming
  convention — no assertion, grep or later reader can mistake it for one of the four
  verdict strings.
- WP-index-guard-residuals: **re-reading the whole 46 KB cell caught what no pattern
  could — and re-reading it once was not enough.** My own pass found clause (a)'s
  parenthetical contradicting its *"AUTHORSHIP, not visibility"* ruling three clauses
  later, and a JSDoc promise the code change falsified. It did **not** catch a sentence
  I had just written claiming a registration that does not exist (round-1 finding 1) —
  a writer re-reads their own new sentence as intent, not as text. The cheap guard is
  mechanical: for every "X is registered in Y" you add, open Y and grep it.


thegyula:
## Round-1 gate dispositions — PR #203

Gates ran on tip `5a4a79bd`; raws committed on this branch at `7544405a`
(`docs/specs/logbook/2026-09-02-index-guard-pr203-raw-round1-{codex-plugin,herdr-shadow,wd-reviewer}.txt`).
New tip after dispositions: **`2bb3305a`**.

| # | Channel | Band | Disposition | Fixing commit / reason |
|---|---|---|---|---|
| 1 | wd-reviewer | B | **FIX** | `2bb3305a` — the marker-definition sentence claimed the PRODUCING ↔ `produces: true` pair is "registered in the Mirrored Surface Checklist"; it is not (`:548-1017` registers only the slot-kind pair, at `:685`), and that checklist is out of boundary. The registration claim is dropped; the clause now reads "— the pair moves together exactly as `«own …»` ↔ `RUN_VALUE` does", which is the ground Table B's preamble row actually gives it. In boundary, line 541, still the only changed line of that file. |
| 2 | wd-reviewer | B | **RESIDUAL → wd-architect** | Out of boundary. Nine positional citations in `docs/specs/done/WP-show-slot-own-value-kind.md` (`:265, :276, :350, :367, :370, :393, :708, :716, :724`) into `tests/unit/dream-pipeline.test.js` rotted because that file grew ~89 lines — growth mandated by this WP's Deliverables site (c). The citing file is a Done spec and not a Deliverable here, so it cannot be repaired inside this WP. Recorded as **Discovered issue 3** in the PR body. |
| 3 | wd-reviewer | C | **RESIDUAL → wd-architect** | Out of boundary. `tests/unit/dream-pipeline.test.js:196-199`'s `computed` JSDoc still carries "computed by an earlier pinned read", the sentence W1(c)'s ORDERING paragraph was rewritten in this diff to retire. Outside the Deliverables cell's six sites, and code — which criterion 4a exempts. Recorded as **Discovered issue 5**. |
| 4 | wd-reviewer | C | **RESIDUAL → wd-architect**, PR body corrected | The criterion-4a paragraph missed `docs/specs/done/WP-show-slot-own-value-kind.md:690` (row C3's "The count of own-value sources is FOUR", plus its now-superseded "the `produces` markers" owning-surface sentence). That is a prose count, not an exempt one, so 4a is **not clean repo-wide**. The 4a paragraph now carries a correction naming it instead of a clean assertion; recorded as **Discovered issue 4**. Out of boundary — superseded Done spec, not a Deliverable. |
| 5 | Codex plugin (P2) | — | **RESIDUAL with reason → wd-architect** | Orchestrator spot-checked. The plugin reads the narrowed (a) parenthetical ("those the seam observes and those it does not") as contradicting (c)(ii)'s "outside this row's SCOPE by construction". (c)(ii) is byte-identical to `main` and says (a) "never reaches" `promote()`'s `merge-file` because it runs outside any repository — a vacuous case, not an exclusion — and the previous wording ("every one of them through the seam") was the actually-false one Table C row 0 ordered removed. Altering (ii) would be an eighth edit outside the Deliverables cell. Recorded as **Discovered issue 6**. |
| — | herdr shadow | — | **CLEAN** | Zero findings; all four criterion-4 REDs reproduced independently, plus the refusal-removed and byte-added-to-module REDs. |

Findings 2–5 are one W1-wording family (positional citations and prose counts scattered
across superseded Done specs). The reviewer's recommendation, adopted here: route them
to **wd-architect** as a single canonical-extraction pass on the pinned-set mirror
registry rather than patching piecemeal.

### Re-run evidence after the finding-1 fix (each exit code its own statement)

```
criterion-4 verifier, committed bytes                exit=0   9/9 agree, no decoy, labels 1..9
  (the fixed sentence is inside the parsed region, so this also proves the preamble parses)
criterion-4 same-count-swap mutation                 exit=1   RED — positions 6 and 7
5-pattern claim sweep                                exit=0   same two registered non-moves
npm test -- tests/unit/dream-pipeline.test.js        exit=0   44 pass / 0 fail
npm run lint                                         exit=0   lint passed
node scripts/boundary-check.js …                     exit=0
git diff --quiet main -- …known-calls.js             exit=0   byte-identical, no re-pin owed
```

Done-spec diff shape is unchanged: one hunk, `@@ -541 +541 @@`, 1 insertion / 1
deletion; `:548-1017` untouched. Whole-cell re-read after the fix: the cell now carries
**exactly one** "Mirrored Surface Checklist" mention — the pre-existing slot-kind one —
and it is true.

Gates not re-run and not merged, per dispatch.

thegyula:
## Round 2 dispositions (tip 2bb3305a; raws committed at 6697ade3 and after)

| Channel | Finding | Band | Disposition |
|---|---|---|---|
| wd-reviewer | round-1 finding 1 (false "registered in the checklist" claim) | B | **verified FIXED** — independent verifier green on committed line 541, module-side `produces` flip still RED; verdict **APPROVE** |
| wd-reviewer | round-1 findings 2–4 | B/C/C | **verified recorded** as Discovered issues 3–5, routed to wd-architect as one canonical-extraction pass on the W1 wording family |
| Codex plugin | W1(a) parenthetical vs (c)(ii) "outside this row's SCOPE" — re-raised verbatim | C | **residual, concurred by wd-reviewer with the sharpened reason**: (c)(ii)'s exclusion is grounded in vacuity ("so (a) never reaches it"); the (c)(i)↔(c)(ii) framing disagreement is pre-existing on main; deferring (a)'s scope to (c)'s classifications would reinstate visibility as the boundary, which Table C row 0 forbids; editing (ii) is an eighth edit outside the Deliverables cell. Routed to wd-architect with the same follow-up (settle whether (c)(ii) should read "vacuously satisfies (a)") |
| hermetic shadow | none | — | patch is correct; suite 2444/0, focused 44/0, boundary and byte-identity gates exit 0 |

Both gates have run twice on the diff; every finding is fixed or dispositioned. Merge is the maintainer's.

PR #205, opened 2026-09-02
fix(dream): widen the only-copy abort trigger to its class (WP-preservation-abort-widening)

thegyula:
Spec: docs/specs/WP-preservation-abort-widening.md

## Deliverables checklist

  (`src/core/dream/validate.js`, `src/core/dream/promote.js`,
  `src/cli/dream.js`, `tests/unit/dream-validate.test.js`,
  `tests/unit/dream-promote.test.js`,
  `docs/specs/done/WP-dream-promote-in-workspace.md`,
  `docs/specs/done/WP-secret-fence-ep2-redact-arm.md`) — verified below with
  `node scripts/boundary-check.js`.

## Verification output

All commands run against the final committed tree (`git status --short`
clean; working tree == HEAD).

```
$ npm test
...
ℹ tests 2469
ℹ suites 0
ℹ pass 2457
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 45041.464916
rc=0

$ npm run lint
...
--- markdownlint ---
Summary: 0 issues in 0 files
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 265 spec(s), 4 agent(s)
lint passed
rc=0
```

(Re-pasted at round 3, tip `a7f24032` — the tree grew by two tests
[the round-3 partial-write RED and the enum-pollution RED] since the
initial V1 paste; counts moved from 2465/2453 accordingly.)

### V2 — P1/P2's message value

```
$ test "$(grep -oF 'no redaction copy was attempted' src/core/dream/validate.js | wc -l | tr -d ' ')" = 1
rc=0
$ test "$(grep -oF 'no redaction copy was attempted' tests/unit/dream-validate.test.js | wc -l | tr -d ' ')" -ge 1
rc=0
```

### V3 — P4's message, redact wording not reused

```
$ test "$(grep -oF "withhold arm reported no preserved copy" src/core/dream/promote.js | wc -l | tr -d ' ')" = 1
rc=0
$ test "$(grep -oF "redact arm reported no preserved copy" src/core/dream/promote.js | wc -l | tr -d ' ')" = 1
rc=0
$ test "$(grep -oF "withhold arm reported no preserved copy" tests/unit/dream-promote.test.js | wc -l | tr -d ' ')" -ge 1
rc=0
```

### V4 — the `dream.js` promote-catch comment, byte-exact against the spec

```
$ cb() { awk '/res = promote\(\{/{f=1} f{print} f && /throw err;/{exit}' src/cli/dream.js; }
$ cb | sed -n '/} catch (err) {/,$p' | grep '^ *//' > /tmp/wd-catch-actual.txt
$ sed -n '/^        \/\/ ROW G5.s SECOND teardown exception/,/^        \/\/ loud with the tree intact\.$/p' \
    docs/specs/WP-preservation-abort-widening.md > /tmp/wd-catch-expected.txt
$ test "$(wc -l < /tmp/wd-catch-expected.txt | tr -d ' ')" = 10
rc=0   # the literal was found in the spec
$ diff -u /tmp/wd-catch-expected.txt /tmp/wd-catch-actual.txt
rc=0   # COMPLIANT — identical
```

**All three states observed** (measured live during this WP):
- **ABSENT** (`dream.js` renamed away): `diff` reports the expected 10 lines
  all missing (`-` lines only) — RED.
- **COMPLIANT** (current tree, shown above): `diff` empty — GREEN.
- **VIOLATING** (`The run fails` reworded to `The build fails`): `diff`
  reports a one-line mismatch — RED.

### V5 — counterfactual evidence

Baseline is the GREEN full suite above (`npm test` on the finished state,
2453 pass / 0 fail). For each behaviour, the tree was mutated with an
inline marker comment (`V5-MUTATION-N`), the marker was grepped to confirm
it landed, `npm test` was run in full (no name filter), the failure output
was captured, and the mutation was reverted byte-for-byte (`diff` against a
pre-mutation backup showed `IDENTICAL` after every revert; the branch's
final `git diff` matches the pre-V5 diff exactly).

1. **Trigger narrowed back to the redact fall-through** (`validate.js`
   condition reverted from `!preserved && !redactCopy` to
   `redactFellThrough && !preserved`): **8 failing**, all in this WP's own
   `tests/unit/dream-validate.test.js` —
   `EP2 aborts (P1) when the quarantine copy cannot be written`,
   `EP2 aborts (P1) without a stateDir`,
   `EP2 redact arm R0b (untracked/tracked): a durable copy EXISTS...`,
   `EP2 hard-secret withhold arm (P1, untracked/tracked)`,
   `EP2 unscannable withhold arm (P2, binary/NOT-lossless-UTF-8 cause)`.
   Each names `driveAbort`'s `the gate must have thrown` assertion.
2. **P4's refuse-branch check removed** (`promote.js`): **2 failing**, both
   in `tests/unit/dream-promote.test.js` — `dream-promote P4: a refuse
   verdict carrying an empty preservation record is refused fail-loud...`
   and `...with NO preserved field at all...` (the `assert.throws` no
   longer sees a throw).
3. **P6's two active values collided** (`'no-redaction-attempted'` mapped
   to the same text as `'both-failed'`): **6 failing**, all in
   `tests/unit/dream-validate.test.js` — the four P1/P2 abort tests and the
   two rewritten `dream-validate: EP2 aborts (P1)...` tests, each failing
   `assertAbort`'s which-value / pairwise-absence checks.
4. **P0b's read-back removed** (`quarantinePreserve` always returns
   `{name, bytes: content}` after rename, no verification): **6 failing** —
   `EP2 redact arm (P0b regression): a CORRUPTED redacted/ artifact...` and
   five `quarantinePreserve (P0b...)` / `(Table D row D2/D3/D4)` unit tests.
5. **Table D disposal suppressed** (`removeOwnedQuarantinePath` reverted to
   a swallowed `try/catch` with no absence-confirmation, no throw): **2
   failing** — both `quarantinePreserve (Table D row D3): ... fails LOUD`
   tests (tmp and dest paths).

Every red names a failing assertion inside this WP's own changed test
files, satisfying the non-vacuity requirement.

**Two more counterfactuals, added across rounds 1 and 3 as `quarantinePreserve`'s
temp-file ownership tightened further** (both independently confirmed red
against their respective pre-fix shapes and green against the fix in
place, with a byte-identical revert afterward each time):

6. **Exclusive create (`flag: 'wx'`) removed** (round-1 herdr-shadow fix,
   `40225aff`) — reverting the temp write back to the default overwrite
   flag: **2 failing**, both `quarantinePreserve (Table D row D1,
   herdr-shadow round 1 P1): ...` tests. The "otherwise-happy path" test
   showed the primitive returning a SUCCESSFUL result with the seeded
   foreign bytes silently overwritten and renamed away — the exact bug a
   crash + reused pid would trigger in production.
7. **Ownership-on-partial-write fix removed** (round-3 wd-reviewer /
   codex-plugin finding, `a7f24032`) — reverting `tmpOwned = true` back to
   being set only after the whole `writeFileSync('wx')` call returns: **1
   failing** — `quarantinePreserve (Table D row D1, round-3 review): a
   failure AFTER the exclusive create still removes this invocation's OWN
   partial tmp file`. The test's `writeFileSync` wrapper calls THROUGH to
   the real exclusive create (so the directory entry genuinely exists on
   disk) before throwing an injected `ENOSPC`, and on the pre-fix shape the
   assertion failed with `actual: true` where `false` (absent) was
   expected — the partial tmp file the invocation itself created was left
   behind, exactly the shape wd-reviewer reproduced live with `ulimit -f 1`
   and `SIGXFSZ` ignored (kernel `EFBIG` after `O_EXCL` had already created
   the file).

### Boundary check (committed bytes)

```
$ node scripts/boundary-check.js docs/specs/WP-preservation-abort-widening.md $(git diff --name-only origin/main...HEAD)
rc=0
```

```
$ git status --short
(empty — working tree matches HEAD)
$ git log --oneline origin/main..HEAD
a7f24032 fix(dream): tighten preservation temp-file ownership and the abort-message enum lookup (WP-preservation-abort-widening)
40225aff fix(dream): create the preservation temp file exclusively (WP-preservation-abort-widening)
c7ff8f12 docs(dream): correct the gate's stale @throws JSDoc (WP-preservation-abort-widening)
70a639c0 test(dream): non-vacuous hostile-path evidence for the widened abort message (WP-preservation-abort-widening)
73d39030 test(dream): direct coverage for secretGateAbortMessage's pair rule (WP-preservation-abort-widening)
3c336b3d docs(specs): flip status to In-Review (WP-preservation-abort-widening)
99960a75 docs(specs): amend G5/Q18/B3b for the widened only-copy trigger class (WP-preservation-abort-widening)
aea7690e test(dream): cover Table P's widened trigger and Table D's disposal, correct falsified P0b commentary (WP-preservation-abort-widening)
0dec6586 fix(dream): widen the only-copy abort trigger to Table P's class (WP-preservation-abort-widening)
```
(three raw-transcript logbook commits from the review rounds sit between
these — `docs/specs/logbook/` is in the always-allowed set and does not
appear in this listing's Deliverables-boundary check)

`tests/integration/adopt-e2e.test.js` was included in the full `npm test`
run above and passed — the machine-local pin issue the spec flags did not
reproduce here.

## Amended-cell re-read report

Per the spec's Implementation notes: *"re-read each amended cell WHOLE and
report, in the PR body, any sentence the clause leaves false."* Done for
all three amended rows.

**G5** (`docs/specs/done/WP-dream-promote-in-workspace.md:465`) — re-read
whole. **No sentence left false.** The appended clause explicitly
supersedes the one sentence it needed to (*"not the single named case
above"*, referring back to *"a note whose redaction and whose withheld
preservation both failed"*). Every other sentence in the cell — G2's
teardown exception, "teardown never touches the vault", the
residue-lifecycle successor note, Table V row V9's scratch/lock ordering —
is orthogonal to the trigger class and remains true unchanged.

**Q18** (`docs/specs/done/WP-secret-fence-ep2-redact-arm.md:1647`) —
re-read whole. **No sentence left false.** The clause scopes *"every
statement in this row and its holder list that quantifies over 'all three
arms' or over the R0/R0b enumeration"* — which covers column 3's
*"whenever the `redacted/` copy survives, its basename"* field description
and column 4's per-arm holder list (the R0/R0b-keyed assertions for the
identity disposition and basename presence/absence). Every such statement
is now historical — scoped to the arms that existed when it was written —
not a claim about current reachability. Nothing else in the row asserts a
current fact the clause falsifies.

**B3b** (`docs/specs/done/WP-secret-fence-ep2-redact-arm.md:1565`) —
re-read whole. **One sentence is left false, and it is not in the cell
this WP's clause was appended to.** B3b's row has three pipe-delimited
cells (id / condition / action); the appended clause lives at the end of
the condition cell, per the spec's instruction to append "at the end of
the row's first content cell". The **action** cell — untouched, and
correctly so — states in its own voice:

> **The error message is the only surface that reaches the user on an
> abort, so it is the one that must carry the basename**

Under this WP's pair rule, no reachable abort may pair a non-null
`redactedName` with an abort message at all: `secretGateAbortMessage`
throws a contract-violation error on any such call, and every reachable
abort's `redactedName` is `null` by construction (Table P row P3's note —
a surviving `redacted/` copy always recovers, so no reachable abort has
one). Read as an operative instruction, this sentence would have an
implementer construct exactly the contract violation the pair rule
forbids. The appended condition-cell clause does not resolve this — it
scopes only the trigger class and the two conjuncts, and says nothing
about the message's basename field.

This is a **real drift**, and it is **out of this WP's boundary to fix**:
the spec's Out of scope list forbids "any change ... to any row of either
`Done` spec other than G5, Q18 and B3b" *by more than the one appended
clause*, Implementation notes pins B3b's edit to exactly that one clause
at the end of the condition cell, and the acceptance criterion requires
`git diff` to show exactly one changed line per row (confirmed: the three
hunks are `:465`, `:1565`, `:1647`, one line each). The action cell is a
different line I am not authorized to touch. Routed to wd-architect —
see Discovered issues.

## Decisions made

- **`quarantinePreserve` and `secretGateAbortMessage` are now exported.**
  Neither was exported before. Table D's disposal states (D0–D4, especially
  D1's "collision candidate survives" and D3's "cleanup failure fails
  loud") and the pair rule are precise, low-level filesystem/contract
  behaviours that are far more directly and non-vacuously testable against
  the primitives themselves than indirectly through a full gate fixture —
  simpler than building elaborate gate-level scaffolding to reach the same
  states. Both exports are documented in `module.exports` with a one-line
  rationale. No other module imports them; `src/` behaviour is unchanged.
- **The withhold-arm trigger is `!preserved && !redactCopy`, not a literal
  "would the record be empty" recomputation.** I verified by exhaustive
  case analysis (documented inline in the diff) that this is exactly
  equivalent to "the preservation record about to be returned is empty" —
  simpler than restructuring the function to build the record first and
  check its length, which would have meant computing the identity-gated
  dedup logic before deciding whether to abort.
- **The abort's `identity` field is now always the literal `'not
  performed, because there was no saved copy to compare against'`,** with
  no runtime branch computing it. Table P's own text ("on every arm this WP
  makes reachable there is no `redacted/` copy... no basename is
  rendered") establishes that the other two identity dispositions are
  unreachable on every abort this WP can raise (P3's escape short-circuits
  before the abort is ever considered when a `redacted/` copy survives) —
  simpler than keeping the old `Buffer.compare`-based branch as dead code.
- **P1/P2's two causes get one shared trigger test shape** (untracked-only
  for the two P2 causes; tracked+untracked for P1, mirroring the existing
  FI-12/13/14 pattern). The tracked/untracked distinction was measured to
  be inert under the new contract (both paths reach the same
  `assertAbort` checks, since the retired revert/checkout machinery no
  longer exists) — chosen to keep the added surface proportionate rather
  than duplicating every combination.
- **The two report-fallback / accounting test gate-stubs across
  `dream-promote.test.js` that used to return `{refuse:true}` with no or an
  empty `preserved` field** (none of them testing P4) were given a minimal
  non-empty record rather than rewritten to test something else — the
  simpler fix that preserves each test's original subject.
- **Comment updates near the P4 check in `promote.js`** were kept to one
  short paragraph explaining the new symmetry with the redact arm's
  existing check, rather than restating Table P.
- **The two retired `assert.ok(!/quarantine copy failed/…)` negative
  guards** (main `tests/unit/dream-validate.test.js:1232` and `:1247`,
  guarding row G7's retirement of that reason suffix) vanished with the two
  rewritten P1 fixtures — those fixtures now abort and carry no
  `entry.reason` to guard at all. The literal `quarantine copy failed` is
  confirmed absent from both `src/` and `tests/` (`git grep`), and it was
  already absent from `src/` on `main`, so nothing is currently unguarded.
  Deliberate retirement, not a silent drop.
- **`quarantinePreserve`'s temp-file ownership tightened across two review
  rounds, both fixes converged on independently by multiple reviewers.**
  (1) Round 1: the herdr-shadow hermetic Codex reviewer found (band A) that
  the temp path `.tmp-<pid>-<stem>` is deterministic and pids are reused,
  so a crash-leftover foreign file at that name would be silently
  overwritten and then deleted as "owned" on a later failure — fixed with
  an exclusive create (`flag: 'wx'`) in `40225aff`; the round-2 codex
  plugin independently reported the identical class of bug against the
  pre-fix tip. (2) Round 3: wd-reviewer found that fix itself was
  incomplete — `O_CREAT|O_EXCL` allocates the directory entry INSIDE the
  write call, so `tmpOwned` was being set only after the whole call
  returned, leaving a partial secret-bearing temp file undetected (and
  un-removed) on any failure AFTER the create (ENOSPC, a filesize limit,
  an I/O error); reproduced live with `ulimit -f 1` and `SIGXFSZ` ignored.
  The round-3 codex plugin converged on the same finding in the same pass
  and additionally caught a related but distinct bug in
  `secretGateAbortMessage`'s enum lookup (prototype pollution via
  `'toString'`/`'__proto__'`). Both fixed in `a7f24032`, with call-through
  RED tests that let the real syscall run before injecting the failure,
  rather than mocking the failure mode away.

- `neutralise()` is duplicated in `src/cli/dream.js` and
  `src/core/dream/promote.js` with no single owner — one security contract,
  two code carriers. Named in the spec's own Out-of-scope list
  (previously routed to the architect, not this WP's subject); not touched.
- **B3b's action cell drifted out of agreement with this WP's pair rule**
  (`docs/specs/done/WP-secret-fence-ep2-redact-arm.md:1565`, action cell):
  *"The error message is the only surface that reaches the user on an
  abort, so it is the one that must carry the basename"* now describes a
  forbidden pairing under `secretGateAbortMessage`'s contract-violation
  guard — see the re-read report above. Out of this WP's boundary (the
  spec pins B3b's edit to one appended clause on the condition cell only,
  and the acceptance criterion requires exactly one changed line per row).
  **Routed to wd-architect**: the action cell needs its own amendment in a
  follow-up so Q4's cited condition owner stops instructing a pair-rule
  violation.

Restating the orchestrator's summary from dispatch (2026-09-02, against
`d8e4afb6`): the Dispatch precondition's one question was answered —
**fail-loud CONFIRMED** by owner ruling
(`docs/specs/logbook/2026-09-02-owner-rulings-stub-queue.md` on PR #201's
branch): rows P0–P3 make the whole run fail loud on the two newly guarded
arms, P0b on a read-back mismatch, Table D row D3 on an unremovable
rejected artifact — nothing becomes refuse-and-continue. 36 executable
claims were re-run and PASSED, 0 STALE, 1 UNRUNNABLE (V5's counterfactual
evidence, which needed the finished state — produced above). Confirmed
exact at dispatch time: `quarantinePreserve` at `:642`, call sites `:1017`
/ `:1040`, guard `:1042`, the `:648-675` try/catch, `:1054`'s
`Buffer.compare` alias bug, `:872`'s three-arg `secretGateAbortMessage`,
`promote.js:1195`/`:1221`'s asymmetry, `dream.js:940-963`'s arm-agnostic
catch, the two falsified test passages at `:1970-1978`/`:2040-2049`, the
G5/Q18/B3b row locations, zero `fsync`/`fdatasync` in `src/`+`tests/`, V1
green on main (2432/0/12), V2–V4 RED on main. Both measured behaviours
were reproduced live before this implementation began: the three-arm
withhold table and the P0b corrupt-artifact case. This implementation
closes every one of those gaps: P1/P2 now abort, P0b now verifies, Table D
now disposes correctly (including the D3 fail-loud case), and P4 now
checks the withhold arm too.

## Lessons

- WP-preservation-abort-widening: several pre-existing tests across
  `dream-validate.test.js` and `dream-promote.test.js` encoded the OLD
  (buggy) behaviour directly as their assertion — not as regression
  fixtures, but as the thing being verified (e.g. "a hard secret whose only
  preserve fails refuses normally with an empty record"). A spec that
  widens a trigger class should expect to find and rewrite these, not just
  add new coverage; grepping for the literal old return shapes
  (`preserved: []`, a bare `{refuse:true, reason}`) across the whole test
  file surfaced them faster than running the suite and fixing failures one
  at a time.
- WP-preservation-abort-widening: a behavioural change to a shared
  contract point (here, "a refuse verdict must carry a non-empty
  preservation record") can be correct in the file it's specified against
  while silently breaking unrelated test fixtures elsewhere in the same
  file that happen to exercise the same code path incidentally. Worth a
  full-file grep for the mutated shape before trusting a green run on the
  new tests alone.
- WP-preservation-abort-widening: assertions like `!message.includes('\n')`
  are vacuously true unless the fixture driving them actually contains a
  newline — a security-checklist requirement phrased as "the rendered
  message contains no raw X" needs at least one fixture that actually
  carries X, or the check never really ran.
- WP-preservation-abort-widening: embedding a literal control byte (e.g. a
  real ESC character) directly in a JS string literal via a tool call is
  fragile — it round-trips fine through `node --check` but produced a raw
  0x1B byte sitting in the source file. `String.fromCharCode(N)` is the
  robust way to construct hostile-byte fixtures in source.

Generated-by: claude-sonnet-5

---


thegyula:
## Round 1 review dispositions — PR #205 (tip `c7ff8f12`)

**Codex plugin**: CLEAN, no findings.

**wd-reviewer** (raw: `docs/specs/logbook/2026-09-02-preservation-abort-pr205-raw-round1-wd-reviewer.txt`):

1. **[B, FIX]** `validate.js:1046-1047` stale `@throws` JSDoc on the `secret` gate — fixed in `c7ff8f12`. Now states the class (Table P row P0, no VERIFIED artifact) plus a second `@throws` line for Table D row D3's cleanup-failure propagation, and drops the "durable" over-claim. Comment-only.
2. **[B, FIX + route]** Missing spec-mandated re-read report — added to the PR body under "## Amended-cell re-read report". G5 and Q18: no sentence left false. B3b: the reviewer's finding confirmed — the action cell's *"the error message ... is the one that must carry the basename"* is now a forbidden pairing under the pair rule, out of this WP's boundary to fix (spec pins B3b's edit to one appended clause on the condition cell only). Routed to wd-architect, recorded under Discovered issues in the PR body.
3. **[C, residual]** The two retired `!/quarantine copy failed/` negative guards — recorded under Decisions made as a deliberate retirement (the literal is confirmed absent from `src/` and `tests/`).
4. **[C, drop]** V4's violating-state wording — no change; the spec's own "synonym of the old cross-product claim" mutation was independently confirmed red by the reviewer.
5. **[C, drop]** V5 count differences (mutation-shape dependent, both directions red) — no action.

Re-verified on `c7ff8f12`: `npm test` rc=0 (2453 pass / 0 fail / 12 skipped — one transient `update-check` network-timing flake seen once, reproduced green in isolation and on a clean full re-run), `npm run lint` rc=0, `node scripts/boundary-check.js` rc=0.

thegyula:
## Round 1 addendum — hermetic Codex shadow (PR #205, new tip `40225aff`)

**Channel: herdr shadow** (raw: `docs/specs/logbook/2026-09-02-preservation-abort-pr205-raw-round1-herdr-shadow.txt`)

**1. [Band A, P1, FIX]** `validate.js:714` (now shifted) — `quarantinePreserve`'s temp path `.tmp-${pid}-${stem}` was opened with the default `'w'` flag. Since pids are reused, a crash-leftover file at that exact name would be silently overwritten on write, and then deleted as "invocation-owned" if a later step (chmod/rename) failed — destroying foreign, possibly secret-bearing bytes with no record, contradicting acceptance criterion D4 and Table D's "removes only what it created" rule.

Fixed in `40225aff`: `tmp` is now created with `flag: 'wx'` (exclusive). Ownership is tracked by a new `tmpOwned` flag, set only once that exclusive create succeeds — an `EEXIST` on a foreign file falls into the same D0/D1 failure path with `tmpOwned` still `false`, so cleanup never touches it. The pre-existing file is never opened, written, or removed; it's reported as an ordinary preservation failure.

Two new RED tests seed the tmp pathname with foreign bytes:
- `quarantinePreserve (Table D row D1, herdr-shadow round 1 P1): a FOREIGN file already at the tmp pathname (crash + reused pid) is never overwritten, and a later failure does not delete it` — combines the seed with an injected rename failure, reproducing the exact reported shape.
- `quarantinePreserve (Table D row D1, herdr-shadow round 1 P1): the EXCLUSIVE create alone fails the preservation over a foreign tmp, on an otherwise-happy path` — no other injected failure, isolating that `flag: 'wx'` alone is what fails the call.

Both verified RED (`fail 2`) against a reverted tree (`flag: 'wx'` removed) before this commit — the "happy path" test showed the primitive returning a successful result with the foreign bytes overwritten and renamed away — and GREEN against the fix, with the reverted tree restored byte-identical afterward (`diff` clean).

Also updated the JSDoc block above `quarantinePreserve` to state the exclusive-create ownership rule explicitly.

Re-verified on `40225aff`: `npm test` rc=0 (2455 pass / 0 fail / 12 skipped), `npm run lint` rc=0, `node scripts/boundary-check.js` rc=0.

thegyula:
## Round 3 dispositions — PR #205 (tip `a7f24032`)

**wd-reviewer** — Band A finding, a regression of the round-1 herdr-shadow fix (`40225aff`):

1. **[A, FIX]** `validate.js` — `tmpOwned = true` was set only after the whole `writeFileSync(..., {flag:'wx'})` call returned. Since `O_CREAT|O_EXCL` allocates the directory entry *inside* that call, any failure after the create (ENOSPC, a filesize limit, an I/O error) threw with `tmpOwned` still `false`, so cleanup skipped a partial secret-bearing temp file this invocation itself created — violating Table D row D1 and acceptance criterion D4. Fixed exactly per the reviewer's verified shape: the exclusive create now runs in its own try/catch, `tmpOwned` is set to `true` on any failure *except* `EEXIST` (the only outcome meaning the path is foreign), then rethrows into the existing D0/D1 cleanup. Kept the catch-based ownership tracking (not split into `openSync`+`writeFileSync(fd)`) per the reviewer's own note about the existing fixtures' `typeof p === 'string'` patching.

**Codex plugin** — converged with wd-reviewer on the same class (P2, same fix, same commit) plus one new finding:

2. **[C, FIX]** `secretGateAbortMessage` — the closed-enum lookup was a bare `messages[which]`. A plain object literal inherits `Object.prototype`, so `which` values like `'toString'`, `'__proto__'`, `'constructor'` resolved to an inherited function/object instead of `undefined`, passing the `=== undefined` guard and rendering a malformed message instead of failing loud. Fixed with `Object.hasOwn(messages, which)`.

Both fixed in `a7f24032`. New tests:
- `quarantinePreserve (Table D row D1, round-3 review): a failure AFTER the exclusive create still removes this invocation's OWN partial tmp file` — a call-through RED: the `writeFileSync` wrapper lets the real exclusive create run (so the directory entry genuinely exists on disk) before injecting `ENOSPC`. Verified RED against the pre-fix (`40225aff`) shape (`actual: true` where `false`/absent was expected — the partial file survived) and GREEN against the fix, tree restored byte-identical after the revert.
- `secretGateAbortMessage (round-3 review, codex plugin P3): an Object.prototype-inherited which fails loud, not a malformed message` — covers `toString`/`__proto__`/`constructor`/`hasOwnProperty`. Verified RED against the pre-fix bare-lookup shape (no throw for `'toString'`) and GREEN against the fix.

PR body updated: V1 re-pasted at the new counts (2469/2457/0/12), V5 extended with items 6 and 7 (the round-1 foreign-leftover pair and this round's call-through RED) with their red evidence, and the fix's origin (shadow r1 + plugin r2 converged on the overwrite bug; wd-reviewer r3 found the incomplete-ownership regression; plugin r3 converged on the same class and added the enum-pollution finding) recorded under Decisions made.

Re-verified on `a7f24032`: `npm test` rc=0 (2457 pass / 0 fail / 12 skipped), `npm run lint` rc=0, `node scripts/boundary-check.js` rc=0.

thegyula:
## Round 4 — closing confirmation (tip a7f24032; raws committed through the shadow's round-4 file)

| Channel | Verdict on a7f24032 | Notes |
|---|---|---|
| wd-reviewer | **APPROVE** | round-3 regression verified fixed at the OS level (genuine `ulimit -f` EFBIG → `dir=[]`; O_EXCL returns EEXIST before the permission check, so the carve-out holds); both new REDs proven red without the fix; `Object.hasOwn` guard checked against 13 prototype/odd keys |
| Codex plugin | no findings | a READING — its sandbox could not run the suite; the other two channels executed it on the same tip |
| hermetic shadow | no findings, "patch is correct" | re-ran the seeded crash-leftover and create-then-fail probes against the committed code; V2–V5 and the Done-spec numstat checks exit 0; porcelain identical |

Four rounds: shadow r1 (band A, temp-path ownership) → plugin r2 converged → fix → wd-reviewer r3 found the fix's own ownership-timing regression (shadow r3 converged independently) → fix → r4 clean on all three channels. Every finding is fixed or dispositioned; CI green. **Merge is the maintainer's.**

Carried forward for wd-architect (out of this WP's boundary): `docs/specs/done/WP-secret-fence-ep2-redact-arm.md:1565` B3b's action cell still instructs carrying the basename in the abort message, which the pair rule now forbids.

PR #4, opened 2026-07-02
feat(install): implement curl installer bootstrapper (WP-016)

thegyula:
Spec: docs/specs/WP-016-curl-installer-script.md

Stacked PR — base wp/003-installer-core (this branch also carries WP-001's CI/lint pipeline and WP-003's installer core, both approved but not yet merged to main).

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table
- [x] Spec status flipped to In-Review

## Verification output

```
$ bash -n install.sh
(exit 0, no output)

$ npm run lint
> wienerdog@0.0.0 lint
> node scripts/lint.js

--- markdownlint ---
markdownlint-cli2 v0.23.0 (markdownlint v0.41.0)
Finding: docs/**/*.md *.md
Linting: 35 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- frontmatter check ---
frontmatter check passed: 6 spec(s), 4 agent(s)

lint passed

$ npm test -- --test-name-pattern install-sh
> wienerdog@0.0.0 test
> node --test --test-name-pattern install-sh

✔ tests/unit/boundary-check.test.js (77.02425ms)
✔ tests/unit/init.test.js (74.501625ms)
✔ install-sh: missing/old Node exits 1 with nodejs.org guidance (281.635542ms)
✔ install-sh: recent Node hands off to npx wienerdog@latest init (641.483208ms)
✔ install-sh: script text contains a root-user check (0.503875ms)
✔ install-sh: sudo/apt/brew only appear in comments or echoed guidance, never invoked (0.795ms)
✔ tests/unit/manifest.test.js (78.191458ms)
✔ tests/unit/uninstall.test.js (73.779833ms)
ℹ tests 8
ℹ pass 8
ℹ fail 0
```

Full `npm test` (no filter) also passes: 30/30.

## Decisions made

- Root check uses `[ "$EUID" -eq 0 ]` (bash builtin, no external `id` process) rather than `id -u`, matching the spec's literal "EUID -eq 0" wording.
- Factored the "missing Node" and "Node too old" per-OS guidance into one `node_guidance()` function since the spec's required message text is identical for both cases — avoids duplicating the Darwin/Linux branches twice.
- Test case (c) (EUID 0 behavior) is not exercised by actually running as root — CI runners aren't reliably root-testable without sudo, and the spec explicitly allows substituting a grep-assert. Instead the test asserts the script text contains both `EUID` and `root` (case-insensitive), matching the spec's suggested fallback.
[...]

PR #215, opened 2026-09-04
fix(dream): deny dot-prefixed path segments as a class (WP-dot-segment-denial)

thegyula:
Spec: docs/specs/WP-dot-segment-denial.md

## Deliverables checklist

Ran by the orchestrator against `5fa23ea5` (this worktree's HEAD = `origin/main`) before dispatch; recorded here per the dispatch message. All items (a)-(j) reproduced as stated: `git diff --stat 29c61d03 HEAD -- src tests scripts package.json .github` empty; every `promote.js` anchor (`:99` `INSTRUCTION_BASENAMES`, `:112` `DENIED_SEGMENTS`, `:228-229` the "Deliberately NOT a dot-rule" comment, `:1045-1057` the row-Z2 comment) resolved; `isSafeRelativePath` byte-identical at `layout.js:65-71` and `layout-infer.js:40-46`; `adopt.js:372-382` the config write; `dream-promote.test.js:1923-1960` the report-fallback test; exports as stated; production `makeAdmit(defaultLayout())` behaviour as measured; spec `status: Ready` with `WP-instruction-basename-currency` Done; V1 exits 1 for the missing deliverable, V2 exits 1 with `ref1 1030/2000 | ref2 1030/2000 | ref3 463/1000 | over 12/12 | handoff 36/87 | boundary 16/39`; two RED-proof declaration files exist, the new test file does not.

**OWNER RULING for the Dispatch precondition**, recorded per the dispatch message: the orchestrator dispatches under the architect's recorded recommendation — the reader's per-key silent fallback stays SILENT after a dot-prefixed layout value is refused; no notice is added (row A7). Owner may reverse.

## Post-open fix: CI was red on both `test` legs (commit `3b5f5b5c`)

CI's first run of `a1bd795f` failed the `test` job on ubuntu and macos (Node 20): the new `adopt --yes` round-trip test (`tests/integration/adopt-e2e.test.js:563`) threw inside `adopt.run` at `src/cli/adopt.js:258` — `adoption needs a working claude on your PATH`. The test passed locally only because a real `claude` happened to be on the developer machine's PATH; CI runners carry none, and `adopt`'s A5 pin preflight verifies `claude` before its first mutation.

**Fix:** the new test now stubs `claude` on PATH exactly as this file's existing adopt-e2e test already does — the `FAKE_BRAIN` fixture installed at `<home>/.local/bin/claude` (chmod 0755), that directory prepended to `PATH` (WP-155's clean-PATH shape). No other line in the test changed. Commit `3b5f5b5c4a5ccecbe81272ec70f14dba4540c5ba`.

**Proof the stub carries it — `claude` genuinely absent from PATH, not just untried:**

```
$ mkdir -p /tmp/.../noclaude-bin
$ ln -sf $(dirname $(which node))/node /tmp/.../noclaude-bin/node
$ PATH="/tmp/.../noclaude-bin:/usr/bin:/bin" sh -c 'which claude; echo exit:$?'
claude-exit:1
$ PATH="/tmp/.../noclaude-bin:/usr/bin:/bin" node --test tests/integration/adopt-e2e.test.js
✔ adopt-e2e: init → adopt → sync → dream through mapped tiers, one revertable commit (2812.905292ms)
✔ adopt-e2e: adopting over an ALREADY-scheduled dream rebinds the per-job digest + catch-up map to the ADOPTED vault (no follow-up sync) — A7 hardening fix #3 (1733.93275ms)
✔ adopt-e2e: a FAILED existing-schedule rebind is SURFACED (not silent success) + entries still bind the adopted vault, and a subsequent sync-heal RETRIES the reload — A7 hardening 2 fix #2 (1925.661958ms)
✔ adopt-e2e: A5 pin preflight — claude unresolvable ⇒ adopt ABORTS before any mutation (vault/config/manifest/pin-store byte-identical) (19.717125ms)
✔ adopt-e2e: symlinked home — TCC guard refuses a Documents vault with zero writes (2.067333ms)
✔ adopt-e2e: adopt --yes round-trips a dot-prefixed tier candidate through denial (WP-dot-segment-denial, acceptance criterion 4) (1128.974333ms)
ℹ tests 6
ℹ pass 6
ℹ fail 0
EXIT: 0
```

`node` was reached only through a scratch directory containing a bare symlink to the real `node` binary (no `claude` alongside it); `git` resolved from `/usr/bin`. All 6 tests in the file pass, including the target test and the *other* pin-preflight test (`A5 pin preflight — claude unresolvable ⇒ adopt ABORTS`), which independently confirms the stripped PATH really has no working `claude` anywhere adopt would find it.

V1–V4 and boundary-check were re-run afterward (normal PATH) and all still pass — see below; the counts are unchanged from the original run since this fix touches no production code, only the test's own setup.

## Round 1 (gate findings, commit `0cbf1b48`)

`docs/specs/logbook/2026-09-05-dot-segment-pr215-raw-round1-*.txt` (wd-reviewer, Codex plugin, hermetic shadow), all against `a1bd795f`, adjudicated. Four findings; three FIX, one already closed:

- **wd-reviewer #1/#2 (band A) — PATH stub.** Already closed by `3b5f5b5c` before this round landed; matches the reviewer's own prescription. No further change.
- **plugin P1 (band B → macOS CI flake) — `tests/unit/dot-segment-denial.test.js`, B3's reference-equality grading.** The test compared `inferLayout`'s output against the *pre-filesystem* string this test drew, not the spelling the filesystem actually returns. On macOS (HFS+/APFS normalization) a composed accented character (é, ñ, å, ç) in a generated directory name can come back decomposed (NFD) from `readdirSync` — which is exactly what `pick()` (and hence `inferLayout`) sees — so the test's `impl` and `ref` would occasionally disagree on a correct implementation, seed-dependently. **Fix:** after `mkdirSync`, read the directory entry back via `fs.readdirSync(bvault, {withFileTypes:true})` and derive `expected` from *that* spelling, not from the drawn string — this is what Table F's B3 pre-step actually specifies ("each segment of the picked directory name after `pick`'s trim"), since the picked name always comes from `topLevelDirs`' own `fs.readdirSync`. See Decisions made for why NFC-normalizing both sides (the plugin's other suggested fix) was not the one taken.
- **shadow P1 (band A) — `tests/integration/adopt-e2e.test.js`, the round-trip test's vacuity.** If `adopt` omitted the `vault_layout:` block, or rendered a shape the capture regex didn't match, `persisted` would be `{}` and both comparison loops would pass vacuously; conjunct 3 alone doesn't discriminate either, since `.projects` sits outside the DEFAULT layout's tiers regardless of the new class rule. **Fix:** assert `persisted` carries all six `_dir` keys `renderLayoutBlock` always writes, pin `persisted.projects_dir` to the concrete value `'01-Projects'` (what this vault's fixture must fall back to), and keep the key-by-key conjunct-2 comparison. Confirmed discriminating: reverted both source edits in a scratch copy and re-ran the test — it now goes red on conjunct 1's concrete-value assertion (`persisted projects_dir must fall back to the built-in default, got ".projects"`), pasted below.
- **wd-reviewer #3 + shadow P3 (band C) — `src/core/dream/promote.js:1063`.** The rewritten row-Z2 comment cited `layout.js:65-71`, a range my own hunk had already moved (the function now spans `layout.js:67-74`). Fixed by citing the construct by name (`isSafeRelativePath` in `src/core/layout.js`), no line range — matching the spec's own Current-state citation policy.

Recorded, no action (per the coordinator): wd-reviewer #4 (commit prefix `feat` vs. the DoD's PR-title `fix` — both valid, no change) and #5 (under `WD_HELD_OUT_REPLAY=1`, T1 and T4 share one seed so their samples coincide — a replay is marginally weaker than acceptance; not a contract violation).

**Vacuity-check transcript (shadow P1's required proof):** scratch copy, both source edits reverted, ran the (already-fixed) test:

```
✖ adopt-e2e: adopt --yes round-trips a dot-prefixed tier candidate through denial (WP-dot-segment-denial, acceptance criterion 4)
  AssertionError [ERR_ASSERTION]: persisted projects_dir must fall back to the built-in default, got ".projects"
  + actual - expected
  + '.projects'
  - '01-Projects'
ℹ tests 6
ℹ pass 5
ℹ fail 1
```

All other findings' fixes, plus a full V1–V4/boundary-check/PATH-stripped re-run and both RED self-checks, are below (updated to reflect commit `0cbf1b48`).

## Round 2 (gate findings, commit `84d8da14`)

`docs/specs/logbook/2026-09-05-dot-segment-pr215-raw-round2-*.txt` (wd-reviewer, Codex plugin, hermetic shadow), all against `0cbf1b48`: wd-reviewer APPROVE, hermetic shadow "patch is correct", Codex plugin P2. Three small findings, all FIX, one commit:

- **plugin P2 (band B) — `tests/integration/adopt-e2e.test.js`, the round-trip oracle only captured six of the seven persisted layout keys.** `renderLayoutBlock` (`src/cli/adopt.js:91-103`) persists SEVEN values, `daily_filename` among them — itself a relative path (can be nested, e.g. `YYYY/MM/YYYY-MM-DD.md`) that `isSafeRelativePath` validates identically to the six `_dir` keys. The old `\w+_dir` regex silently skipped it: if `adopt` omitted or mis-rendered only `daily_filename`, presence, no-dot and read-back checks all still passed. **Fix:** parse the `vault_layout:` block the same way `readVaultLayout` itself does — find the block header, then consume every subsequently-indented `key: value` line — rather than a suffix-matching regex, so all seven keys are captured, presence-checked, and included in the no-dot-segment and key-by-key read-back comparisons. The `projects_dir === '01-Projects'` concrete pin is unchanged.
- **wd-reviewer (band C) — `tests/unit/dot-segment-denial.test.js`, `.find(...)` + `assert.ok` checked existence, not count.** A second (leaked or unexpectedly co-located) directory entry under `bvault` would have `.find()` silently pick one of them and grade a sample the loop never actually drew, rather than failing loud. **Fix:** `.filter((e) => e.isDirectory())` and `assert.equal(entries.length, 1, ...)`, keeping the `B2/B3 reference` band marker on the assertion. Re-ran the C2 RED self-check afterward to confirm the T4/T5 partition is unchanged (below).
- **wd-reviewer (band C) — `tests/integration/adopt-e2e.test.js`, a comment cited `src/cli/adopt.js:~245`, a line number.** Fixed by naming the construct instead ("the dry createPins block in adopt.run"), no line number — matching the same citation policy as round 1's fix #4.

Not mine to fix: the spec's own literal V2 script (`docs/specs/WP-dot-segment-denial.md:951`) still grades B3 against the pre-filesystem spelling — spec text, routed to wd-architect per the coordinator; this branch touches the spec only at the `status:` flip.

All fixes, plus a full V1–V4/boundary-check/PATH-stripped re-run, the non-vacuity check, and both RED self-checks, are below (updated to reflect commit `84d8da14`).

## Verification output

All commands run in the worktree at `/Users/gyulafeher/Documents/Claude_Projects/wienerdog-wt/dot-segment-denial`. Full transcripts (V1's complete 2618-line TAP output, V2's full acceptance run, V3's complete red-proofs REACH section) are in the first PR comment to keep this body under 30 KB.

### RED self-checks (manual, before V1-V4)

Scratch copies with each mutation from `tests/red-proofs/dot-segment-denial.proofs.json` applied by hand, per the spec's required self-check.

**C1 — revert the `promote.js` loop:** exactly T1, T2, T3 fail, each `ERR_ASSERTION`, each carrying its band marker (`B1 reference`, `B1 boundary`, `B1 handoff`). T4-T6 pass. `pass 3 / fail 3`.

**C2 — revert the `layout.js` clause:** exactly T4, T5 fail, each `ERR_ASSERTION`, each carrying its band marker (`B2/B3 reference`, `B2/B3 boundary`). T1-T3, T6 pass. `pass 4 / fail 2`.

Both mutations reverted afterward; scratch copies discarded.

### V1 — `test -f ... && npm test`

```
$ test -f tests/unit/dot-segment-denial.test.js && npm test
deliverable present
...
ℹ tests 2618
ℹ suites 0
ℹ pass 2606
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
EXIT: 0
```

Baseline (untouched tree) was 2611 tests / 2599 pass / 12 skipped. +7 tests (6 new top-level tests in `dot-segment-denial.test.js` + 1 new test in `adopt-e2e.test.js`), all passing, 0 failing. Exactly one existing test edited (`dream-promote.test.js:1923-1960`'s case list), as required.

### V2 — acceptance mode (`WD_HELD_OUT_SEED` unset)

```
seed 1788540180053  samples 1000 per enforcement point  (reproduce with WD_HELD_OUT_REPLAY=1 WD_HELD_OUT_SEED=1788540180053)
generator: 109 distinct first code points of 112, leading-dot share 0.512
ref1 2000/2000 | ref2 2000/2000 | ref3 1000/1000 | over 12/12 | handoff 87/87 | boundary 39/39
V2 OK
EXIT: 0
```

Matches the spec's "FULL FIX" row exactly. (Re-run post-round-1 with the B3 filesystem-spelling fix in place; same shape, unaffected — this V2 command is my own reproduction of the spec's literal fenced script, not a deliverable file. See Decisions made for a residual noted about the spec's own literal V2 script sharing the pre-fix B3 comparison.)

### V3 — `npm run red-proofs` (unfiltered)

`node_modules` symlink replaced with a real copy for the SNAPSHOT phase, run unfiltered, then the symlink restored (never committed).

```
5 declared proof(s), 5 selected
snapshot: 1267 entries under the declared domain (`.git/` and `node_modules/` excluded)

Criteria:

RUN: PROVEN
EXIT: 0
```

One `PROVEN` roll-up line for `WP-dot-segment-denial criterion 1` naming both `dot-segment-admit-reverted` and `dot-segment-layout-reverted`. Declared-proof count rose from 3 to 5, as criterion 5 requires.

### V4 — `npm run lint`

```
--- markdownlint ---
Linting: 635 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 267 spec(s), 4 agent(s)

lint passed
EXIT: 0
```

### `node scripts/boundary-check.js`

```
$ node scripts/boundary-check.js docs/specs/WP-dot-segment-denial.md docs/specs/WP-dot-segment-denial.md src/core/dream/promote.js src/core/layout.js src/core/layout-infer.js tests/unit/dot-segment-denial.test.js tests/unit/dream-promote.test.js tests/integration/adopt-e2e.test.js tests/red-proofs/dot-segment-denial.proofs.json
EXIT: 0
```

No output on success; every touched file is within the spec's Deliverables table plus the always-allowed set (the spec file's own `status:` flip).

### Round 1 re-verification (commit `0cbf1b48`), same shape as above, all still green

- **RED self-checks, re-run against the fixed test file:** C1 → exactly T1, T2, T3 fail (`pass 3 / fail 3`); C2 → exactly T4, T5 fail (`pass 4 / fail 2`). Unchanged partition.
- **V1** — `npm test`: `2618 tests, 2606 pass, 0 fail, 12 skipped`, exit 0.
- **V2** — acceptance mode: `ref1 2000/2000 | ref2 2000/2000 | ref3 1000/1000 | over 12/12 | handoff 87/87 | boundary 39/39`, `V2 OK`, exit 0.
- **V3** — `npm run red-proofs`, unfiltered (symlink workaround): `5 declared proof(s), 5 selected`, `RUN: PROVEN`, exit 0.
- **V4** — `npm run lint`: clean, exit 0.
- **`node scripts/boundary-check.js`**: exit 0.
- **PATH-stripped `adopt-e2e` run**, same construction as the round-0 fix: `claude` confirmed absent (`which claude` → exit 1 on the constructed PATH), then `node --test tests/integration/adopt-e2e.test.js` under that PATH: `tests 6, pass 6, fail 0`, exit 0.

### Round 2 re-verification (commit `84d8da14`), same shape as above, all still green

- **RED self-checks, re-run against the round-2-fixed test files:** C1 → exactly T1, T2, T3 fail (`pass 3 / fail 3`); C2 → exactly T4, T5 fail (`pass 4 / fail 2`). Unchanged partition, confirming the B3 directory-count fix and the band marker didn't move the grading.
- **Non-vacuity check, re-run:** both source edits reverted in a scratch copy, ran the round-2-fixed round-trip test: `AssertionError: persisted projects_dir must fall back to the built-in default, got ".projects"` — still red on conjunct 1 with all seven keys now captured.
- **V1** — `npm test`: `2618 tests, 2606 pass, 0 fail, 12 skipped`, exit 0.
- **V2** — acceptance mode (using the filesystem-spelling-corrected B3 comparison, see the round-1 residual note above): `ref1 2000/2000 | ref2 2000/2000 | ref3 1000/1000 | over 12/12 | handoff 87/87 | boundary 39/39`, `V2 OK`, exit 0.
- **V3** — `npm run red-proofs`, unfiltered (symlink workaround): `5 declared proof(s), 5 selected`, `RUN: PROVEN`, exit 0.
- **V4** — `npm run lint`: clean, exit 0.
- **`node scripts/boundary-check.js`**: exit 0.
- **PATH-stripped `adopt-e2e` run**: `claude` confirmed absent, `tests 6, pass 6, fail 0`, exit 0.

## Decisions made

- **Commit-message convention vs. spec's Definition-of-done PR title.** The dispatch message specified the exact commit-message text verbatim (`feat(dream): ...`), while the spec's "Definition of done" item 2 gives a PR title with a `fix(dream): ...` prefix. Followed the dispatch message's exact text for the git commit (an explicit, literal instruction), and used the spec's exact wording for this PR's title. Simpler option: don't reconcile the two conventions myself; surface the discrepancy here rather than silently picking one spelling for both.
- **`adopt --yes` round-trip test's conjunct-3 refusal reason.** The Exact contracts' round-trip conjunction (criterion 4, conjunct 3) says the promotion allowlist "refuses one under the dot-prefixed directory the vault actually contains" without naming a specific refusal reason. In the test's own fixture (a vault whose only project-tier candidate is `.projects/`), the persisted layout falls back to the default `01-Projects` tier, so the dot-prefixed directory is refused as "not under a writable vault tier directory" (clause (a), before the class-rule loop is ever reached) rather than with the dot-segment reason — both are correct refusals of the same directory, and the conjunction's wording doesn't require the dot-specific reason here (that reason is separately pinned by T2/T3/E1-E4 in the class-proof suite). Asserted only `refusal !== null`, the simpler option, rather than requiring a specific reason string this conjunct doesn't name.
- Otherwise implemented as specified; no other ambiguity encountered.
- **Round 1, plugin P1's B3 fix: filesystem re-read vs. NFC-normalize-both-sides.** The finding offered two valid fixes. Took the filesystem-re-read form (`fs.readdirSync` after `mkdirSync`, compare against that spelling) rather than NFC-normalizing both `impl`'s and `ref`'s inputs, because it is the one that keeps the reference predicate applied to *exactly* the string `inferLayout`/`pick()` actually sees — which is what Table F's B3 pre-step already specifies ("each segment of the picked directory name after `pick`'s trim") — rather than to a third, normalized string neither production code path ever constructs. Normalizing both sides would have been fewer lines but would have graded a value the implementation never receives.
- **Round 1, shadow P1's fix: a concrete pinned value vs. a generic "has all keys" check.** The finding asked for both "carries the projects key (and every key…)" and "at least one key asserted to a concrete non-dot value". Implemented both: `EXPECTED_DIR_KEYS` presence-checks all six `_dir` keys `renderLayoutBlock` always writes, and `persisted.projects_dir` is pinned to the literal `'01-Projects'` (the only key this fixture's vault — one dot-prefixed `.projects/` directory — actually exercises through the producer's fallback). Simpler option taken over also pinning the other five keys to their defaults: they're not exercised by this fixture (no `.identity`/`.skills`/etc. directories exist in it), so a concrete assertion on them would be asserting a fact this test's setup doesn't test. **Superseded in round 2**: plugin P2 found the six-`_dir`-key regex silently skipped `daily_filename`, so this key set widened to all seven and the capture mechanism changed from a regex to a block parser — see below. The `projects_dir` pin itself is unaffected.
- **Round 2, plugin P2's fix: widen the regex vs. parse the block like `readVaultLayout` does.** Considered widening the capture regex to `/^  (\w+):\s*(.*)$/gm` (match any two-space-indented key, not just `_dir`-suffixed ones), which would have been a one-line change. Rejected it: `config.yaml` can carry other two-space-indented `key: value` lines outside the `vault_layout:` block (the managed jobs block adopt writes, e.g. `  name: dream`), and a generic regex would silently capture those into `persisted` too — trading one silent gap (missing `daily_filename`) for another (extra, unrelated keys polluting the non-vacuity and no-dot-segment checks). Instead parsed the block the same way `readVaultLayout` itself does: locate the `vault_layout:` header line, then consume only the contiguous run of indented lines that follows it. This is the simpler-to-reason-about option even though it is a few more lines, because it can't drift from what the block boundary actually is.

1. **The now-constant `admits` branch in the edited report-fallback test.** `tests/unit/dream-promote.test.js`'s `report-fallback: a reports_dir with a trailing slash still produces a report` test computes `const admits = !dir.split('/').includes('.')`. With the case list now holding only `'reports/dreams/'` and `'reports//dreams'` (neither contains a bare `.` segment), this condition is always `true`. Named in the spec's own Implementation notes and Definition of done as a fact to report, not refactor — the branch and its surrounding comment are left exactly as they are.
2. **The `.`-segment branch of row Z1's filter loses its only exercise.** `promote.js`'s `reportRel` derivation still has `.filter((seg) => seg !== '' && seg !== '.')`, which is correct and untouched (it operates on an in-memory `layout` object that never passes through `readVaultLayout`), but after this change no production caller can construct a layout value containing a `.` segment through config — so the `.`-branch of that filter is no longer reachable by any test. Named in the spec's Implementation notes as a residual owned by `docs/specs/done/WP-dream-promote-report.md` (row Z1's owner), not this work package.
3. **`docs/adr/0010-vault-adoption-paths.md` says `adopt` "requires the user to confirm it before writing config"**, which `--yes` has not done since it existed. Named explicitly in the spec's Out-of-scope section as something to report but not fix here.
4. **The spec's own literal V2 verification script (the "Verification steps" fenced command) carries the same B3 pre-filesystem-spelling comparison that plugin finding P1 found in the deliverable test file.** It is spec text, not a deliverable — untouched, only the `status:` flip is in scope. When reporting V2's output above, ran the corrected form (matching `dot-segment-denial.test.js`'s fix) rather than the spec's literal, theoretically-flaky-on-macOS text; both are reference-equality checks over the same predicate and differ only in whether B3's `expected` is read back from the filesystem, so this changes no result, only removes a latent flake from my own report. Flagging so a future spec-maintenance pass can apply the same one-line fix to the fenced script.

## Lessons

- WP-dot-segment-denial: the spec's V2 node -e script is a complete, self-contained reference implementation of Table F's grading (generator, distribution check, seed/N guards) — writing the test file's T1/T4 was mostly a matter of splitting that script's logic across two `test()` bodies with band-marker messages, not designing new grading logic from scratch. Specs that inline a runnable oracle this literally save real implementation time.
- WP-dot-segment-denial: mixed-encoding Unicode literals (precomposed vs. NFD-decomposed accented characters) in a spec's prose are easy to transcribe inconsistently into source by eye. Worth a codepoint-level `node -e` sanity pass over every non-ASCII string literal in the finished test file before trusting it — in this case the transcription happened to land correctly, but there was no way to know that without checking codepoints directly.
- WP-dot-segment-denial: for a RED-proof `find`/`replace` pair with an interpolated template literal containing escaped backticks (`` `not admitted: path segment \`${seg}\` begins with a dot` ``), constructing the JSON via a small Python script (rather than hand-escaping the string inline) made it easy to verify `content.count(find) == 1` before writing the file, and avoided a subtle escaping bug that would only surface when `red-proofs.js` tried to apply the mutation.
- WP-dot-segment-denial: a new integration test that spawns/verifies external tools (here, `adopt`'s pin preflight resolving `claude`) must build its ambient-tool assumptions on the same file's *existing* fixture, not on the developer machine's PATH — a real `claude` on my own PATH masked the gap through every local run, and it only showed up as a CI failure. Verifying "no real tool leaks through" needs an actual PATH-stripped run, not just a passing local one: `which <tool>` on the constructed PATH is cheap insurance before trusting a stub. Also noted: on this machine, both `/Users/gyulafeher/.local/bin/claude` and its homebrew symlink `/opt/homebrew/bin/claude` had to be excluded — `node`'s own directory was not a safe PATH entry to reuse as-is.
- WP-dot-segment-denial: a randomized test that writes real filesystem entries and then checks a validator's output must compare against what the filesystem hands BACK (`readdirSync`), not the string used to create the entry — on a normalizing filesystem (macOS HFS+/APFS) those two can differ for any composed accented character, and the divergence is seed-dependent, so it will pass on most seeds and most CI runs and still be a real bug. This is the same shape of trap as comparing case-folded output against un-folded input, just one layer removed (the OS does the folding instead of the code under test).
- WP-dot-segment-denial: a "no false negative" oracle over a persisted/read-back-and-recompared pair (the adopt round-trip test's conjuncts) needs at least one CONCRETE expected value asserted somewhere in the chain, not just internal cross-consistency checks (persisted-vs-read-back, or "no dot segment") — internal consistency is satisfied by an empty result just as well as by a correct one. The fix that actually discriminates is cheap (one `assert.equal(x, '01-Projects')`) but easy to omit when every other assertion in the test already "looks like" it's checking something real.
- WP-dot-segment-denial: a capture regex built around a NAMING CONVENTION (`_dir` suffix) is a silent coverage gap waiting for the one key that doesn't follow the convention — here `daily_filename`, which is validated by the identical function as the six `_dir` keys but doesn't share their name shape. When a test needs "every key a producer writes", parsing the actual block boundary (as the production reader does) is more robust than pattern-matching on a naming habit that happens to hold for most of the keys.
- WP-dot-segment-denial: `.find(predicate)` plus `assert.ok(result)` checks "at least one exists", not "exactly one exists" — a cheap substitution for `.filter(predicate)` plus `assert.equal(length, 1)` when the test's whole argument depends on there being exactly one match (here, a single directory the loop itself created and is about to read back). The two read almost identically at a glance; only the second one fails loud on a leaked or duplicated entry instead of silently grading whichever one `.find()` happened to return first.

---
Generated-by: claude-sonnet-5


thegyula:
## Full verification transcripts

### V1 — `npm test`, targeted excerpts (full run is 2618 tests; here: the new/changed test blocks plus the final summary)
```

> wienerdog@0.13.0 test
> node tests/with-temp-root.js tests/run.js

error: your local changes would be overwritten by revert.
hint: commit your changes or stash them to proceed.
fatal: revert failed
✔ adopt-e2e: init → adopt → sync → dream through mapped tiers, one revertable commit (2425.130416ms)
✔ adopt-e2e: adopting over an ALREADY-scheduled dream rebinds the per-job digest + catch-up map to the ADOPTED vault (no follow-up sync) — A7 hardening fix #3 (1416.474125ms)
✔ adopt-e2e: a FAILED existing-schedule rebind is SURFACED (not silent success) + entries still bind the adopted vault, and a subsequent sync-heal RETRIES the reload — A7 hardening 2 fix #2 (1152.158333ms)
✔ adopt-e2e: A5 pin preflight — claude unresolvable ⇒ adopt ABORTS before any mutation (vault/config/manifest/pin-store byte-identical) (10.694667ms)
✔ adopt-e2e: symlinked home — TCC guard refuses a Documents vault with zero writes (5.179375ms)
✔ adopt-e2e: adopt --yes round-trips a dot-prefixed tier candidate through denial (WP-dot-segment-denial, acceptance criterion 4) (682.327125ms)
✔ Claude present, plain init: skills + hooks registered, NO memory (161.643375ms)
✔ Claude present, init --fresh-vault: everything incl. digest + managed block (722.179084ms)
✔ Codex present, plain init: skills + hooks under <CODEX_HOME>/skills, NO memory (172.726333ms)
✔ dream-integration: a set WIENERDOG_FAKE_TODAY has ZERO effect — dream derives the date from the injected clock (WP-155) (197.508875ms)
error: your local changes would be overwritten by revert.
hint: commit your changes or stash them to proceed.
fatal: revert failed
...
✔ doctor: post-open ENOENT at fstat on digest.md → B9 doubt warn, not silence, no target inspected (FIX 1) (478.280709ms)
✔ doctor: post-open ENOENT at read on CLAUDE.md → doubt warn, never the confident no-block message (FIX 1, D-E1/D-E2) (471.585833ms)
✔ doctor: post-open ENOENT at read on digest.md → B9 doubt warn, not silence, no target inspected (FIX 1) (469.777833ms)
✔ doctor: read-marker control — a normal target read IS observed by the shim (FIX 2 fixture sanity) (474.162625ms)
✔ doctor: over-cap CLAUDE.md fast path → B5 warn with st_size and ZERO content reads (FIX 2, A-H7 fast tier) (476.612ms)
✔ doctor: over-cap digest.md fast path → B9 warn and ZERO content reads of the digest (FIX 2) (477.753708ms)
dot-segment-denial B1: seed 1788540555579  samples 1000  (reproduce with WD_HELD_OUT_REPLAY=1 WD_HELD_OUT_SEED=1788540555579)
dot-segment-denial B2/B3: seed 1788540555593  samples 1000  (reproduce with WD_HELD_OUT_REPLAY=1 WD_HELD_OUT_SEED=1788540555593)
✔ dot-segment-denial B1: makeAdmit equals the reference predicate over a full-alphabet sample (13.926084ms)
✔ dot-segment-denial B1: the boundary matrix — the U+002E first-character rule at makeAdmit (0.161959ms)
✔ dot-segment-denial B1: the handoff oracle — Table D's 29 paths, hand-written (0.32325ms)
✔ dot-segment-denial B2/B3: readVaultLayout and inferLayout equal the reference predicate over a full-alphabet sample (837.145084ms)
✔ dot-segment-denial B2/B3: the boundary matrix — the U+002E first-character rule at both layout validators (13.8585ms)
✔ dot-segment-denial: no over-denial — dot-free paths and layout values stay admitted and honoured (1.158667ms)
✔ dream-brain: buildClaudeArgs composes the hermetic argv (WP-130), no model (1.780417ms)
✔ dream-brain: DREAM_PROMPT tier lines are ABSOLUTE and vault-prefixed (D-DREAM-CWD) (0.113333ms)
...
✔ dream-promote report-fallback: EVERY preserved copy is announced — a redaction's, a refused path's, and the report body's (1.867666ms)
✔ dream-promote report-fallback: the REPORT BODY's own preserved copies are announced on every arm of the union (12.603333ms)
✔ dream-promote report-fallback: no field of a preserved copy is recovered from a refusal reason (2.907ms)
✔ dream-promote report-fallback: the second write is refused after the first published — `promoted` with `accounting.published === false` (4.92ms)
✔ dream-promote report-fallback: a symlinked report target refuses the FALLBACK's write and delivers the record (1.467333ms)
✔ dream-promote report-fallback: a `reports_dir` with a trailing slash still produces a report (3.966667ms)
✔ dream-promote report-fallback: two fold-equal candidates collide — neither is the body, and every delta record keeps exactly one carrier (2.21875ms)
✔ dream-promote report-fallback: each consumer takes the path Table Z names, and `rel` carries it per arm (8.302958ms)
✔ dream-promote report-fallback: the body is matched by CANONICAL path identity, not by literal equality (3.52225ms)
✔ dream-promote report-fallback: an UNREADABLE report path fails closed, and the record still reaches the caller (1.418083ms)
✔ dream-promote: a run in which the brain writes nothing promotes nothing and changes no existing note (1.703375ms)
...
✔ vendor: an attended sync carries containment forward or refuses — no DATA-shaped A7 write moves it (129.5675ms)
ℹ tests 2618
ℹ suites 0
ℹ pass 2606
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 51921.183625
exit:0
```

### V2 — acceptance mode, full output
```
seed 1788540670373  samples 1000 per enforcement point  (reproduce with WD_HELD_OUT_REPLAY=1 WD_HELD_OUT_SEED=1788540670373)
generator: 109 distinct first code points of 112, leading-dot share 0.514
ref1 2000/2000 | ref2 2000/2000 | ref3 1000/1000 | over 12/12 | handoff 87/87 | boundary 39/39
V2 OK
exit:0
```

### V3 — `npm run red-proofs`, full output
```

> wienerdog@0.13.0 red-proofs

RED proofs — root /Users/gyulafeher/Documents/Claude_Projects/wienerdog-wt/dot-segment-denial
5 declared proof(s), 5 selected
snapshot: 1267 entries under the declared domain (`.git/` and `node_modules/` excluded)

Criteria:

RUN: PROVEN

```

thegyula:
## Round 1

Gate round 1 (wd-reviewer + Codex plugin + hermetic shadow, all against `a1bd795f`) adjudicated. Four findings, one already closed, three FIX — addressed in `0cbf1b48`:

1. **wd-reviewer #1/#2 (band A) — PATH stub.** Already closed by `3b5f5b5c`, before this round landed. No further change.
2. **plugin P1 (band B → macOS CI flake) — `tests/unit/dot-segment-denial.test.js` B3 grading compared against the pre-filesystem string, not the spelling `readdirSync`/`pick()` actually returns (macOS HFS+/APFS can hand back NFD for a composed accented character).** Fixed: read the directory entry back via `fs.readdirSync` and derive the expected value from that spelling.
3. **shadow P1 (band A) — the `adopt --yes` round-trip test was vacuous when the `vault_layout:` block was empty or unparsed; conjunct 3 didn't discriminate either.** Fixed: assert all six `_dir` keys are present, pin `projects_dir` to the concrete value `'01-Projects'`. Confirmed discriminating — reverting both source edits in a scratch copy now sends the test red on conjunct 1 (`got ".projects"` vs. expected `'01-Projects'`).
4. **wd-reviewer #3 + shadow P3 (band C) — the row-Z2 comment in `promote.js` cited a `layout.js` line range my own hunk had already moved.** Fixed: cite `isSafeRelativePath` by name, no line range.

Recorded, no action: wd-reviewer #4 (commit-prefix convention, both valid) and #5 (replay-mode seed sharing between T1/T4, not a contract violation).

Re-ran V1–V4, boundary-check, both RED self-checks (C1 → exactly T1–T3, C2 → exactly T4–T5), and the PATH-stripped `adopt-e2e` run — all still green. Full detail in the PR body's "Round 1" and "Verification output" sections.

thegyula:
## Round 2

Gate round 2 (wd-reviewer + Codex plugin + hermetic shadow, all against `0cbf1b48`) adjudicated: wd-reviewer APPROVE, hermetic shadow "patch is correct", Codex plugin one P2. Three small fixes, all in tests — addressed in `84d8da14`:

1. **plugin P2 (band B) — `tests/integration/adopt-e2e.test.js`: the round-trip oracle only captured 6 of 7 persisted layout keys.** `renderLayoutBlock` also writes `daily_filename`, itself a relative path validated identically to the six `_dir` keys, but the old `\w+_dir` regex silently skipped it. Fixed: parse the `vault_layout:` block the same way `readVaultLayout` does (find the header, consume the indented run that follows) so all seven keys are captured, presence-checked, and compared key-by-key against the read-back layout. `projects_dir === '01-Projects'` pin unchanged.
2. **wd-reviewer (band C) — `tests/unit/dot-segment-denial.test.js`: `.find(isDirectory)` + `assert.ok` checked existence, not count.** Fixed: `.filter(isDirectory)` + `assert.equal(length, 1, ...)`, band marker (`B2/B3 reference`) unchanged.
3. **wd-reviewer (band C) — a comment cited `src/cli/adopt.js:~245`.** Fixed: names the construct ("the dry createPins block in adopt.run"), no line number.

Not mine to fix: the spec's own literal V2 script still grades B3 against the pre-filesystem spelling — spec text, routed to wd-architect.

Re-ran V1–V4, boundary-check, both RED self-checks (C1 → exactly T1–T3, C2 → exactly T4–T5), the non-vacuity check (both source edits reverted → still red on conjunct 1), and the PATH-stripped `adopt-e2e` run — all green. Full detail in the PR body's "Round 2" and "Verification output" sections.

thegyula:
## Gate closed — round 3 clean (head `84d8da14`; raws committed pre-adjudication through `…round3-codex-plugin.txt`)

| Round | wd-reviewer | Codex plugin `review` | hermetic shadow (pr-rubric) |
|---|---|---|---|
| 1 (`a1bd795f`) | REQUEST-CHANGES: the new adopt round-trip test needs the fake-`claude` PATH stub [A]; row-Z2 comment cites `layout.js:65-71` positionally [C] | P1: randomized B3 grading compared against the pre-filesystem spelling (NFD on a normalizing volume) | P1/A: the round-trip test vacuous when no layout block is captured; C: the same positional citation |
| 2 (`0cbf1b48`) | **APPROVE**; one spec-side B routed to wd-architect (the spec's own V2 B3 loop still uses the pre-filesystem spelling); two optional C's | P2: `daily_filename`, the seventh persisted key, missing from the round-trip oracle | **patch is correct** |
| 3 (`84d8da14`) | — (the two C's applied) | **no findings** | — |

Orchestrator's closing checks on a copy of `84d8da14`: new suite 6/6; adopt-e2e with `claude` absent from PATH 6/6; C1 reddens exactly T1–T3, C2 exactly T4–T5; with both source edits reverted the round-trip test fails on conjunct 1 (`got ".projects"`). CI green on both Node 20 legs.

Dispatched under the architect's Dispatch-precondition recommendation (the reader's per-key silent fallback stays silent) — the owner may reverse by dated amendment.

Residuals carried forward to wd-architect: the spec's V2 B3 loop (grade against the `readdirSync` spelling, and state the B3 pre-step once with V2 and the test as registered mirrors); the adopt-e2e Deliverables cell's setup enumeration omits the PATH stub; ADR-0010's "adopt requires the user to confirm" sentence vs `--yes`.

PR #218, opened 2026-09-04
fix(dream): stop the ledger-derived banners naming a quarantine shelf (WP-quarantine-banner-location)

thegyula:
Spec: docs/specs/WP-quarantine-banner-location.md

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table (`git status --porcelain` == exactly the Deliverables rows; `node scripts/boundary-check.js` exit 0)

## Verification output

Full `npm test` and `npm run red-proofs` transcripts are in the first PR comment (this body would exceed ~30 KB otherwise). Condensed results below.

### V1 / V2 — four states each, extracted from the spec's fenced block and piped to `sh`

**1. deliverable absent** (`src/core/dream/warnings.js` moved aside):
```
V1 MISSING DELIVERABLE: src/core/dream/warnings.js
V2 MISSING DELIVERABLE: src/core/dream/warnings.js
V1/V2 RED
rc=1
```

**2. compliant** (this tree):
```
V1 OK
V2 OK
rc=0
```

**3. violating** (sentence retyped contiguously a second time, under `src/cli/`):
```
V1 OK
V2 SENTENCE HAS 2 AUTHOR(S) IN src/, expected 1
V1/V2 RED
rc=1
```

**4. composed-duplicate** (`warnings.js` imports the constant, never uses it, rebuilds the sentence from split literals — the state neither V1 nor V2 can catch, matching the spec's own measurement):
```
V1 OK
V2 OK
rc=0
```

### V5 — seven states, each a scratch git repo whose `main` holds the pristine `8302ce8e` copy of `docs/specs/done/WP-secret-fence-ep2-redact-arm.md`

| tree | V5 says | rc |
|---|---|---|
| untouched | 3 × `IS NOT ITS BASE ROW PLUS ITS CLAUSE` + `DIFF IS empty, expected 3/3` | 1 |
| each clause at the end of cell 4 (shipped) | `V5 OK` | 0 |
| marker-only (round 1's defeat) | 3 × `IS NOT ITS BASE ROW PLUS ITS CLAUSE` | 1 |
| each clause in cell 2 | 3 × `IS NOT ITS BASE ROW PLUS ITS CLAUSE` | 1 |
| compliant + one unrelated edited line | `DIFF IS 4/4, expected 3/3` | 1 |
| compliant + Q1 reauthored BEFORE its clause | `V5 ROW Q1 IS NOT ITS BASE ROW PLUS ITS CLAUSE` | 1 |
| compliant + Q1's middle cell reauthored | `V5 ROW Q1 IS NOT ITS BASE ROW PLUS ITS CLAUSE` | 1 |

All seven reproduced exactly as the spec's own table predicts.

### V3 — `npm run red-proofs` (real `node_modules`, so the lane runs)

```
11 declared proof(s), 11 selected
...
PROVEN       WP-quarantine-banner-location criterion 4 — doctor-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 1 — banner-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 2 — summary-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 3 — warnings-shelf-claim-restored=PROVEN

RUN: PROVEN
```
Five roll-up lines for this WP (criteria 1–5), criterion 5 naming both derivation proof ids, as the spec requires. (The other six PROVEN lines are baseline proofs from other, already-shipped WPs in this repo; unrelated to this package.)

### Row L7 — criterion 6, three-state failure-injection run (`tests/unit/dream-pipeline.test.js`, the new `[L7]` test)

The seam patches `fs.renameSync` for the destination ending in `transcript-ledger.json`, **delegates** the rename first (so the final ledger is genuinely durable) and then throws, on the refused-report arm (a symlinked report target). Ran on scratch copies of `src/cli/dream.js`, all three built from the pristine `8302ce8e` file plus the L0–L4 constant/import changes (so only the block's position varies):

| placement of the record print | result |
|---|---|
| shipped ordering (step 20, after `writeLedger`) | `tests 1 / pass 0 / fail 1` — announcing line and record lines are missing at the crash |
| after `writeLedger`, before the digest (round 2's rejected seam) | `tests 1 / pass 0 / fail 1` — same failure |
| step 17b (specified) | `tests 1 / pass 1 / fail 0` |

Only the step-17b placement passes, matching the spec's own three-state table.

### V4 — `npm test`, `npm run lint`

```
npm test  → tests 2623 / pass 2611 / fail 0 / skipped 12
npm run lint → 0 error(s) (637 markdown files), frontmatter check passed: 267 spec(s), 4 agent(s)
```

(Baseline before this WP: `tests 2618 / pass 2606 / fail 0 / skipped 12`. The delta is exactly the 5 new tests this WP adds: `[QBL-1]`, `[QBL-2]`, `[QBL-3]` in `ledger.test.js`, `[QBL-4]` in `doctor.test.js`, and the `[L7]` failure-injection test in `dream-pipeline.test.js`.)

### `node scripts/boundary-check.js`

```
$ git status --porcelain | awk '{print $2}' | xargs node scripts/boundary-check.js docs/specs/WP-quarantine-banner-location.md
exit: 0
```

### Re-read of the three amended `docs/specs/done/WP-secret-fence-ep2-redact-arm.md` cells, whole, per the spec's instruction

Re-read Q1, Q2 and Q9's `why` cells in full after appending each clause. No sentence is left false beyond what its own clause already scopes:
- **Q1**: the pre-existing sentence *"The parenthetical is byte-identical to Q2's, deliberately"* is exactly the sentence Q1's clause addresses (there is no second phrasing left to disagree with after Q2's retirement). No other sentence in the cell references the folder/lifecycle claim this package changes.
- **Q2**: the pre-existing *"same trap as Q1 … the prefix … is byte-identical on purpose"* sentence is exactly what Q2's clause retires (the stability the integration test depended on ends here, and that test is in this package's Deliverables).
- **Q9**: the pre-existing *"it is the one member of the withhold family this WP does not falsify … which stays true"* sentence is exactly what Q9's clause withdraws.

No further false sentence found.

## Decisions made

- **L7's "eleven-line" header vs. the ten literal lines quoted in the spec's fenced block.** The spec's fenced block (`Implementation notes`) contains exactly ten lines of comment text, but calls the result "these eleven lines". Resolved as the simpler reading: the pre-existing blank `//` separator line that already sat between the old two-line header and the `(i) THE REFUSED REPORT ARM.` paragraph is left untouched, immediately after the given ten-line block — giving eleven physical comment lines from `// 17b. ROW G11` through the blank line before `(i)`, and reproducing the block's own internal blank line (its line 3) plus that trailing one. This is the only reading under which the "byte-exact" ten-line block and the "eleven lines" claim are both literally true.
- **The two-line comment in `tests/unit/digest.test.js` (row L3/Deliverables) is prose, not spec-owned text** — the spec fixes only that the byte-identity claim must go, not replacement wording. Wrote a short comment explaining that the exhausted-transcript banner no longer carries a parenthetical to be identical to, and that the digest banner is now the only one naming `state/quarantine/`.
- **Table C's `find` strings were copied verbatim** rather than re-derived, per the spec's explicit instruction; each was confirmed (via the red-proofs run) to occur exactly once in its file before the corresponding source edit was written.

- **The `remediation: 'delete'` tension** (spec's Out of scope, quoted): `promote.js`'s `withRemediation(d.preserved, 'delete')` on the refusal arms gives every preserved copy `remediation: 'delete'`, which `REMEDIATION_GUIDANCE` renders as *"Nothing was promoted for this path; delete that copy."* This value is deliberate and owner-ruled (`WP-dream-promote-module` Table Q row Q9, 2026-08-29 round 3: `delete` is *"every refusing arm's value on both shelves — the note never entered the vault, the workspace is destroyed with the run, and what the copy holds is brain-authored text the product refused"*). What is unresolved is that the digest's own banner, in the same digest, says *"restore what you meant to keep"* about the same file, and Q9's reasoning does not consider the arm this package measures, where the copy is the note's ONLY one and sits on the bounded `redacted/` shelf. Not changed — it is a question about a ratified value and belongs to the owner and to Q9's own spec.
- **The legacy crash-window record** (Dispatch precondition's third owner item, accepted as a named residual, 2026-09-05 rulings): a `secret-revert-exhausted` record left by a pre-upgrade run that died inside the window row L7 closes renders `PRESERVED_COPIES_POINTER` for a record no run ever delivered. Registered at `WP-quarantine-preserve-durability` (Out of scope); not this package's.
- **Row L5's clearing sentence** (Dispatch precondition's second owner item, routed to the successor per the 2026-09-05 ruling): `src/core/digest.js`'s pending-review banner closes with *"this notice clears when no withheld copies are left"*, which is false in one measured state (`listSecretQuarantine` lists direct file entries only, so a fall-through only-copy under `redacted/` is not counted). `src/core/digest.js` stays outside this package's boundary per the ruling.

---
Generated-by: claude-sonnet-5 (implementer)


thegyula:
## Full verification transcripts

The PR body links here because the full transcripts exceed the ~30 KB body budget.

### V1 / V2 / V3 (`npm run red-proofs`) — extracted verification block, compliant tree
```
V1 OK
V2 OK
V5 OK

> wienerdog@0.13.0 red-proofs

RED proofs — root /Users/gyulafeher/Documents/Claude_Projects/wienerdog-wt/quarantine-banner-location
11 declared proof(s), 11 selected
snapshot: 1288 entries under the declared domain (`.git/` and `node_modules/` excluded)

PROVEN       doctor-shelf-claim-restored  (WP-quarantine-banner-location criterion 4)
             why: reverting doctor's interpolation to the shipped shelf sentence must redden the doctor line's own full-line pin; the four other Table A rows are untouched, so a pin that only checked row order would stay green here
             why: the derivation proof for the FOURTH carrier, which cannot join File A because its evidence is a CLI spawn and a declaration names one suite. Same mutation, different marker, different suite: the doctor line must move with the constant, or doctor has a second author
PROVEN       banner-shelf-claim-restored  (WP-quarantine-banner-location criterion 1)
             why: reverting the exhausted banner's use of the constant to the shipped shelf-and-delete literal must redden the banner's own full-string pin and nothing else in this suite
PROVEN       summary-shelf-claim-restored  (WP-quarantine-banner-location criterion 2)
             why: reverting the summary line's use of the constant to the shipped shelf sentence must redden the summary line's own full-string pin; the three counts are untouched, so a pin that only checked the integers would stay green here
PROVEN       warnings-shelf-claim-restored  (WP-quarantine-banner-location criterion 3)
             why: restoring warnings.js's own retyped shelf literal must redden the warnings document's pin — that carrier is the one this WP converts from a retyped copy into an import
             why: THE DERIVATION PROOF. Appending a marker inside the canonical constant's own literal must move EVERY carrier that derives from it, so all three identities in this suite go red together. A carrier that imported the constant and then composed a byte-identical copy of its own does not move, its identity stays green, and this proof cannot be PROVEN — which is the only observable difference between one author and two

Criteria:
PROVEN       WP-quarantine-banner-location criterion 4 — doctor-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 1 — banner-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 2 — summary-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 3 — warnings-shelf-claim-restored=PROVEN

RUN: PROVEN

```

### V4 — `npm test` summary + the five new tests this WP adds
```
ℹ tests 2623
ℹ suites 0
ℹ pass 2611
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 75530.851416

> wienerdog@0.13.0 lint
> node scripts/lint.js

--- markdownlint ---
Linting: 637 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 267 spec(s), 4 agent(s)

lint passed

New tests (all pass):
✔ doctor: [QBL-4] the secret-exhausted line renders the preserved-copies pointer verbatim (227.0595ms)
✔ dream-pipeline: [L7] the undelivered record is already in the output at the instant the ledger becomes durable (row G11, criterion 6) (383.207541ms)
✔ ledger: [QBL-1] the exhausted banner renders the preserved-copies pointer verbatim (0.052667ms)
✔ ledger: [QBL-2] the secret-revert summary line renders the preserved-copies pointer verbatim (0.034625ms)
✔ ledger: [QBL-3] the vault warnings document renders the preserved-copies pointer verbatim (0.177917ms)
```

### Row L7 — criterion 6, three-state failure-injection run (`tests/unit/dream-pipeline.test.js` [L7] test in isolation)
```
shipped ordering (step 20):
  ✖ dream-pipeline: [L7] the undelivered record is already in the output at the instant the ledger becomes durable (row G11, criterion 6)
  AssertionError: the announcing line was not delivered before the ledger became durable:
  actual: '', expected: /the report could not be written to your vault/
  tests 1 / pass 0 / fail 1

after writeLedger, before the digest (round 2's rejected seam):
  ✖ dream-pipeline: [L7] the undelivered record is already in the output at the instant the ledger becomes durable (row G11, criterion 6)
  AssertionError: the announcing line was not delivered before the ledger became durable:
  actual: '', expected: /the report could not be written to your vault/
  tests 1 / pass 0 / fail 1

step 17b (specified, shipped in this PR):
  ✔ dream-pipeline: [L7] the undelivered record is already in the output at the instant the ledger becomes durable (row G11, criterion 6)
  tests 1 / pass 1 / fail 0
```

thegyula:
## PR gate — round 1 (orchestrator's adjudication)

All three channels ran on `0fc37e0e`; raws committed in the follow-up commit under `docs/specs/logbook/2026-09-05-quarantine-banner-pr218-raw-round1-*`.

- **wd-reviewer: APPROVE.** Boundary, V1/V2/V5, red-proofs (11/11, five roll-up lines, RUN: PROVEN), npm test 2623/2611/0/12, lint 0 errors, declaration files byte-identical to Table C, the L7 three-state evidence re-run (17b GREEN; shipped RED; after-writeLedger RED). Four Band C findings, all spec-prose slips, none blocking: (1) "eleven lines" over a ten-line fence — the implementer's Decision stands; (2) the orphaned `(ii)` sub-bullet at step 19 — correctly left untouched; (3) "steps 18-20 keep their ids" — step 20 no longer exists; (4) the "two-line comment" mirror is six lines as shipped, including inside the shipped Q1 clause. **Disposition: routed to wd-architect for dated errata in the Done spec at the done-flip** (ADR-0031 note: three of the four are the same family — prose restating a count the tables fix).
- **Codex plugin (native review): clean** — "no functional regression or actionable defect".
- **Hermetic shadow (pr-rubric): "patch is correct"**, zero findings; its sandbox could not run the full suite (mkdtemp EPERM, disclosed), covered by wd-reviewer's run.

Clean or C-only on every channel → the gate is closed; merging under the owner's session authorization.

PR #221, opened 2026-09-05
fix(dream): make a preserved quarantine artifact durable (WP-quarantine-preserve-durability)

thegyula:
Spec: docs/specs/WP-quarantine-preserve-durability.md

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table (`src/core/dream/validate.js`, `tests/unit/dream-validate.test.js`, `tests/red-proofs/quarantine-preserve-durability.proofs.json`, `docs/specs/done/WP-preservation-abort-widening.md` row P0b, `docs/specs/done/WP-secret-fence-ep2-redact-arm.md` row B3b) — verified with `node scripts/boundary-check.js` (below)

## Round-1 gate fixes (commit `6b6c5596`, on top of `ad8c62dc`)

Applied after round-1 adjudication (wd-reviewer REQUEST-CHANGES with no `src/` change requested; plugin one P2; shadow two C):

1. **[plugin P2] win32 portability of the new tests.** `flushPreservation` issues no flush on win32 (row F5), but QPD-1/2/3/4/5/7 unconditionally inject or observe `fsyncSync` calls (and QPD-5 plants a symlink), so `npm test` would fail on a supported Windows checkout even though production behaves exactly as specified there. Fixed by skipping those six identities on win32 via `node:test`'s `skip` option (`{ skip: process.platform === 'win32' && 'row F5: no flush on win32' }`), naming the reason. QPD-6 is left unskipped — its commit/ownership mechanics (`linkSync`, `ownsName`, the exclusive create) are platform-neutral.
2. **[shadow C] flush-coverage comment overclaimed.** Qualified the flush-then-verify block's comment (not one of the spec's byte-exact source forms) from "the comparison that follows is against bytes a completed flush has already covered" to "the read occurs after a completed flush OF THAT INODE — not a claim that the flush covered the returned bytes, which coincides only absent a concurrent writer of that inode (row F10 (v))."

## Round-2 gate fix (commit `85514b9a`, on top of `6b6c5596`)

wd-reviewer APPROVEd round 1 in full (every disposition re-verified by execution, not re-reading); plugin and shadow each raised one P2, the SAME item, converged:

3. **[plugin P2 + shadow P2, same finding] The D3 temp test still carried two added `assert` statements** (`assert.ok(heldFd >= 0, …)` plus `assert.throws(() => fs.fstatSync(heldFd), /EBADF/, …)`) where the Deliverables cell and criterion 10 permit exactly one. **Folded both conditions into a single `assert.ok(...)`:**
   ```js
   assert.ok(
     heldFd >= 0 && (() => { try { fs.fstatSync(heldFd); return false; } catch (e) { return e && e.code === 'EBADF'; } })(),
     'the tmp create was observed AND the held descriptor is CLOSED even though the post-commit removal threw'
   );
   ```
   The `patchFs('openSync', …)` observer that captures `heldFd` stays — it is not an assertion. This is now genuinely ONE added assertion, with the non-vacuity condition folded inside it rather than split across two statements.
   - **(a) `post-commit-descriptor-leaked` still reddens exactly this test.** `npm run red-proofs`: `37 declared proof(s), 37 selected`; `post-commit-descriptor-leaked` → `PROVEN` (criterion 7); `RUN: PROVEN`. The message text still contains the declaration's `signal` ("the held descriptor is CLOSED"), which the folded assertion's single message preserves verbatim.
   - **(b) Non-vacuity, proven by hand.** On a scratch copy, I removed the finalizer by hand (`} finally { closeQuietly(fd); fd = -1; }` → `} finally { /* removed by hand for vacuity check */ }`) and re-ran the test in isolation. It fails:
     ```
     ✖ quarantinePreserve (Table D row D3): a tmp that cannot be removed after the commit fails LOUD, and the held descriptor is CLOSED
       AssertionError [ERR_ASSERTION]: the tmp create was observed AND the held descriptor is CLOSED even though the post-commit removal threw
           at TestContext.<anonymous> (tests/unit/dream-validate.test.js:2619:10)
       actual: false
       expected: true
     ```
     Confirms the folded assertion is not vacuous: with the finalizer removed, `heldFd` is still `>= 0` (a real descriptor was observed) but `fs.fstatSync(heldFd)` no longer throws (the leak), so the `&&` is `false` and the assertion fails — exactly the failure mode `assert.throws(…, /EBADF/)` alone would have missed only if `heldFd` had stayed `-1`, and exactly the failure mode `assert.ok(heldFd >= 0)` alone would have missed entirely.

4. **[plugin P2, other] "Avoid skipping identities selected by red-proofs" on win32 — NO CHANGE, per wd-reviewer's round-2 measurement.** `scripts/red-proofs.js` refuses at LOAD with `UNSUPPORTED` on win32 (its own chmod-enforcement probe, part of its sandboxing REACH disclosure), so the lane never runs there at all — no declaration is ever selected, and a skipped identity cannot produce a false `PROVEN` because nothing runs to produce any verdict. CI's `test` matrix is `ubuntu-latest, macos-latest` only (`red-proofs` is not a CI job at all), so this is not reachable in practice either. The skip guards from round 1 stand unchanged. (Decision recorded under "Decisions made".)

### boundary-check

```
$ git status --short | awk '{print $2}' | xargs node scripts/boundary-check.js docs/specs/WP-quarantine-preserve-durability.md
(exit 0, no output — every changed path is in the Deliverables table)
```

### V1 / V2 — extracted verification block, on the real candidate tree (re-run after the round-2 commit)

```
V1 OK
V2 OK
```

### V1 / V2 — the full six-state matrix (Definition of done item 1), run by me in a scratch git repo whose `main` holds the `bb58e398` base of both Done specs, candidate files copied in from this branch

| tree | my output | rc |
|---|---|---|
| deliverable absent (`src/core/dream/validate.js` removed) | `V1 MISSING DELIVERABLE: src/core/dream/validate.js or docs/specs/WP-quarantine-preserve-durability.md` (V2 still OK) | 1 |
| compliant (this candidate) | `V1 OK`, `V2 OK` | 0 |
| compliant, guarantee sentence HARD-WRAPPED over 4 JSDoc lines | `V1 OK`, `V2 OK` | 0 |
| P0b's clause moved into cell 2 instead of cell 5 | `V2 ROW P0b IS NOT ITS BASE ROW PLUS ITS CLAUSE IN CELL 5 — the clause is missing, truncated, in another cell, or something else on that line changed` | 1 |
| compliant + one unrelated edited line in the widening spec | `V2 WP-preservation-abort-widening.md DIFF IS 2/2, expected 1/1` | 1 |
| compliant + P0b's cell 5 REAUTHORED before its clause | `V2 ROW P0b IS NOT ITS BASE ROW PLUS ITS CLAUSE IN CELL 5 — the clause is missing, truncated, in another cell, or something else on that line changed` | 1 |
| guarantee sentence REWORDED away | `V1 GUARANTEE SENTENCE APPEARS 0 TIME(S) IN src/core/dream/validate.js, expected 1` + `V1 'on the medium' APPEARS 0 TIME(S) …, expected 1` | 1 |
| guarantee sentence RETYPED a second time | `V1 GUARANTEE SENTENCE APPEARS 2 TIME(S) …, expected 1` + `V1 'on the medium' APPEARS 2 TIME(S) …, expected 1` | 1 |
| sentence present, `fs.fsyncSync` removed | `V1 NO FLUSH IS ISSUED AT ALL in src/core/dream/validate.js` | 1 |
| the round-2 shadow's OVER-CLAIM fixture — approved sentence + one extra `A preserved copy is on the medium.` | `V1 'on the medium' APPEARS 2 TIME(S) IN src/core/dream/validate.js, expected 1 (inside the approved sentence)` | 1 |
| compliant + a DIFFERENTLY WORDED over-claim (`The preserved copy is safely on disk and will survive a crash.`) | `V1 OK`, `V2 OK` — the measured bound: V1 is a lexical guard against the one direct contradiction, not a semantic check | 0 |
| restored control (undo the last mutation) | `V1 OK`, `V2 OK` — confirms the mutations above, not tree damage, produced every RED | 0 |

Every row matches the spec's own predicted message; wd-reviewer independently re-ran the same twelve states in round 2 and reports identical results.

### No-over-claim obligation (Definition of done item 2)

Durability-related prose this diff adds to `src/core/dream/validate.js`, outside the one approved guarantee sentence (full diff in the commits): descriptions of what the FLUSH CALLS do and in what order (e.g. "the artifact's bytes and the directory chain that names it are flushed", "a flush of that inode had COMPLETED"; "durability to the extent the platform's flush provides" is not new prose — that phrase exists only once, inside the guarantee sentence itself), plus mechanism descriptions (exclusive create, no-clobber commit, three-valued ownership predicate) and, after the round-1 fix, the qualified flush-coverage comment. None of these states or implies that a preserved copy is safely "on disk", "on the medium" outside the one approved sentence, or will "survive a crash" — confirmed by the differently-worded-over-claim fixture above staying green (the measured bound) and by manual review of the diff. wd-reviewer's round-2 pass independently confirmed the qualified comment sits strictly outside all thirteen byte-exact source forms and narrows rather than adds a claim.

Both amended Done-spec cells were re-read whole after their clause: `WP-preservation-abort-widening` row P0b's original sentence ("This row establishes byte-identity, NOT durability — see... below Table D") is explicitly addressed by the new clause's opening sentence, which marks that pointer superseded as a forward-looking statement and standing as the historical record of `fc506110`. `WP-secret-fence-ep2-redact-arm` row B3b's 2026-09-02 clause ("has never been enforced — the product has no `fsync`") is explicitly quoted and marked as the historical record by the new clause. No sentence in either cell is left false and unaddressed.

### V3 — `npm run red-proofs` (real `node_modules`, re-run after the round-2 commit)

```
> wienerdog@0.13.0 red-proofs

RED proofs — root <worktree>
37 declared proof(s), 37 selected
snapshot: 1322 entries under the declared domain (`.git/` and `node_modules/` excluded)

PROVEN       preservation-flush-removed  (WP-quarantine-preserve-durability criterion 1)
PROVEN       artifact-flush-failure-swallowed  (WP-quarantine-preserve-durability criterion 2)
PROVEN       directory-flush-failure-swallowed  (WP-quarantine-preserve-durability criterion 2)
PROVEN       containing-directory-not-flushed  (WP-quarantine-preserve-durability criterion 3)
PROVEN       intermediate-shelf-not-flushed  (WP-quarantine-preserve-durability criterion 3)
PROVEN       state-directory-not-flushed  (WP-quarantine-preserve-durability criterion 4)
PROVEN       anchor-not-flushed  (WP-quarantine-preserve-durability criterion 4)
PROVEN       flush-order-inverted  (WP-quarantine-preserve-durability criterion 5)
PROVEN       destination-ownership-gate-removed  (WP-quarantine-preserve-durability criterion 6)
PROVEN       destination-removal-not-gated  (WP-quarantine-preserve-durability criterion 6)
PROVEN       ownership-check-follows-symlinks  (WP-quarantine-preserve-durability criterion 6)
PROVEN       commit-clobbers-destination  (WP-quarantine-preserve-durability criterion 7)
PROVEN       tmp-removal-dropped  (WP-quarantine-preserve-durability criterion 7)
PROVEN       tmp-removal-not-gated  (WP-quarantine-preserve-durability criterion 7)
PROVEN       d1-cleanup-not-gated  (WP-quarantine-preserve-durability criterion 7)
PROVEN       provenance-adopted-not-created  (WP-quarantine-preserve-durability criterion 7)
PROVEN       create-failure-removes-unowned  (WP-quarantine-preserve-durability criterion 7)
PROVEN       ownership-check-not-failure-closed  (WP-quarantine-preserve-durability criterion 6)
PROVEN       ownership-check-narrows-to-number  (WP-quarantine-preserve-durability criterion 6)
PROVEN       post-commit-descriptor-leaked  (WP-quarantine-preserve-durability criterion 7)
PROVEN       returned-bytes-rereads-the-artifact  (WP-quarantine-preserve-durability criterion 6)
PROVEN       read-before-flush  (WP-quarantine-preserve-durability criterion 1)
PROVEN       ownership-check-collapses-indeterminate  (WP-quarantine-preserve-durability criterion 6)
PROVEN       ownership-error-swallowed-at-last-gate  (WP-quarantine-preserve-durability criterion 6)
PROVEN       shared-catch-descriptor-leaked  (WP-quarantine-preserve-durability criterion 7)
PROVEN       directory-descriptor-leaked  (WP-quarantine-preserve-durability criterion 7)
             ... (plus the 11 pre-existing baseline proofs from other shipped WPs, all PROVEN, unaffected)

Criteria roll-up (this WP, 7 lines):
PROVEN       WP-quarantine-preserve-durability criterion 1 — preservation-flush-removed=PROVEN; read-before-flush=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 2 — artifact-flush-failure-swallowed=PROVEN; directory-flush-failure-swallowed=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 3 — containing-directory-not-flushed=PROVEN; intermediate-shelf-not-flushed=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 4 — state-directory-not-flushed=PROVEN; anchor-not-flushed=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 5 — flush-order-inverted=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 6 — destination-ownership-gate-removed=PROVEN; destination-removal-not-gated=PROVEN; ownership-check-follows-symlinks=PROVEN; ownership-check-not-failure-closed=PROVEN; ownership-check-narrows-to-number=PROVEN; returned-bytes-rereads-the-artifact=PROVEN; ownership-check-collapses-indeterminate=PROVEN; ownership-error-swallowed-at-last-gate=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 7 — commit-clobbers-destination=PROVEN; tmp-removal-dropped=PROVEN; tmp-removal-not-gated=PROVEN; d1-cleanup-not-gated=PROVEN; provenance-adopted-not-created=PROVEN; create-failure-removes-unowned=PROVEN; post-commit-descriptor-leaked=PROVEN; shared-catch-descriptor-leaked=PROVEN; directory-descriptor-leaked=PROVEN

RUN: PROVEN
```
Exit 0. Full raw transcript (including the REACH disclosure block and all pre-existing baseline proofs) posted in the first PR comment (round 0) and reproduced identically (modulo the snapshot entry count, which grows as gate-raw logbook files accumulate) after both round-1 and round-2 commits.

### V4 — `npm test` (re-run after the round-2 commit)

```
ℹ tests 2630
ℹ suites 0
ℹ pass 2618
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
```
Exit 0 — unchanged: the round-2 fix only reshapes two assertions into one inside an already-passing test, and the six win32-skip-guarded identities have `skip: false` on this (darwin) run, so the count and pass/fail split are identical throughout all three commits.

### `npm run lint` (re-run after the round-2 commit)

```
--- markdownlint ---
Linting: 640 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 268 spec(s), 4 agent(s)

lint passed
```
Exit 0.

### Occurrence audit (re-run after the round-2 commit)

All 13 byte-exact source-form fences and all 22 unique Table C `find` strings still match `src/core/dream/validate.js` exactly once each (`src/core/dream/validate.js` is untouched by the round-2 commit — only `tests/unit/dream-validate.test.js` changed).

## Decisions made

- **[round-2, corrected accounting — supersedes the round-1 bullet below] The D3 temp test now carries exactly ONE added assertion: the closure check with its non-vacuity guard folded in** (`assert.ok(heldFd >= 0 && (() => { try { fs.fstatSync(heldFd); return false; } catch (e) { return e && e.code === 'EBADF'; } })(), …)`), plus the unchanged `patchFs('openSync', …)` observer, which is not an assertion. This matches the Deliverables cell ("(3) ONE shipped test GAINS one assertion") and criterion 10 literally, while remaining non-vacuous — proven by hand (see "Round-2 gate fix" above): removing the finalizer that closes the descriptor makes this exact assertion fail.
- ~~[round-1, superseded] The D3 temp test gained TWO `assert` statements … routed to wd-architect as a spec defect (Deliverables cell vs Table C row QPD-7).~~ Two consecutive review rounds (wd-reviewer round 1 and round 2, independently; plugin and shadow round 2, converging on the same item) held that the two-assertion form was correct *as a spec matter* — the spec asked for a non-vacuity guard in one row (QPD-7) and forbade a second assertion in another (this cell) — but round 2 also surfaces a form that resolves it without violating either: folding the guard and the check into one Boolean expression under one assertion satisfies the "one added assertion" boundary literally and is still non-vacuous. Applied; the underlying spec inconsistency (a cell that can be satisfied by rewording rather than only by amendment) is narrower than originally escalated, but the Deliverables-cell/criterion-10 wording itself is still imprecise (it does not anticipate a folded form) and the census-defect item below remains routed to wd-architect.
- **[round-2] Plugin's win32 red-proofs P2 — no change**, per wd-reviewer's round-2 measurement that `scripts/red-proofs.js` refuses at LOAD with `UNSUPPORTED` on win32 (its own chmod-enforcement probe), so no declaration is ever selected there and a skipped identity cannot produce a false `PROVEN`; CI's `test` matrix is `ubuntu-latest, macos-latest` and does not run `red-proofs` at all. The round-1 skip guards stand unchanged.
- **Two of the three shipped `renameSync` test injections (the "FOREIGN file already at the tmp pathname" test and the D3 "tmp that cannot be removed" test) were left completely unmigrated**, per the spec's own Current-state finding that they don't break: the first never reaches its injection (the exclusive create refuses first); the second still gets its `WienerdogError` for a different, equivalent reason (the commit now succeeds via `linkSync`, and the post-commit gated temp removal hits the same `rmSync` injection).
- **Test-file helper design (not spec-fixed).** `patchDestRead` matches the descriptor read-back by comparing `fstatSync(fd)` against `statSync(dest)` (by `(dev,ino)`), since the shipped pathname-matched injection (`readFileSync(dest)`) no longer has a pathname to match once the read moved onto the held descriptor. The QPD-2/QPD-6 "distinct-site matrix" and "provenance" checks needed new injection shapes (openSync/writeSync fd-tracking, and a direct check of the create's `O_CREAT|O_EXCL` flags for `provenance-adopted-not-created`) that Table C names as proofs but leaves the test mechanics to the implementer, per "Test design, fixture shapes and the mechanics of each RED are the implementer's beyond what Table C fixes."
- **Every assertion inside each QPD-1..QPD-7 test identity carries that identity's `[QPD-N]` band marker in its message**, per Table C's requirement that `evaluateRed`'s signal match be satisfiable regardless of which specific assertion a mutation reddens.
- **[round-1] The six V1/V2 defeat states are now reproduced with my own measured output** (table above), superseding the prior decision to decline them.

- **[routed to wd-architect] `docs/specs/done/WP-preservation-abort-widening.md` row P0b's amended clause mirrors Table F row F3's chain membership** ("the containing directory, the shelf above it on the redacted arm, the state directory, and the core directory that names it") inside a `Done` spec, without that cell being registered in the Mirrored Surface Checklist. F3's membership now lives in four places (Table F, `quarantineDirChain`, the QPD-3/QPD-4 ordered `deepEqual`s, and this P0b cell) and only the first three are registered. Route: add an F3 chain-membership bullet to the Mirrored Surface Checklist, or trim P0b's clause to a citation of Table F. The clause is byte-exact per the spec and I had zero discretion over its wording; not actioned.
- **[routed to wd-architect] A third shipped test title is now factually stale under row F9, and the spec permits only two title changes.** `tests/unit/dream-validate.test.js`: `quarantinePreserve (Table D row D1): the write fails before any rename — tmp is removed, a \`dest\` COLLISION CANDIDATE survives byte-unchanged` — under F9 there is no rename in this path at all; the migrated injection (`patchFs('openSync')` on the `.tmp-` basename) fails the exclusive **create**, not "the write". The Deliverables cell's census of shipped-test changes found and permits exactly two title changes (the D1 commit test and the D3 temp test); this third stale title was missed by that census and the boundary forbids fixing it as an unlisted third change. Left untouched, correctly, per the permission boundary. Coverage is unaffected: the underlying contract this title describes still passes.
- **[routed to wd-architect] `src/core/dream/validate.js`: `DIR_OPEN_FLAGS`'s `O_DIRECTORY` is OR'd in unguarded, while `O_NOFOLLOW` on the same line (and on `TEMP_CREATE_FLAGS`) uses the repo's explicit `typeof fs.constants.X === 'number' ? … : 0` idiom.** `DIR_OPEN_FLAGS`'s own definition is not inside any of Table C's byte-exact source forms (only `flushDir`'s body, which references the constant by name, is byte-exact), so this is a hygiene inconsistency in code I authored — flagged here rather than fixed, per review instruction to route it to wd-architect rather than expand this PR's diff on a finding with no behavioural consequence (`O_DIRECTORY` is present on both CI platforms, and `flushDir` is unreachable on the one platform — win32 — that lacks the flag, since `DURABILITY_AVAILABLE` short-circuits first).
- **[routed to wd-architect, new in round 2] The Deliverables cell's prose census of permitted shipped-test changes has now mispredicted the diff in two independent directions across two review rounds** (the assertion-count item above — now resolved by folding, but only after two rounds — and the missed third stale title, still open). wd-reviewer's round-2 Contract-density check recommends extracting the census into a small canonical table (one row per permitted shipped-test change: seam migration, title change, assertion change) with criterion 10 asserting *that table* rather than a prose enumeration, so a future implementer's diff can be checked against rows rather than against prose that has already needed reconciling twice. Not actionable within this PR's boundary; routed to wd-architect.
- None beyond the above and what the spec's own "Out of scope" section already routes to `WP-quarantine-disposal-durability` (Draft) and `WP-ledger-delivery-stamp` (not filed, declined). No new functional gap was found during implementation or during either round of fixes.

---
Generated-by: claude-sonnet-5 (implementer)


thegyula:
Full raw transcripts referenced in the PR body.

## `npm run red-proofs` — full transcript

```

> wienerdog@0.13.0 red-proofs

RED proofs — root /Users/gyulafeher/Documents/Claude_Projects/wienerdog-wt/quarantine-preserve-durability
37 declared proof(s), 37 selected
snapshot: 1319 entries under the declared domain (`.git/` and `node_modules/` excluded)

PROVEN       doctor-shelf-claim-restored  (WP-quarantine-banner-location criterion 4)
             why: reverting doctor's interpolation to the shipped shelf sentence must redden the doctor line's own full-line pin; the four other Table A rows are untouched, so a pin that only checked row order would stay green here
             why: the derivation proof for the FOURTH carrier, which cannot join File A because its evidence is a CLI spawn and a declaration names one suite. Same mutation, different marker, different suite: the doctor line must move with the constant, or doctor has a second author
PROVEN       banner-shelf-claim-restored  (WP-quarantine-banner-location criterion 1)
             why: reverting the exhausted banner's use of the constant to the shipped shelf-and-delete literal must redden the banner's own full-string pin and nothing else in this suite
PROVEN       summary-shelf-claim-restored  (WP-quarantine-banner-location criterion 2)
             why: reverting the summary line's use of the constant to the shipped shelf sentence must redden the summary line's own full-string pin; the three counts are untouched, so a pin that only checked the integers would stay green here
PROVEN       warnings-shelf-claim-restored  (WP-quarantine-banner-location criterion 3)
             why: restoring warnings.js's own retyped shelf literal must redden the warnings document's pin — that carrier is the one this WP converts from a retyped copy into an import
             why: THE DERIVATION PROOF. Appending a marker inside the canonical constant's own literal must move EVERY carrier that derives from it, so all three identities in this suite go red together. A carrier that imported the constant and then composed a byte-identical copy of its own does not move, its identity stays green, and this proof cannot be PROVEN — which is the only observable difference between one author and two
PROVEN       preservation-flush-removed  (WP-quarantine-preserve-durability criterion 1)
             why: restoring a success path that reports a verified read-back with NO flush issued at all must redden every flush identity at once — QPD-5's flush-time substitutions included, since they are injected from inside a flush that no longer happens. QPD-6 stays green: the commit and the temp removal are not flushes, which is what makes ownership a contract of its own
PROVEN       artifact-flush-failure-swallowed  (WP-quarantine-preserve-durability criterion 2)
             why: swallowing the ARTIFACT flush's failure makes an unflushed copy report SUCCESS, the false statement Table F row F4 forbids. Only QPD-2 moves: the call is still made and the flushed sequence is unchanged
PROVEN       directory-flush-failure-swallowed  (WP-quarantine-preserve-durability criterion 2)
             why: the DIRECTORY branch has its own failure path — distinct open flags, distinct filesystem behaviour — and this is the ONE branch every directory site shares, which is why the site MATRIX lives in QPD-2 and the mutation lives here
PROVEN       containing-directory-not-flushed  (WP-quarantine-preserve-durability criterion 3)
             why: flushing the artifact bytes and not the directory entry that names them is the half-protocol that looks correct and is not: the bytes have had a completed flush and the entry naming them has not
PROVEN       intermediate-shelf-not-flushed  (WP-quarantine-preserve-durability criterion 3)
             why: THE TWO-LEVEL CASE, invisible on the withheld arm where qdir IS the shelf. QPD-4 moves, and QPD-2 with it — its redacted-arm shelf sites lose the flush they fail — while QPD-3 stays green, which is exactly the per-arm discrimination round 2 required
PROVEN       state-directory-not-flushed  (WP-quarantine-preserve-durability criterion 4)
             why: the middle of the chain, and the one entry always present on both arms: dropping stateDir leaves quarantine/'s own entry unflushed
PROVEN       anchor-not-flushed  (WP-quarantine-preserve-durability criterion 4)
             why: THE ROUND-1 ANCHOR FINDING. Dropping the anchor leaves state/'s own entry unflushed, and acquireLock creates state/ with mkdirSync(recursive) earlier in the SAME run
PROVEN       flush-order-inverted  (WP-quarantine-preserve-durability criterion 5)
             why: flushing the directory chain before the artifact leaves an interval in which the only COMPLETED flush covers an entry whose bytes have had none — row F6's rule stated as an order, never as a claim about what a crash or a device leaves. The flushed SET is identical, so only the two sequence identities move, and they see it as the FIRST flush no longer being the held inode
PROVEN       destination-ownership-gate-removed  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-1 AND ROUND-3 FINDING TOGETHER. Without the last gate the function reports success although `dest` no longer names the inode it verified and flushed — and because the read now goes THROUGH that inode's descriptor, the byte comparison can no longer mask it, so this gate is the only thing that can
PROVEN       destination-removal-not-gated  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-3 DATA-LOSS PATH. Ungating the failure-path removal makes this invocation delete whatever is at `dest` — including a file the user restored there during the run, which Table F row F10 puts squarely IN scope
PROVEN       ownership-check-follows-symlinks  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-2 SYMLINK FINDING, now in the shared predicate. A FOLLOWING stat matches through a symlink planted at the name, so a hard link of the held inode in a directory the chain never flushes would pass
PROVEN       commit-clobbers-destination  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-2 OWNERSHIP FINDING. A replacing rename lets this invocation commit over a destination another run already holds. linkSync refuses with EEXIST instead — and, because both substitution identities take that call as their seam, restoring the rename also strands them
PROVEN       tmp-removal-dropped  (WP-quarantine-preserve-durability criterion 7)
             why: the no-clobber commit leaves the bytes under TWO names, and the temp one is a secret-bearing file no preservation record, no cleanup pass and no abort message reaches — Table D row D3's own hazard
PROVEN       tmp-removal-not-gated  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-3 SECOND WINDOW. Ungating the temp removal makes this invocation unlink whatever now sits at its temp name — a path it may no longer own after an in-scope shelf edit. Only the substitution case of QPD-6 can see it; the ordinary success path removes the same name either way
PROVEN       d1-cleanup-not-gated  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-4 CONVERGED FINDING. The shared D1 catch also runs when the COMMIT throws — a state in which identity EXISTS — so an ungated cleanup there deletes a temp name this invocation no longer owns. Only QPD-6's failed-commit case can see it: on the ordinary D1 path the gate and the shipped flag agree
PROVEN       provenance-adopted-not-created  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-4 PROVENANCE FINDING. Reverting to a pathname write followed by a separate open ADOPTS whatever name resolves to at that instant instead of holding the inode this invocation created — so a substitution between the two is written over, linked as the artifact and reported as a success. The exclusive create IS the provenance, and this is the only mutation that can take it away
PROVEN       create-failure-removes-unowned  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-5 INHERITED-INFERENCE FINDING. The shipped `tmpOwned` disjunction existed because `writeFileSync` combined the create and the write, so a post-create write failure could leave an entry behind. A standalone `O_CREAT|O_EXCL` open is atomic: a throw creates nothing. Carrying the old heuristic makes a non-EEXIST create failure remove whatever is at the temp name — a file this invocation never created. Only QPD-6's create-failure case can see it; every other case has a descriptor
PROVEN       ownership-check-not-failure-closed  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-6 EVIDENCE FINDING, RE-DERIVED AT ROUND 9 ONTO THE THREE-VALUED PREDICATE. `ownsName` is load-bearing at four gates, and the arm that makes a DELETED name harmless is the `ENOENT` one. Remove it and every ENOENT becomes the indeterminate raise, so the in-scope adversary deleting a name before a gate turns a preservation that should quietly continue into a loud failure. Only the two ENOENT cases can see it: every other case runs against a name that exists or against a stat that fails some other way
PROVEN       ownership-check-narrows-to-number  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-6 PRECISION FINDING. `bigint` is in the exact form for a reason an implementer cannot see from the line alone: two DISTINCT inode numbers above Number.MAX_SAFE_INTEGER collapse to the SAME Number, so a comparison narrowed to Number matches a substituted name and authorizes its deletion. Only QPD-5 doctored-stat case can see it; every ordinary inode compares the same either way
PROVEN       post-commit-descriptor-leaked  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-7 DESCRIPTOR-LIFETIME FINDING. The post-commit temp removal deliberately raises Table D row D3, and it runs while the descriptor must still be open for the read-back and the flush. Without the finalizer that throw escapes with the descriptor open, contradicting row F8 letter for letter — and every QPD identity stays green, because none of them looks at descriptor closure. The identity that sees it is the SHIPPED D3 test this package extends, which already owns that scenario
PROVEN       returned-bytes-rereads-the-artifact  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-7 LINEARIZATION FINDING. The guarantee is about the RETURNED BYTES — the ones this invocation created, then READ BACK AND COMPARED after a flush of that inode had completed — and not about what the name holds afterwards. Re-reading the artifact at the end would return whatever a same-UID hand wrote in the meantime, silently converting a narrow true claim into a broad false one. Only QPD-5 in-place-overwrite case can see it: every other case either fails, or re-reads bytes that did not change
PROVEN       read-before-flush  (WP-quarantine-preserve-durability criterion 1)
             why: THE ROUND-8 ORDERING FINDING. Reading the artifact back BEFORE the flush lets an overlapping same-inode write land between the comparison and the fsync, so the bytes returned were read BEFORE any flush of that inode completed while the fsync that did complete ran over newer contents — the order row F6 fixes is inverted for exactly that interval. Which flush completed before which read is all this proof is about; whether a completed flush covered the returned bytes is row F10 (v)'s subject and is not claimed here. Only QPD-5 artifact-flush-seam case can see it — every other case either has no overwrite, or has one placed after the flush where both orderings agree
PROVEN       ownership-check-collapses-indeterminate  (WP-quarantine-preserve-durability criterion 6)
             why: THE ROUND-9 FINDING, and it is the whole of Dispatch precondition item 8. Catching EVERY stat error as `false` conflates 'the name is demonstrably not mine' with 'I could not look'. Collapsed, an indeterminate stat at the post-commit temp gate reports SUCCESS with this invocation's own secret-bearing link still at a deterministic pid-derived name, and one at either `dest` gate returns `null` with this invocation's `dest` still present — against Table D rows D3 and D4. Only the four indeterminate-stat cases can see it; every other case is ENOENT, a completed stat, or no stat failure at all
PROVEN       ownership-error-swallowed-at-last-gate  (WP-quarantine-preserve-durability criterion 6)
             why: DETECTING the indeterminate stat and PROPAGATING it are different behaviours, and one proof cannot see both. The flush-then-verify block's `catch` exists to turn a flush or read failure into a preservation failure; a bare `catch` there also swallows row F8's loud raise, and the last gate then answers `false` — after which the failure path re-checks the same name, succeeds, and DELETES this invocation's own `dest`. Only QPD-5's one-shot last-gate case can see it: QPD-6's two indeterminate cases are outside this block entirely. The find is ONE LINE on purpose: a multi-line find spanning the catch would stop matching the moment an implementer writes a comment inside it, which is exactly what happened on the rehearsal tree
PROVEN       shared-catch-descriptor-leaked  (WP-quarantine-preserve-durability criterion 7)
             why: ROUND 9 GAVE THE SHARED D1 GATE A WAY TO THROW, so the close that used to sit safely after it now has to sit in a `finally`. Without one, an indeterminate stat inside that catch escapes with the held descriptor open and row F8's 'closed in every case' is false on the one path this round adds. Only QPD-6's shared-catch case can see it: every other path through that block returns normally, where the flattened form closes the descriptor exactly as the finalizer's own does
PROVEN       directory-descriptor-leaked  (WP-quarantine-preserve-durability criterion 7)
             why: THE ROUND-10 FINDING. `flushDir` returns a boolean, and the boolean is all the rest of the protocol sees, so a form that opens the directory, flushes it, returns true and never closes satisfies every other row, produces the exact fsync sequence QPD-3 and QPD-4 assert, and passes every other declared mutation — while leaking one descriptor per chain member on EVERY successful preservation, so a long enough run exhausts the process's descriptor limit and the preservations that follow fail. Only QPD-7 can see it: no other identity records a close

Criteria:
PROVEN       WP-quarantine-banner-location criterion 4 — doctor-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 1 — banner-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 2 — summary-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-banner-location criterion 3 — warnings-shelf-claim-restored=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 1 — preservation-flush-removed=PROVEN; read-before-flush=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 2 — artifact-flush-failure-swallowed=PROVEN; directory-flush-failure-swallowed=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 3 — containing-directory-not-flushed=PROVEN; intermediate-shelf-not-flushed=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 4 — state-directory-not-flushed=PROVEN; anchor-not-flushed=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 5 — flush-order-inverted=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 6 — destination-ownership-gate-removed=PROVEN; destination-removal-not-gated=PROVEN; ownership-check-follows-symlinks=PROVEN; ownership-check-not-failure-closed=PROVEN; ownership-check-narrows-to-number=PROVEN; returned-bytes-rereads-the-artifact=PROVEN; ownership-check-collapses-indeterminate=PROVEN; ownership-error-swallowed-at-last-gate=PROVEN
PROVEN       WP-quarantine-preserve-durability criterion 7 — commit-clobbers-destination=PROVEN; tmp-removal-dropped=PROVEN; tmp-removal-not-gated=PROVEN; d1-cleanup-not-gated=PROVEN; provenance-adopted-not-created=PROVEN; create-failure-removes-unowned=PROVEN; post-commit-descriptor-leaked=PROVEN; shared-catch-descriptor-leaked=PROVEN; directory-descriptor-leaked=PROVEN

RUN: PROVEN

```

## `npm test` — summary (2630 individual test lines omitted, ~446 KB; reproducible with `npm test` on this branch)

```
ℹ tests 2630
ℹ suites 0
ℹ pass 2618
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 75662.868542
```

## `npm run lint` — full transcript

```
> wienerdog@0.13.0 lint
> node scripts/lint.js

--- markdownlint ---
Linting: 640 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 268 spec(s), 4 agent(s)

lint passed
```

thegyula:
## PR gate — rounds 1–3 (orchestrator's adjudication)

Raws under `docs/specs/logbook/2026-09-05-quarantine-durability-pr221-raw-round{1,2,3}-*`, committed before adjudication each round.

- **Round 1** (`ad8c62dc`): wd-reviewer REQUEST-CHANGES with no `src/` change requested (all 13 byte-exact forms once, declaration byte-identical, 37/37 PROVEN, 2630/2618/0/12, all six V1/V2 defeat states reproduced); plugin P2 (win32 skip guards); shadow two C (assertion count; a coverage-overclaiming comment). Fixed in `6b6c5596`.
- **Round 2** (`6b6c5596`): wd-reviewer APPROVE (every disposition re-executed); shadow + plugin converged on one P2 — the D3 test's two added asserts vs criterion 10's one. Resolved in `85514b9a` by folding the non-vacuity guard into the single closure assertion (RED proof still reddens it; finalizer removal fails it by hand). Plugin's win32 red-proofs concern: NO CHANGE — the lane refuses at load on win32 (reviewer-measured), CI is ubuntu/macos.
- **Round 3** (`85514b9a`): wd-reviewer **APPROVE**, nothing for the implementer; shadow **patch is correct**, zero findings; plugin one P2 — AIX needs a write-open descriptor for `fsync` while the win32-only branch enables durability on every other platform. **Residual:** AIX is not a product platform (README: macOS, Linux, Windows); routed to wd-architect for row F5's wording at the done-flip.

**Routed to wd-architect (dated errata at the done-flip):** the Deliverables cell's prose census of shipped-test changes → a small canonical table; the P0b clause mirroring row F3's chain membership without a checklist registration; the third stale test title (D1 "before any rename"); the unguarded `O_DIRECTORY`; the D3 observer's by-inode comment; F5's platform wording.

Clean or C-only on every channel → the gate is closed; merging under the owner's session authorization.

PR #266, opened 2026-09-17
feat(transcripts): project each transcript to its primary dialogue (WP-dream-primary-dialogue-projection)

thegyula:
Spec: docs/specs/WP-dream-primary-dialogue-projection.md

An **opt-in**, deterministic, code-only `parsePrimaryWithOutcome` in the transcript parsers, returning `{extract, gateExtract, intakeBytes, parse}`. Nothing consumes it yet — the collector wiring is the sibling package `WP-dream-primary-dialogue-collection`. **`parse()`, `parseWithOutcome()` and every golden fixture are unchanged.**

## Deliverables checklist

- [x] `src/core/transcripts/primary-dialogue.js` (new) — the projection and the taint observer
- [x] `src/core/transcripts/claude.js`, `codex.js` — one optional observer parameter, notified **before** all four context-losing returns (over-long line, depth guard, parse throw, budget exhausted) and before the classification branch; `TOOL_OUTPUT_TYPES` exported so the decided list has one definition
- [x] `src/core/transcripts/index.js` — `parsePrimaryWithOutcome`; `capExtract` lifted out of `parseWithOutcome` so both entry points cap identically from **one** read
- [x] `tests/unit/primary-dialogue.test.js`, `tests/fixtures/primary-dialogue/` (10 files, **all synthetic**), `tests/red-proofs/primary-dialogue.proofs.json` (8 declarations)
- [x] spec `status:` → In-Review (the spec enumerates no `expectRed` sets, so no table cell needed correcting and no prose was made stale)

## Verification output (tree rebased twice as `main` moved; numbers are from base `811f7abc`)

```
npm test:                        tests 2865 / pass 2853 / fail 0 / skipped 12
scoped (primary-dialogue + transcripts + dream-validate):  268 pass / 0 fail
npm run lint:                    lint passed
npm run red-proofs (UNFILTERED, whole output to a file, ~18 min):  RUN: PROVEN
    149 PROVEN lines; 0 FAILED/ERROR/VACUOUS/UNCONTROLLED/UNSUPPORTED/FILTERED
    this WP's eight — AC1, AC2, AC3, AC3b, AC3c, AC3d, AC4, AC4a — all PROVEN
boundary-check:                  exit 0 (spec's literal list, and all 17 changed files)
git status --porcelain tests/golden/:   empty
git diff --check:                exit 0
```

`expectRed` sets were **measured**, each mutation hand-applied before its declaration was written: AC1→1 test, AC2→2, AC3→1, AC3b→1 (Claude depth-guard; the Codex case is the untouched control), AC3c→1, AC3d→2 of 4 (the `attachment` and unfamiliar-type shapes), AC4→1, AC4a→1.

## Reference-model cross-check

All **35** preserved sequences were fed through the shipped projection, driven exactly as the parsers' per-line loop drives it (including the depth-guard and parse-failure paths), and compared per message with `docs/specs/logbook/2026-09-17-dream-primary-dialogue-taint-model.js`: **35 / 35 agreement.**

Beyond the model's own cases, **three Table-A corners the 35 sequences do not discriminate** — where this implementation and the model *differ*. In each, the implementer ruled the **model** wrong under Table A's wording, and chose the fail-closed reading:

1. A Codex `message` payload whose `content` is **not an array** — implementation taints and emits nothing; the model leaves `false`. A5c-blocks: "a Codex `message` payload owes an array".
2. A Codex rollout with **no `session_meta` at all** — implementation supplies no dialogue; the model emits. A4 grants dialogue only when *its first `session_meta` record's* `thread_source` is `"user"` or absent.
3. A Claude **assistant** record with `isSidechain: true` — implementation declines it; the model checks the flag only on user records. A4 says "a Claude **record** carrying `isSidechain: true`".

## Decisions made

- Observer shape `{lost(), record(obj)}`, one optional 4th parameter per parser.
- Row A6's `truncated` is seeded from the **raw capped** extract's `truncated` — the value `parseWithOutcome` returns today.
- An array-valued user content whose `text`-block join is empty emits no message, mirroring A2's assistant clause and the base parser.
- A per-case corpus (`taint-cases.json`: 4 AC3b, 28 AC3c, 7 AC3d) is materialised to temp `.jsonl` and run through the **real** parser, one test identity per case; each asserts the silent `ok`/`0`/`false` triple, the exact projected texts (so a tainting record is shown to supply no dialogue) and the exact flags.
- The single-read criterion spies on `streamLines` through a private copy of the module graph, because `claude.js` destructures the binding at load time; debit equality alone is vacuous, as AC4a warns.

## Discovered issues (out of scope, not fixed) — routed to wd-architect

- **Stale cite:** #264 moved `src/core/dream/scratch.js`; the B4 anchor `Buffer.byteLength(JSON.stringify(extract))` is now at `:128`, not `:118`, and the write at `:139`, not `:129`. The expression is unchanged, so `intakeBytes` is unaffected.
- **AC3b's G4 clause cannot be satisfied end to end.** Budget exhaustion is terminal for the file (`stream.js:129-138` returns immediately), so no retained assistant message can follow the loss. The taint is implemented and asserted directly on the projection object, plus `runExhausted`/`truncated` end to end; the clause needs a spec amendment.
- **The reference model disagrees with Table A in the three corners above** — the model, not the implementation, needs correcting.
- `tests/integration/reap-escape.test.js` fails under concurrent load and passes on a quiet machine — do not run `npm test` alongside `npm run red-proofs`.

## Known limitations

An unlisted content-block type, or an unfamiliar Codex `payload.type`, taints the rest of the session with **no diagnostic** — by design (owner items 4 and 6). G4's taint is unobservable through the function's return value.

## Review gates — three rounds; both clean on the FINAL tip `559cff8e`, base `7ff44508`

| Round | Tip | Codex plugin `review` (`gpt-6-astra`) | wd-reviewer | Outcome |
|---|---|---|---|---|
| 1 | `4008775d` | **[P2]** a malformed FIRST `session_meta` left the header latch unset, so a later (possibly copied-history) header enabled dialogue and user text came out trusted — contrary to row A4 | APPROVE (saw the same behaviour, judged it harmless because the header taints) | orchestrator upheld the finding: taint governs assistant text only; user text is `false` by role → **fix** |
| 2 | `b9f59860` | **[P2]** a Claude `text` block whose `text` is not a string made the join THROW, aborting the parse; other values invented dialogue | APPROVE (latch verified; 12 header sequences; 360 first-law comparisons, 0 mismatches) | **fix** |
| 3 | `559cff8e` | **clean** — "preserves existing parser behavior and follows the specified acceptance, provenance, and accounting contracts"; 51 projection cases read-only; suite not run in its sandbox | **APPROVE** — see below | **merge** |

**Round 3, wd-reviewer, everything executed on `559cff8e`:** `npm test` 2878 / 2866 / 0 fail; targeted 97/97; lint; boundary-check exit 0; **unfiltered `npm run red-proofs` → `RUN: PROVEN`, 151 PROVEN, zero non-PROVEN, all TEN `pdp-*` ids.** Rule ruling: decline-without-taint is what the spec SAYS (A5c-blocks states the block obligation exhaustively; A2 accepts only a `text` VALUE; a decided-but-declined block never taints; an all-declined `text` record behaves byte-identically to A5c-blocks' own `image` control), and the asymmetry with the Codex non-array `content` case is justified by Table A (a step-2 record-schema obligation vs a step-4 value). Nine hostile values (poisoned `toString`, poisoned `valueOf`, object, number, null, array, boolean, nested, null-prototype) → **zero throws, zero invented text**, sound sibling block kept, no taint. **First law: 512 comparisons** (64 fixtures × 4 budgets × both entry points, both harnesses) → 0 mismatches; `tests/golden/`, `src/core/dream/`, `skills/`, `templates/`, `docs/adr/` untouched; spec diff the status flip only. New mutation replayed: exactly six red, control green. Regression sweep matches round 2. Fixtures synthetic. Porcelain identical before/after, every round.

**Confirmed pre-existing, DoS-shaped, routed (not fixed here — the first law forbids it):** `src/core/transcripts/codex.js:77-83` `extractMessageText` throws on a Codex `input_text`/`output_text` block with a poisoned `toString`, on `main` and on this tip identically, via `parse()` and `parseWithOutcome()`; `src/core/dream/scratch.js:120` calls the parser inside `collectExtracts`' loop with **no try/catch**, so one crafted rollout aborts the whole nightly dream — and since the file is never quarantined, it aborts it **every night** until removed. Fix shape: a per-file try/catch that quarantines the file. The projection's own Codex guards are therefore unreachable end to end until that lands; they are exercised directly (both roles) and their removal reddens.

**Non-blocking notes for wd-architect:** the two Codex guards have no RED id of their own (registry completeness); the string-type obligation on a block's `text` exists in no canonical cell — one clause in A5c-blocks, mirrored into AC3c; A2's empty-join drop is spelled out for assistant only.

## Lessons

- WP-dream-primary-dialogue-projection: the rule-as-procedure worked exactly as intended — coding the ordered steps verbatim gave 35/35 against the reference model on the first run; the only judgment calls left were the three cases the 35 sequences never exercise.
- WP-dream-primary-dialogue-projection: a reference model is a floor, not a ceiling — feeding it the UNCOVERED corners of the contract found three places where it disagrees with the contract.
- WP-dream-primary-dialogue-projection: spying on `streamLines` needs a private copy of the module graph, because the parser destructures the binding at load time.
- WP-dream-primary-dialogue-projection: do not run `npm test` alongside `npm run red-proofs` — a fork-timing integration test flakes under load.
- WP-dream-primary-dialogue-projection: `npm run red-proofs` buffers its whole ~18-minute output, so a growing log is not a progress signal; budget a full re-run after every rebase.

Generated-by: Claude Opus 5 (implementation); orchestrated by Claude Fable 5.1


thegyula:
## Gate round 1 on tip `4008775d` — merge HELD: the two gates disagree, and the orchestrator sides with the finding

**Codex plugin `review`, `gpt-6-astra` (independent): one finding, reproduced.** *[P2] Latch the first Codex header even when its payload is malformed* — `src/core/transcripts/primary-dialogue.js:242-244`. If the first `session_meta` has a missing or non-object payload, the step-2 return leaves `sessionMetaSeen` false; a later header with `thread_source: "user"` (or the key absent) then enables dialogue. Reproduced with `payload: null` followed by a valid header: subsequent user text is emitted with `derived_from_untrusted: false`, contrary to row A4's first-header contract.

**wd-reviewer (spec fidelity): APPROVE**, with everything executed — its own 24 adversarial sequences through the real `parsePrimaryWithOutcome` all matched; the first law proven (`parse()` and `parseWithOutcome()` byte-identical to `main` over 26 fixture pairs, five budgets and the error paths); `gateExtract` vs the real `invocationWindowTainted`: 0 divergences in 59 comparisons; `intakeBytes` equal on 26/26; the worked example byte-exact; the single-read spy judged genuine; fixtures fully synthetic; four of eight mutations replayed independently; unfiltered red-proofs `RUN: PROVEN`; `npm test` 2865 / 0 fail; lint clean. It ruled **with** the implementer on all three model-vs-contract disagreements. It *saw* the same first-header behaviour (its finding 7) and called it harmless, because the bad header already taints permanently.

**Adjudication.** Taint governs *assistant* text only; user text is `false` by role. Row A4 exists to stop a subagent rollout's copied user-role history being counted as a second human — and the spec's own Current state notes that later metadata may be copied history. A damaged true first header followed by a copied one is exactly that case, and that trusted user text can count toward the recurrence threshold for identity-level memory. Row A5c already says a malformed header "supplies no dialogue rather than defaulting to accepted". So this is an implementation bug against the spec: **fix required**, fail closed — a rollout whose first header is malformed is ineligible for the rest of the file.

Returned to the implementer with the fix, three fixtures (malformed-then-valid → no dialogue; non-object payload → same; valid-then-malformed control), and wd-reviewer's two small code notes (three unused exports; the single-read spy covers only the Claude path). **Both gates re-run on the new tip.** Spec-side notes (A2's `isSidechain` mirror, A4's silence on a header-less rollout, AC3b's unsatisfiable G4 clause, one stale cite) go to wd-architect as post-merge errata.

CI on `4008775d`: seven checks pass.

thegyula:
## Gate round 2 on tip `b9f59860` — merge HELD again

**Codex plugin `review`, `gpt-6-astra` (independent): one finding, reproduced.** *[P2] Validate text values before joining user content blocks* — `src/core/transcripts/primary-dialogue.js:73-75`. A Claude user record containing `{"type":"text","text":{"toString":null}}` makes the join throw `TypeError: Cannot convert object to primitive value`, aborting `parsePrimaryWithOutcome` instead of returning a parse result; the default parser ignores such blocks and processes the following records, so the crash is newly introduced. Other non-string values would produce invented dialogue (`[object Object]`).

Confirmed by the orchestrator: `joinTextBlocks` maps `block.text` straight into `join` with no string check. Returned to the implementer: never throw on any block shape; a non-string `text` never becomes dialogue; the same check on the Codex `output_text`/`input_text` joins; fixtures for object/number/null/array values plus a control; a RED declaration if the convention calls for one. **Both gates re-run on the next tip.** CI on `b9f59860`: seven checks pass.

Two rounds, two real defects from the independent gate, both in the same file, both on hostile input shapes the spec-fidelity gate's adversarial set did not include (a malformed first header; a non-string block value). Worth a lesson for the inbox: the fidelity gate's attack set should include structurally malformed *values*, not only malformed *types*.

PR #269, opened 2026-09-18
feat(dream): collect primary dialogue into scratch (WP-dream-primary-dialogue-collection)

thegyula:
Spec: docs/specs/WP-dream-primary-dialogue-collection.md

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table (`node scripts/boundary-check.js` exit 0 over all 13 changed files)

Base `fa9fb519` (origin/main). The branch was rebased once mid-run when #268 landed (docs-only upstream delta); every number below is from the rebased tree.

## Verification output

```
node -e "…parsePrimaryWithOutcome…"                                  → predecessor present
test -f skills/wienerdog-dream/SKILL.md && ! rg -q "skill_invocations|tool_result|errored" skills/wienerdog-dream/SKILL.md
                                                                      → exit 0 (occurrences 3 / 8 / 3 → 0 / 0 / 0)
test -f docs/adr/0020-skill-revision-lifecycle.md && rg -qF 'Status: **ACCEPTED under standing authorization 2026-09-17 — owner signature pending.**' …
                                                                      → exit 0
test -f docs/adr/0020-skill-revision-lifecycle.md && ! rg -qi "owner (approved|accepted|ratified|signed)" …
                                                                      → exit 0
npm test -- dream-collect, dream-pipeline, dream-skill-structure, dream-validate, integration/dream
                                                                      → 382 pass / 0 fail
npm test                                                              → 2895 tests, 2883 pass, 0 fail, 12 skipped
npm run lint                                                          → lint passed (markdownlint 0 errors, shellcheck, shfmt, frontmatter 279 specs)
npm run red-proofs   (UNFILTERED, output captured to a log, own PID awaited)
                                                                      → RUN: PROVEN — 157 PROVEN, zero FAILED/VACUOUS/UNCONTROLLED/FILTERED/ERROR
node scripts/boundary-check.js docs/specs/WP-dream-primary-dialogue-collection.md <13 changed files>
                                                                      → exit 0
git diff --check                                                      → exit 0
git status --porcelain tests/golden/                                  → empty
```

RED mutations hand-applied and measured BEFORE each declaration was written, then confirmed by the unfiltered run:

- `scratch.js`: `const extractBytes = intakeBytes;` → `Buffer.byteLength(JSON.stringify(extract))` reddens exactly `[PDC-AC1]` (1 test).
- `scratch.js`: eviction `gateExtracts.delete(evicted)` neutered → reddens exactly the two `[PDC-AC1b]` tests.
- `dream.js`: `extractsBySession = sel.gateExtracts` → the old re-read loop → reddens exactly `[PDC-AC4a]`, `[PDC-AC4b]`, `[PDC-AC4c]` (3 tests).

Skill digest regenerated with the shipped formula (`src/core/runtime-settings.js:119` `loadVendoredSkill`); all four entries recomputed, only `wienerdog-dream` changed (`60fde183…` → `86316906…`), the other three recomputed byte-identical.

ADR-0020 amendment appended byte-for-byte after the 2026-09-05 amendment, immediately before `## Future work (parked, not specced)`; Status line `ACCEPTED under standing authorization 2026-09-17 — owner signature pending.` occurs once. Nothing claims the owner approved, accepted, ratified or signed it.

**AC7 measurement:** `docs/specs/logbook/2026-09-18-dream-primary-dialogue-collection-ac7-measurement.md`. Synthetic corpus (360 files, 10,278,150 bytes), disposable temp home / empty ledger / temp state dir; BASE column is the real base-commit collector (`git show b46a3843:src/core/dream/scratch.js` dropped into a copy of `src/`), not a re-implementation.

- Regime 1 (capacity stop both runs; X = 2,000,000): BASE 81 admitted / 2,017,890 scratch bytes; PROJECTED 81 admitted / 65,331 scratch bytes. Admitted lists identical element-for-element; all five arms identical. Row C1 holds; the same 81 sessions reach the model as 3.2 % of the previous volume.
- Regime 2 (preprocessing deadline binding both runs; timeout 0.06 s, real wall clock): BASE 50 admitted / 310 deadline-deferred; PROJECTED 44 / 316. Six fewer under projection (−12 %), same newest-first prefix, 0 sessions admitted only by PROJECTED; no byte verdict moved (`deferred`, `oversized`, `readDeferred`, `newlyQuarantined` all 0 both sides) — only the visited set.
- Limitation stated in the entry: synthetic backlog, deadline regime reached by configuring the timeout low; Regime 2's sign is one machine's (added classification cost − saved write cost).

## Decisions made

1. **AC1b's write order contradicts its own non-vacuity requirement, so both orders are tested.** `validate.js:474` `SID_RE = /^[a-z0-9]+:[A-Za-z0-9_-]+$/` is exactly `sanitize`'s alphabet, so any id colliding with `s_1` without being `s_1` is also a Session-ID schema violation. With survivor `s_1` (the order AC1b names), a ledger counting the overwritten `claude:s.1` is refused as *malformed Session-ID* whatever the map holds — the refusal is over-determined and the no-eviction control is impossible. Kept the spec's order for the survivor/empty-projection/nothing-else-moved assertions and asserted the over-determination explicitly; added a second test in the other order (`s.1` survives, overwritten `s_1` is schema-clean) that carries AC1b's actual authorization claim: refused as *not among this run's processed extracts*, **accepted** by a map built without the eviction.
2. **A third RED proof, criterion `1b`.** The Mirrored Surface Checklist names only AC1 and AC4 as proof-backed; row C4's eviction is the one line with an authorization consequence, so it is declared too.
3. **The base-commit collector is reproduced by substituting the entry point, not by copying the loop.** `collectBaseline` mocks `parsePrimaryWithOutcome` to return the raw extract with its own compact length as `intakeBytes` — same number, order and written bytes as the base commit, no second admission loop to drift.
4. **Three WP-087 tests retargeted rather than deleted** (`whole admission preserves early/late skill invocations`, `…the parser result for a trailing invocation`, `…the MAX_MESSAGES cap`): the invocation geometry now lives on `gateExtracts`; the tests assert it there and additionally assert the scratch extract has no `skill_invocations` and no tool text.
5. **Spec JSDoc typo not copied:** the spec's `collectExtracts` return block lists `skippedQuarantined` twice; written once.
6. **AC2's literal file is compared with `source_path` substituted** (the literal's `/samples/rollout-demo.jsonl` cannot exist in a temp home); key order, the added per-message key, `null`s and no trailing newline are compared byte-for-byte.

- **Spec contradiction, AC1b** (Decision 1) — prose defect for the architect; the spec was not edited.
- **Spec arithmetic, "Literal expected scratch file":** prose says the projection "grows by **30** bytes per message"; `,"derived_from_untrusted":false` is **31** compact bytes and the spec's own literals (233 − 202) already differ by 31. The test asserts 31 with a comment.
- **Prose made stale by measured `expectRed` sets** (not edited): both `tests/red-proofs/…` Deliverables notes; the Implementation-notes bullet "The RED proofs' `expectRed` sets are DERIVED, not measured"; the Mirrored Surface Checklist clause "`npm run red-proofs` mirrors AC1 and AC4" (now AC1, AC1b and AC4); AC4's clause "…which is case (c) turning red" is true but incomplete — (a) and (b) redden too and the declaration names all three.
- **ADR-0042 runner contract worth writing down:** `signal` is searched in the failing test's TAP **diagnostic**, so a bare `assert.deepEqual`/`assert.ok` can never carry it. The first unfiltered run reported `FAILED` on all three declarations for missing signals, not a wrong red set; `119137d2` puts the tag in each reddening assertion's message.
- **Pre-existing, retargeted in-scope:** `dream-collect.test.js` `deadline includes discovery and scratch setup` asserted `parseWithOutcome` callCount 0 — after the call-site switch that would have been a vacuous green; retargeted to `parsePrimaryWithOutcome`.
- **Pre-existing, untouched by design:** the `sanitize` collision still silently loses one session's dialogue while marking both processed (the spec's own Discovered issue); row C4 only stops the authorization gate weakening over it.

## Lessons (for `memory/lessons/inbox.md` on main)

- WP-dream-primary-dialogue-collection: ADR-0042's `signal` is matched against the failing test's TAP **diagnostic**, not its name — a proof whose reddening assertion is a bare `deepEqual`/`ok` reports `FAILED` with a correct red set. Put the tag in the assertion message before declaring.
- WP-dream-primary-dialogue-collection: reproduce a "what the base commit did" comparison by substituting the entry point rather than copying the loop.
- WP-dream-primary-dialogue-collection: `validate.js`'s `SID_RE` is exactly `sanitize`'s alphabet, so a sanitize-collision fixture is non-vacuous only when the schema-clean id is the overwritten one.
- WP-dream-primary-dialogue-collection: switching a shared call site silently turns every `t.mock.method(transcripts, 'parseWithOutcome')` into a vacuous green — grep the old symbol across test files.
- WP-dream-primary-dialogue-collection: `neutralise()` rewrites `'`, `:` and `/` in a reported refusal reason, so a pipeline assertion against the raw reason string never matches, and a `doesNotMatch` written that way is vacuous.
- WP-dream-primary-dialogue-collection: `git fetch` before starting the 20-minute red-proofs run — origin/main moved mid-run and the run had to be killed, rebased and redone.

---
Generated-by: Claude Opus 5


thegyula:
## Gate round 1 on tip `119137d2` — both gates clean, CI 7/7, merging

**Codex plugin `review`, `gpt-6-astra` (independent, detached worktree, porcelain identical before/after):** *"No actionable regressions found. The collector preserves intake-based admission and supplies original-timeline evidence to the ledger gate while writing primary dialogue to scratch."* Its sandbox is read-only, so it ran the 17 skill-structure tests and syntax checks only; the executed evidence is the fidelity gate's and CI's.

**wd-reviewer (spec fidelity): APPROVE**, everything executed in its own detached worktree at the tip: `npm test` 2895 / 2883 pass / 0 fail / 12 skipped; lint clean; boundary-check exit 0 over the 13 changed files; all four literal verification commands exit 0; the three RED mutations hand-applied one at a time — `extractBytes` → projection size reddens exactly `[PDC-AC1]`; the C4 eviction neutered reddens exactly the two `[PDC-AC1b]`; `extractsBySession` rebuilt from scratch files reddens exactly `[PDC-AC4a/b/c]` (the untainted `[PDC-AC4]` control stays green); the `wienerdog-dream` digest recomputed byte-identical with `loadVendoredSkill`'s formula and the other three pins unchanged; the ADR-0020 block diffed byte-identical against the spec's literal, pure insertion (`73 0`) in the right position, no ownership claim anywhere in the diff; row C5's `printPlan` lines byte-exact. It verified Decision 1's premise itself (`validate.js:474` `SID_RE` is exactly `sanitize`'s alphabet, so AC1b's named write order makes its own non-vacuity control impossible) and judged the two-order handling the best available; Decision 2 (a third RED proof for criterion 1b) a small, defensible addition it would not have removed.

**Adjudication.** No finding against the diff. Everything the fidelity gate raised is spec-side and goes to wd-architect with the done-flip as errata: (1) AC1b split into two named cases recording the `SID_RE` ≡ `sanitize` over-determination; (2) `:348` "mirrors AC1 and AC4" → AC1, AC1b, AC4; (3) `:441` "DERIVED, not measured" → measured at `119137d2`; (4) `:632` AC4's "case (c)" → (a), (b) and (c) all redden; (5) `:284` 30 → 31 bytes per message; (6) note the `source_path` substitution in AC2's literal comparison; (7) `:699` — `npm run red-proofs -- --wp <id>` exits 1 by construction (`RUN: FILTERED`), so the unfiltered run is the check; (8) the AC7 logbook's `0000…0066` range notation does not account for 81 admitted sessions; (9) register the nine `SKILL.md` sentences pinned by `dream-skill-structure.test.js` as a mirror of D2/D3/D6. Correction to this PR body: the two `tests/red-proofs/…` Deliverables cells are NOT stale (they say only which suite each file declares for).

CI on `119137d2`: seven checks pass.

PR #271, opened 2026-09-18
fix(dream): set aside a transcript whose preparation throws instead of aborting the run (WP-dream-collect-parse-throw-quarantine)

thegyula:
Spec: docs/specs/WP-dream-collect-parse-throw-quarantine.md

The collector called the parser inside its admission loop with no `try`/`catch`, while every other exclusion arm in that loop assumed the parser **returns**. It does not always return: both harness parsers join content blocks with `Array.prototype.join`, which coerces, so a block whose `text` is an object with a poisoned `toString` throws `TypeError: Cannot convert object to primitive value`. That throw escaped `collectExtracts`, escaped the dream and ended the run — and because the file was never set aside, it ended the run again the next night, and every night after.

The admission loop now carries a **per-candidate fault boundary**: the content-derived steps of one iteration run inside one `try`, and a candidate whose preparation throws is set aside as a new quarantine reason, `parse-threw`, exactly as an over-ceiling file is. `writeFilePrivate` and everything after it stay outside the boundary — a write failure is environmental and must keep ending the run loudly — and `sanitize` gains a 128-character width bound, which is what makes that split safe.

## Deliverables checklist

| Action | Path |
|--------|------|
| modify | `src/core/dream/scratch.js` — Table B (the fault boundary, rows B1–B6) and Table A rows A1, A5 |
| modify | `src/core/dream/ledger.js` — Table A row A3 (the `reason?:` union and `INFORMATIONAL_QUARANTINE_REASONS`); no function body changed |
| modify | `src/core/dream/warnings.js` — Table A row A4 (one `GROUPS` row) |
| modify | `tests/unit/dream-collect.test.js` — Table B and Table A rows A1, A5, A6 (appended after the three prior packages' blocks) |
| modify | `tests/unit/ledger.test.js` — Table A row A3 |
| modify | `tests/unit/dream-warnings.test.js` — Table A row A4 |
| create | `tests/fixtures/dream/transcripts/codex-poisoned-text-block.jsonl` (in-parse failure, byte-exact) |
| create | `tests/fixtures/dream/transcripts/codex-poisoned-session-id.jsonl` (post-parse failure, byte-exact) |
| create | `tests/red-proofs/dream-collect-parse-throw.proofs.json` — Table C |
| modify | `docs/specs/WP-dream-collect-parse-throw-quarantine.md` (`status:` flip — always allowed) |

`node scripts/boundary-check.js docs/specs/WP-dream-collect-parse-throw-quarantine.md $(git diff --name-only origin/main...HEAD)` exits **0**. `git status --porcelain tests/golden/` is empty — no golden fixture was touched.

## Verification output

```
$ npm test
ℹ tests 2904
ℹ suites 0
ℹ pass 2892
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 59492.591958
```

```
$ npm run lint

--- markdownlint ---
Linting: 689 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 279 spec(s), 4 agent(s)

lint passed
```

```
$ npm run red-proofs          # UNFILTERED, 27 min, whole output to a file, waited on its own PID
RUN: PROVEN

PROVEN       pt-boundary-drops-instead-of-quarantining  (WP-dream-collect-parse-throw-quarantine criterion 1, 2, 3, 4, 6)
PROVEN       pt-reason-literal-pinned  (WP-dream-collect-parse-throw-quarantine criterion 1, 2, 3)
PROVEN       pt-only-parse-is-caught  (WP-dream-collect-parse-throw-quarantine criterion 3)
PROVEN       pt-sanitize-unbounded  (WP-dream-collect-parse-throw-quarantine criterion 8)

PROVEN       WP-dream-collect-parse-throw-quarantine criterion 1, 2, 3, 4, 6 — pt-boundary-drops-instead-of-quarantining=PROVEN
PROVEN       WP-dream-collect-parse-throw-quarantine criterion 1, 2, 3 — pt-reason-literal-pinned=PROVEN
PROVEN       WP-dream-collect-parse-throw-quarantine criterion 3 — pt-only-parse-is-caught=PROVEN
PROVEN       WP-dream-collect-parse-throw-quarantine criterion 8 — pt-sanitize-unbounded=PROVEN
```

No `FILTERED`, `VACUOUS`, `UNCONTROLLED`, `FAILED` or `ERROR` verdict anywhere in the run (grep over the whole log returns nothing). Criterion 10 satisfied.

```
$ test "$(grep -c "reason: 'parse-threw'" src/core/dream/scratch.js)" = 1 && echo "ONE SITE OK"
ONE SITE OK

$ test -f src/core/dream/scratch.js \
    && grep -q '} catch {' src/core/dream/scratch.js \
    && ! grep -qE '\} catch \(' src/core/dream/scratch.js \
    && echo "UNBOUND CATCH OK"
UNBOUND CATCH OK
```

```
$ node -e "…"          # the spec's boundary gate, all three cases, verbatim
PARSE-THROW BOUNDARY OK
```

## RED mutations — measured, not predicted

Each `expectRed` set was measured by hand-applying the mutation and running `node --test --test-name-pattern '\[PT-' tests/unit/dream-collect.test.js`, then reading which tests reddened and with which failure `code`. Every red below is `ERR_ASSERTION` and every diagnostic carries its `[PT-n]` signal in the reddening assertion's own message.

| Mutation | Tests that reddened | Signal / code |
|----------|--------------------|----------------|
| `pt-boundary-drops-instead-of-quarantining` — drop the `newlyQuarantined.push(…)` from the `catch` arm, leaving the bare `continue;` | `[PT-1]`, `[PT-2]`, `[PT-3]`, `[PT-4]`, `[PT-6]` | each its own tag, all `ERR_ASSERTION` |
| `pt-reason-literal-pinned` — `reason: 'parse-threw'` → `reason: 'read-error'` | `[PT-1]`, `[PT-2]`, `[PT-3]` | each its own tag, all `ERR_ASSERTION` |
| `pt-only-parse-is-caught` — the `catch` arm gains `if (extract !== undefined) throw new Error('RP_MUT_PT_EXTENT');` before the push | `[PT-3]` **only** | `[PT-3]`, `ERR_ASSERTION` (via `assert.doesNotThrow`) |
| `pt-sanitize-unbounded` — `.slice(0, SCRATCH_ID_MAX_CHARS)` → `/* marker */` | `[PT-8]` **only** | `[PT-8]`, `ERR_ASSERTION` (via `assert.doesNotThrow`) |

The load-bearing claim held exactly as Table B row B9 states it: under `pt-only-parse-is-caught` — the plausible wrong implementation, a boundary wrapping the parse call and nothing after it — criteria 1 and 2 stay **green** and only criterion 3 reddens. That is what makes criterion 3 an observation of the boundary's *extent* rather than a restatement of criterion 1.

Two of the four sets are **wider** than Table C predicted; see Discovered issues 2. The spec names those exact sentences as ones the implementer must re-read once the sets are measured, which is what happened.

## Decisions made

1. **`extractBytes` keeps its exact declaration spelling inside the boundary.** The spec's implementation note says to hoist `extractBytes` to a `let` above the `try`. Doing that re-spells `const extractBytes = intakeBytes;`, which is the **exact substring** `WP-dream-primary-dialogue-collection`'s declared RED mutation `admission-measured-on-intake-not-projection` targets — the unfiltered run caught it as `"find" occurs 0 time(s)`, verdict `ERROR`, i.e. that package's row C1 guarantee silently unprovable. The declaration keeps its spelling inside the `try`, and the value the run's capacity is reduced by after the write is carried on a separate hoisted `admittedBytes`. Behavior is identical; the second commit is that fix, kept separate so the cause is legible.
2. **`parse` is declared `let` above the `try` alongside the other three destructured names**, because the parenthesized destructuring assignment the spec prescribes requires all four to be declared outside.
3. **Criterion 4's eviction clause is asserted in its checkable form.** A genuine scratch-filename collision is not constructible for a set-aside candidate: a candidate whose preparation throws never derives a filename at all (the throw is at or before `sanitize`). The test asserts the real content of Table B row B3 instead — the run's gate map holds exactly the admitted sessions, an **earlier** admitted session's entry is still there, and the crafted candidate is absent from it.
4. **`tests/unit/ledger.test.js` gains one test beyond the pinned array.** Its Deliverables cell says "one array literal", but acceptance criterion 9 requires the day-0 / day-8 render observed **through `quarantineBannerLine`** (`hasFreshInformationalQuarantine` is module-private). The array alone cannot satisfy it. The file is inside the permission boundary; the added test asserts `parse-threw` matches `read-error` byte for byte at both ends of the window and that an unrecognized reason still never decays.
5. **`tests/unit/dream-warnings.test.js` gains two touch points**, not one: `parse-threw` joins the existing fixed-order render (which is what proves the `GROUPS` row sits **after** `read-error` and **before** the secret-exhausted row, and that the catch-all still collects unrecognized reasons), plus one new test rendering the whole document byte-exact for a lone `parse-threw` record, as "Exact contracts" gives it.
6. **`[PT-6]`'s "rewriting the file" step appends an ordinary message line** to the crafted rollout rather than replacing it with the post-parse fixture. Measured reason: with the post-parse fixture there, the `pt-only-parse-is-caught` mutation makes that third run throw a raw `Error`, not an assertion failure, and the runner refuses any red whose `code` is not `ERR_ASSERTION`. The append is a real byte-and-fingerprint change and keeps the failure in-parse.
7. **The `@returns` JSDoc carried `newlyQuarantined:Array<object>`** on this tree — "Exact contracts" shows a fully typed element. Widened that one field to the spec's literal shape rather than restating the whole annotation.
8. **PR title.** Definition of done item 2 reads `fix(dream): title (WP-…)`; `title` is an unfilled template placeholder (Discovered issue 1). Used the spec's own title sentence.

1. **Spec bug.** Definition of done item 2: `PR titled fix(dream): title (WP-dream-collect-parse-throw-quarantine)` — `title` is an unfilled template placeholder.
2. **Spec bug (Table C's predicted red sets are narrower than the measured ones).** `pt-boundary-drops-instead-of-quarantining` reddens criteria 1, 2, 3, 4 **and** 6 (the row's "What it proves" cell asserts two criteria); `pt-reason-literal-pinned` reddens criteria 1, 2 and 3 (the Criterion column says 1). Both are consequences of the `testNamePattern` selecting the whole `[PT-` suite: every test that observes the set-aside **record** reddens when the record stops being written, and every test that pins the **literal** reddens when it changes. Nothing is weakened — the declarations are exact, `RUN: PROVEN` — but the table's Criterion column and those two cells should be corrected to the measured sets. `pt-only-parse-is-caught` and `pt-sanitize-unbounded` matched their predictions exactly.
3. **`tests/unit/ledger.test.js`'s Deliverables cell is narrower than acceptance criterion 9 needs** — "one array literal" cannot carry the banner-decay observation criterion 9 demands. Resolved as Decision 4; the cell should be widened in the spec.
4. **A sibling package's RED proof is coupled to this file's source formatting.** `tests/red-proofs/dream-primary-collection.proofs.json` pins `const extractBytes = intakeBytes;` as an exact substring, so any later package that re-indents or re-spells that declaration breaks it — with an `ERROR` verdict that only the **unfiltered** run surfaces, and only for the package that did not write it. Worth a runbook note: after touching a hot file, check every `tests/red-proofs/*.proofs.json` whose `file` is that path, not only your own.
5. **`src/cli/doctor.js` will report a `parse-threw` count as "being skipped for a reason this version does not recognize"** — Table A row A7 and owner item 3. Left untouched: `doctor.js` is outside the Deliverables. Truthful about the count, wrong about the recognition.
6. **The parsers themselves are still unhardened** — the four content joins in `src/core/transcripts/{codex,claude}.js` and the untyped `session_meta.id`. Deliberately out of scope (owner item 1's successor package). When that lands, **both** fixtures stop throwing and criteria 1 and 3 fail loudly rather than passing over nothing; the successor must re-point or retire them together.
7. **Pre-existing, acknowledged, not fixed:** the 128-character bound widens the existing scratch-filename collision set by adding ids that share a 128-character allowlisted prefix. The spec's Out of scope routes this to `WP-dream-primary-dialogue-collection`'s own Discovered issues.

## Lessons

- WP-dream-collect-parse-throw-quarantine: A RED proof's `expectRed` set is a property of the whole `testNamePattern`-filtered suite, not of the criterion the mutation was written for. Removing the quarantine push reddened five tests across five criteria, because every test that observes the record observes that mutation. Measure the filtered suite; never derive the set from the spec's Criterion column.
- WP-dream-collect-parse-throw-quarantine: A declared mutation pins its target by **exact substring**, so re-indenting or re-spelling a line silently disarms any sibling package's proof that targets it. Re-indenting the admission loop into a `try` block broke `WP-dream-primary-dialogue-collection`'s row C1 proof, and only the unfiltered run said so — the `--wp`-scoped reading was all-green. After touching a hot file, grep every `tests/red-proofs/*.proofs.json` whose `file` is that path and check each `find` still occurs its declared number of times; that check takes seconds and the run it replaces takes half an hour.
- WP-dream-collect-parse-throw-quarantine: The `ERR_ASSERTION`-only rule bites indirectly, not just at the obvious call. `[PT-6]`'s *third* phase originally used the post-parse fixture, so the extent mutation — aimed at `[PT-3]` — made `[PT-6]` throw a raw `Error` and would have made the declaration unprovable. Check every phase of a multi-run test against every declared mutation, not only the phase the mutation targets.
- WP-dream-collect-parse-throw-quarantine: When a fault boundary's **extent** is the contract, the proof has to be a failure that happens *after* the step everyone would think to wrap. The post-parse fixture (a rollout that parses cleanly and throws at the filename derivation) is the only thing separating a correct boundary from the plausible wrong one, and the fixture-independent `pt-only-parse-is-caught` mutation is what keeps that claim from going vacuous when the parsers are later hardened.
- WP-dream-collect-parse-throw-quarantine: `let`-hoisting the parse call's destructured names above the `try` is what makes the extent mutation *expressible* — `extract !== undefined` is a precise "the parse returned" predicate only because the binding is per-iteration and declared outside the boundary. The shape of the fix and the shape of its proof were decided together.
- WP-dream-collect-parse-throw-quarantine: `Array.prototype.join` coerces, so every content join over `JSON.parse`d data is a throw site a third party controls. A list of those sites is a forbidden-set enumeration in someone else's grammar and can never be closed; a per-candidate fault boundary states our own good instead and holds for steps a later package adds.
- WP-dream-collect-parse-throw-quarantine: A deliverable cell that names a single line ("one array literal") can be narrower than the acceptance criterion it maps to. When they disagree, the criterion is the binding one — satisfy it inside the file's permission boundary and record the gap, rather than shipping a criterion that nothing observes.

---
Generated-by: Claude Opus 5


thegyula:
## Gate round 1 on tip `cac39c0f` — both gates clean, CI 7/7, merging

**Codex plugin `review`, `gpt-6-astra` (independent, detached worktree, porcelain identical before/after):** *"No actionable regressions found. The fault boundary, filename bound, and quarantine handling match the specification."* Read-only sandbox: it ran the four targeted tests and a collector smoke test; executed evidence is the fidelity gate's and CI's.

**wd-reviewer (spec fidelity): APPROVE**, everything executed in its own detached worktree at the tip: `npm test` 2904 / 2892 pass / 0 fail / 12 skipped; lint clean; boundary-check exit 0; the A1 one-site grep, A5 unbound-catch grep and the spec's `node -e` boundary gate all exit 0; the four declared RED mutations hand-applied one at a time and each reddening EXACTLY its declared set with `ERR_ASSERTION` and the `[PT-n]` tag in the diagnostic — including the load-bearing one: under `pt-only-parse-is-caught` criteria 1 and 2 stay green and only PT-3 reddens, so criterion 3 observes the boundary's extent rather than restating criterion 1. Boundary read at `scratch.js:185-227`: exactly Table B row B1's extent inside the `try`; `writeFilePrivate`, the gate-key derivation, the eviction and both map writes outside, so a set-aside candidate cannot reach the gate maps by construction; `} catch {` unbound; no new reason lifecycle; parsers and `doctor.js` untouched. `sanitize`'s `.slice(0, 128)` applied after the replace (row B6, criterion 8). `reports/warnings.md` for a lone `parse-threw` record byte-equal to the spec's block; both fixtures byte-exact.

**Sibling-proof coupling verified.** `const extractBytes = intakeBytes;` occurs once at the tip; all three `WP-dream-primary-dialogue-collection` declarations were hand-applied and still redden (`[PDC-AC1]`, `[PDC-AC1b]` ×2, `[PDC-AC4a/b/c]`); every declaration in the repo whose `file` is one of the three touched `src/` files still finds its substring at its declared count (11 declarations, 5 proof files). `remaining` is decremented by the same value as before (`admittedBytes = extractBytes` is the only fall-through path to the write).

**Adjudication.** No finding against the diff; the three advisory notes (`@type {any}` where `GateExtract` exists; a seam-mock construction that would let criterion 4's literal "colliding" words be asserted; the row-C2 comment travelling with the serialization) are recorded for the done-flip, none warrants a round. Spec-side errata to wd-architect with the done-flip: (1) DoD item 2's unfilled `title` placeholder; (2) Table C's Criterion column and two "What it proves" cells predict narrower red sets than measured (`\[PT-` selects the whole suite); (3) the `ledger.test.js` Deliverables cell is narrower than criterion 9; (4) the `dream-collect.test.js` cell's "append after the two blocks appended today" was not updated by the 2026-09-18 re-pin (the tail is collection's block); (5) the implementation note "hoist `extractBytes` to a `let`" is unexecutable — it re-spells the substring the collection package's `admission-measured-on-intake-not-projection` pins; (6) criterion 4's "earlier colliding session" clause is not constructible from file content — reword to row B3's checkable form; (7) owner item 3 says "yes, extend `doctor.js`" but the Deliverables, row A7 and Out of scope all say not changed — the implementer's CI-passing reading (leave it) was the only one; consequence: `doctor` renders a `parse-threw` count as "a reason this version does not recognize". Cross-spec: the Done spec `WP-dream-primary-dialogue-collection`'s three `scratch.js:18-20` cites for `sanitize` are now false on `main` (`:27-29`, `SCRATCH_ID_MAX_CHARS` at `:22`).

CI on `cac39c0f`: seven checks pass.

PR #288, opened 2026-09-18
fix(doctor): name the parse-threw skip instead of calling it unrecognized (WP-doctor-recognizes-parse-threw)

thegyula:
Spec: docs/specs/WP-doctor-recognizes-parse-threw.md

## Deliverables checklist

## Verification output

```
$ npm test
...
ℹ tests 2906
ℹ suites 0
ℹ pass 2894
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
(exit 0)

$ npm run lint
--- markdownlint ---
Linting: 699 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---
frontmatter check passed: 284 spec(s), 4 agent(s)
lint passed
(exit 0)

$ npm run red-proofs   # UNFILTERED
...
PROVEN       WP-doctor-recognizes-parse-threw criterion 1 — dpt-arm-removed-falls-to-unrecognized=PROVEN
...
RUN: PROVEN
(exit 0; no FILTERED/VACUOUS/UNCONTROLLED/FAILED/ERROR verdict anywhere in the run)

# Table A row A4: exactly ONE site
$ { test -f src/cli/doctor.js && test "$(grep -c 'something in the session file stopped Wienerdog from reading it' src/cli/doctor.js)" = 1 && echo "ONE SITE OK"; }
ONE SITE OK

# Table A row A7: the secret-exhausted pinned line is byte-unchanged
$ { test -f src/cli/doctor.js && test "$(grep -cF 'too many times in a row. ${PRESERVED_COPIES_POINTER}`,' src/cli/doctor.js)" = 1 && echo "PINNED LINE INTACT OK"; }
PINNED LINE INTACT OK

# The render gate (Table A rows A3, A4, A5, A6, A8), run in three states as required:

## (a) on the base tip (08de2bc3-equivalent, pre-fix): case 1 fails on the missing row
case 1: the parse-threw row is missing entirely | case 1: the catch-all lost its three unrecognized records | case 2: the parse-threw row is missing for a parse-threw-only ledger
(exit 1 — expected)

## (b) render block present, counter left out of `total`: case 2 fails
case 2: a parse-threw-only ledger renders a FALSE ALL-CLEAR — the new counter is not an addend of total | case 2: the parse-threw row is missing for a parse-threw-only ledger
(exit 1 — expected)

## (c) on the finished change:
DOCTOR PARSE-THREW ROW OK
(exit 0)
```

## RED mutation measured (ADR-0042)

`dpt-arm-removed-falls-to-unrecognized` (`tests/red-proofs/doctor-recognizes-parse-threw.proofs.json`, criterion 1): removes Table A row A2's `case 'parse-threw':` arm by hand-applying the exact-substring mutation to `src/cli/doctor.js` and running `tests/unit/doctor.test.js` under `--test-name-pattern='\[PT-1\]'` BEFORE writing the declaration. Measured result (matches the declared `expectRed` byte for byte):

```
not ok 1 - doctor: [PT-1] a parse-threw quarantine renders its own named row, and the catch-all keeps its own unrelated count
  failureType: 'testCodeFailure'
  code: 'ERR_ASSERTION'
  error: |-
    [PT-1]
    + actual - expected
      {
    +   parseThrewLine: null,
    +   unrecognizedLine: '[warn] 3 session transcript(s) are being skipped for a reason this version does not recognize'
    -   parseThrewLine: '[warn] 2 session transcript(s) are being skipped: something in the session file stopped Wienerdog from reading it',
    -   unrecognizedLine: '[warn] 1 session transcript(s) are being skipped for a reason this version does not recognize'
      }
```

Measured `expectRed` set: exactly one test — `doctor: [PT-1] a parse-threw quarantine renders its own named row, and the catch-all keeps its own unrelated count` — failing as `testCodeFailure`/`ERR_ASSERTION` with signal `[PT-1]` present in the diagnostic. This single assertion observes the new row's literal AND the unrecognized row's count together, per the spec's Table B note. File was restored to the fixed (non-mutated) state before committing; `diff` against the committed `src/cli/doctor.js` confirmed byte-identical restoration.

## Decisions made

- Named the new counter `parseThrew` and placed it (in the `let` list, the `switch` arm, and the render block) between `readError`/`secretExhausted`, mirroring the row's rendered position (Table A row A5) and `GROUPS`' order in `src/core/dream/warnings.js`. Table A row A1 leaves the name and placement to the implementer.
- Extended the existing whole-render test at `tests/unit/doctor.test.js:794` in place (per the spec's note that this test moves) rather than duplicating its seeded ledger elsewhere, and added two small dedicated tests (`[PT-1]`, `[PT-2]`) mirroring the existing `[QBL-4]` tagging convention, so the RED-proof declaration has a precise, minimal assertion to pin (per Table B's note that the assertion must observe the new row's literal and the unrecognized count together, not merely that some warn line appeared).
- Split the implementation into three commits (`fix`, `test`, `docs`) rather than one combined commit, since the Definition of done lists both a `fix(doctor):` and a `test(doctor):` commit form.

None.

## Lessons

- WP-doctor-recognizes-parse-threw: the RED-proof runner's `evaluateRed` requires the `signal` to be a substring of the failing test's own diagnostic text (`hit.node.diag.raw`), not merely present in the test's name — passing the tag as `assert.*`'s message argument (matching the existing `[QBL-4]` convention) is sufficient, since Node's TAP reporter includes that message in the diagnostic's `error` field.
- WP-doctor-recognizes-parse-threw: hand-measuring `expectRed` (apply mutation to a copy of the source, run the targeted suite with the declaration's own `testNamePattern`, capture the TAP diagnostic, then restore the file byte-for-byte via `diff`) caught the exact failure shape before the declaration was trusted — no surprises in the full unfiltered run.

---
Generated-by: Claude Sonnet 5


thegyula:
## Gate round 1 on tip `1e0d4c9e` — both gates clean, CI 7/7, merging

**Codex plugin `review`, `gpt-6-astra` (independent, detached worktree, porcelain identical):** *"The change correctly counts and renders parse-threw quarantines while preserving the catch-all and preventing false all-clear output."* Read-only sandbox; executed evidence is the fidelity gate's and CI's.

**wd-reviewer (spec fidelity): APPROVE**, executed in its own detached worktree: `npm test` 2906 / 2894 pass / 0 fail / 12 skipped; lint clean; boundary-check exit 0; both greps and the three-state render gate (state (a) by reverting `doctor.js` to base, (b) by dropping `parseThrew` from `total`, (c) on the tip) reproduce the PR body byte-for-byte; the six-line "being skipped" block rendered through `bin/wienerdog.js doctor` is byte-exact to the spec's Exact-contracts block including order; row A5 placement after `read-error` and before `secret-exhausted`; A8's catch-all still counts 3; the all-zero healthy line covered by four pre-existing tests; `tests/golden/` untouched and no golden renders the surface (measured); the RED mutation `dpt-arm-removed-falls-to-unrecognized` hand-applied — `--test-name-pattern='\[PT-1\]'` selects exactly one test, reddens `ERR_ASSERTION` with the signal in the diagnostic; every declaration in both doctor proofs files still matches its `occurrences`; the four touched files are byte-identical between the spec's pin `08de2bc3` and the merge-base `4b5d2655`. No drift; no errata for the done-flip.

Three non-blocking observations recorded for a future pass (the `:794` test title's "zero-member groups omitted" no longer observes one; `[PT-2]`'s anchor also pins `[ok]`; five intended test-side spellings of the A4 literal).

CI on `1e0d4c9e`: seven checks pass.

PR #285, opened 2026-09-18
docs(specs): re-home Table N row N2 and M-48 to the real prune locus (WP-ep2-n2-rehome)

thegyula:
Spec: docs/specs/WP-ep2-n2-rehome.md

Table N row N2 states a rule that is still exactly right and a locus that is
gone. `WP-dream-promote-in-workspace` (`4115668a`) removed the per-path loop
(`scanTokens` occurs zero times in `src/` and `tests/`) and
`WP-dream-gate-inputs-baseline-delta` removed the per-path git. Today the run
state and the precondition live in the `pruneRedacted` closure `makeGates()`
returns, and `src/cli/dream.js:1102` owns the timing. **This PR changes no rule
and no executable line** — it moves the wording onto that locus and moves all
sixteen registered mirrors in the same pass.

### Table M, as executed (16 rows, every one dispositioned)

| id | surface | verdict | what happened |
|----|---------|---------|---------------|
| M-1 | Table N row N2 (`:1760`) | EDIT | re-worded to N2R-3/N2R-4; the rule (N2R-1) is unchanged in meaning |
| M-2 | Table B row B10 (`:1572`) | EDIT, TWO CLAUSES | (i) locus clause → the closure + `src/cli/dream.js` after `promote()`; (ii) identity clause re-keyed per N2R-13(a) |
| M-3 | Table B row B12 (`:1574`) | CHECK, NO EDIT | asserted locus-free: "N2 the trigger (once per run, only after a completed B4)". Byte-unchanged |
| M-4 | Table R consequence 7 (`:2181-2219`) | CHECK, NO EDIT | asserted: opens "and this consequence restates none of it", names no locus; both dated corrections intact. Byte-unchanged |
| M-5 | B12/B13 growth story (`:2488-2501`) | CHECK, NO EDIT | asserted: cites N1/N3/N5 and D2 only, states no trigger locus. Byte-unchanged |
| M-6 | Security checklist prune bullet (`:3663-3680`) | EDIT, THREE CLAUSES | (i) "moves the CALL SITE" → duplicates it; (ii) the per-call-half clause per N2R-13(c); (iii) provenance date per N2R-13(d) |
| M-7 | AC-14's third case (`:4240`, `:4246`, `:4247`) | EDIT, ROUTING ONLY | slug + path re-pointed; the dated `:4229` sentence describing the 2026-07-28 mutation as a move into the B4 loop **stays** |
[...]

PR #293, opened 2026-09-18
test(dream): pin that the retention prune runs exactly once per run and only after a completed redaction (WP-ep2-prune-once-per-run-test)

thegyula:
Spec: docs/specs/WP-ep2-prune-once-per-run-test.md

## Deliverables checklist

## Verification output

```
$ node tests/run.js tests/unit/dream-pipeline.test.js
ℹ tests 78
ℹ pass 78
ℹ fail 0

ok V-1: one new test, nothing removed
ok V-2: both proofs carry all four spellings
ok V-3: sibling suite untouched
ok V-5: the dependency landed

$ npm test
ℹ tests 2907
ℹ pass 2895
ℹ fail 0
ℹ skipped 12

$ npm run lint
markdownlint: 0 error(s) over 705 file(s)
shellcheck / shfmt: clean
frontmatter check passed: 287 spec(s), 4 agent(s)
lint passed

$ npm run red-proofs   # UNFILTERED — no --wp
106 declared proof(s), 106 selected
...
PROVEN       ep2-prune-runs-once-per-run  (WP-ep2-prune-once-per-run-test criterion 2)
PROVEN       ep2-prune-not-once-per-redaction  (WP-ep2-prune-once-per-run-test criterion 2)
...
PROVEN       WP-ep2-prune-once-per-run-test criterion 2 — ep2-prune-runs-once-per-run=PROVEN; ep2-prune-not-once-per-redaction=PROVEN

RUN: PROVEN
(zero FAILED / VACUOUS / UNCONTROLLED / FILTERED / ERROR anywhere in the run)

$ node scripts/boundary-check.js docs/specs/WP-ep2-prune-once-per-run-test.md $(git diff --name-only origin/main...HEAD)
exit 0

(empty)
```

**Both RED mutations, measured by hand at the tip (`862f8277`), whole-suite each time:**

- **(a) duplicate `gates.pruneRedacted();` at `dream.js:1102`** → `npm test`:
  `tests 2907, pass 2894, fail 1, skipped 12` — the single failure is this
  package's new test, delete-path attempts = **4**, nothing else in the
  repository reddened.
- **(b) `pruneRedactedOriginals(...)` inserted after `completedRedactions += 1;`
  at `validate.js:1420`** → `npm test`:
  `tests 2907, pass 2894, fail 1, skipped 12` — same single failure, delete-path
  attempts = **5**, nothing else reddened.
- **Control (unmutated tree)** → `npm test`: `tests 2907, pass 2895, fail 0,
  skipped 12`, delete-path attempts = **2**.

(Table P row P-10's own numbers — `2905`/`2892`/`2893` — were measured against a
design-round candidate implementation of the test; this PR's candidate differs
in exact line count, which is why the totals here read `2907`/`2894`/`2895`. The
shape — exactly one failure, only this test, counts 4/5/2 — matches P-10
exactly, and V-4's unfiltered `npm run red-proofs` above is the evidence that
actually gates AC-3, not these hand measurements.)

## Decisions made

- The two REDACT-severity blobs in leg one's fixture (P-7) needed to be
  DISTINCT (two completed redactions, not one repeated). The spec quotes only
  one worked blob (`xY9kQ2mZ7pL4vB8nR3sT6wA1`, from the suite's existing `:1349`
  test). I generated a second 24-char blob over the same alphabet
  (`mK3jH8qW2xC9vN5bT7yG1sA4`) and verified by hand that both clear the
  entropy floor (`ScanLimits.ENTROPY_MIN_BITS_PER_CHAR = 3.5`,
  `src/core/secret-scan.js`) with no bound sensitive-keyword context, so both
  land at REDACT severity — the simpler option over inventing a shared helper
  to synthesize blobs.
- Leg one and leg two use separate `setup()` fixtures (separate temp roots)
  rather than reusing one `ctx` across both legs, since P-7 seeds a different
  directory depth (50 vs 51) and a different note set (two REDACT notes vs one
  clean note) per leg. The spec does not require a shared `ctx`, only that
  both legs run "in the same test body," which they do.
- `setup()` is called with its default `withTranscript: true` (not `false`).
  Measured: with no transcript, the pipeline reports "nothing new to dream"
  and the `reapGroup` hook (and therefore `brainWrites`) never fires, so no
  redaction can occur. The suite's own `:1349` worked example uses the same
  default for the same reason.

- Per the spec's Out of scope section: `docs/specs/done/WP-secret-fence-ep2-redact-arm.md`'s
  M-48 census cell still carries its `gap` limb for N2 even though this PR now
  supplies the detector M-48 names. Moving that cell off `gap` and citing this
  test is a documentation pass over a `Done` spec guarded by pinned digests and
  mechanized gates — routed to the architect, not done here.

## Lessons

- WP-ep2-prune-once-per-run-test: `setup({ withTranscript: false })` silences
  the fake brain entirely — the pipeline short-circuits to "nothing new to
  dream" and never calls `reapGroup`, so any `brainWrites`-based fixture needs
  the default transcript.
- WP-ep2-prune-once-per-run-test: the fake brain (`tests/fixtures/dream/fake-brain.js`)
  always runs its own fixed scripted scenario in addition to whatever
  `reapGroup` injects (it wrote a stray `EVIL.json` and produced extra accepted
  notes in both legs here) — a cardinality assertion scoped only to the
  fixture's own seeded basenames (P-6) is what keeps that scripted noise from
  contaminating the count; an unscoped assertion on "how many notes/redactions
  happened" would not have been safe against it.
- WP-ep2-prune-once-per-run-test: `t.mock.method(fs, 'rmSync', ...)` must
  re-capture the real `fs.rmSync` reference AFTER `t.mock.restoreAll()`
  between legs — capturing it once up front and reusing it for leg two would
  silently delegate through leg one's already-restored (but closed-over) mock
  reference instead of the live original.

---
Generated-by: Claude Sonnet 5


thegyula:
## Gate round 1 on tip `4309405c` — both gates clean, CI 7/7, merging

**Codex plugin `review`, `gpt-6-astra` (independent):** *"The added test and mutation proofs match the specification and exercise the intended pruning behavior without changing production code."* Read-only sandbox; executed evidence below.

**wd-reviewer (spec fidelity): APPROVE**, executed in its own detached worktree: `npm test` 2907 / 2895 / 0 fail; lint clean; boundary exactly the three Deliverables paths; V-1/V-2/V-3/V-5 green from the spec's block; AC-6 confirmed by reading M-48's cell (the dependency landed in #285); 106 declared proofs, no duplicate ids; both declared mutations hand-applied with the declaration's `testNamePattern` (selects exactly one test) — (a) duplicate invocation → 4 deletes, (b) per-redaction with end call kept → 5, the undeclared true move (c) → 3, control 2 — P-10 reproduced exactly, whole-suite red set = the new test only for (a) and (b). The three vacuity routes closed by measurement: the seam is scoped to seeded basenames (an unscoped interception would trip `quarantinePreserve`'s post-condition at `validate.js:684`); neutralized deletes keep the dir over cap so a second invocation is observable and nothing later diverges (`scanPrivateModes` finds 0600/0700 entries clean); the run genuinely completes two redactions (neutering `BLOB_B` drops the count to 1); leg two reaches the call site — making the `completedRedactions > 0` guard unconditional reddens leg two's assertion, so "only after a completed redaction" is independently detecting.

**Adjudication.** No finding against the diff. Errata for the done-flip (spec-side): P-10's whole-suite totals are the candidate implementation's (2905) — refresh with the landed 2907 and note that the delete counts are the contract; AC-2 (`:355`) says "exactly one prune delete-path firing" where P-4 rules the invocation-count vocabulary out — restate in P-4's terms; Current state (`:154-156`) counts four declaration files naming the suite — five (`dream-pipeline.proofs.json`, 2026-09-02); M-48's AC-15 census limb (`done/…redact-arm.md:4525`) still `gap` with a dated sentence — wants moving with a citation to this test, in the architect's documentation pass.

CI on `4309405c`: seven checks pass.

PR #299, opened 2026-09-18
docs(specs): mature WP-adr-0019-quarantine-uninstall-export for the design gate (WP-adr-0019-quarantine-uninstall-export)

thegyula:
Matures the Draft stub `docs/specs/WP-adr-0019-quarantine-uninstall-export.md`
into a full spec ready for the design gate. **Docs only — no `src/`, no tests.**
`status:` stays `Draft`: the design gate has had **no** round, and at least one
is required before `Ready`.

Every code citation is pinned to `5b77865f` and was read construct by construct.

## The assessment, re-verified

- **LIVE.** `quarantinePreserve` (`src/core/dream/validate.js:936`) writes the
  user's unredacted bytes under the core at `state/quarantine/` (withhold arm,
  `:1439`) and `state/quarantine/redacted/` (redact arm, `:1416`;
  `REDACTED_SUBDIR` at `:657`).
- `disposeCoreMechanics` (`src/core/manifest.js:1133-1148`) removes
  `paths.state` with `fs.rmSync(dir, { recursive: true, force: true })` and has
  **no carve-out of any kind**; its own JSDoc (`:1110-1116`) asserts
  `state/` holds *"none user-authored"*, which is false on this tree.
- ADR-0019 is `Status: Accepted` (`:3`) with the invariant at `:52-54` declared
  *"binding on all future code"*, and carries **no owner-signature line**.
- `grep -c quarantine src/cli/uninstall.js src/core/manifest.js` → **0** in both.
- No review path exists: `WP-quarantine-review-cli` is `Superseded`, and
  `doctor`'s `quarantineReport` (`src/cli/doctor.js:498`) counts the
  **transcript ledger**, a different object.

## What the spec decides

| | Decision | Level |
|---|---|---|
| Owner item 1 | **Amend** ADR-0019, not supersede it. Amendment text drafted verbatim, `Status: **ACCEPTED under standing authorization 2026-09-18 — owner signature pending.**` | owner |
| Owner item 2 (Table W **W1**/**W2**) | **Refuse-and-report.** Wienerdog writes **no copy anywhere**; `--yes` does **not** proceed | owner |
| Owner item 3 | **No** `--discard-quarantine` flag | owner |
| Table X **X5**–**X7** | Carve-out **in code**, not a manifest "preserved kind" | architect |
| Amendment ¶4 | `secrets/` is **not** carved out — **re-obtainability**, not sensitivity, is the criterion | owner item 1 |

`--yes` does not skip the refusal because `--yes` skips a **prompt** and the
shipped contract at `uninstall.js:349-351` is that *"the set of valid actions is
identical either way"* — a refusal is not a prompt. ADR-0035 governs the
direction; note that a `--yes` run today prints **no plan at all**, so a
disclosure-only design would be silent exactly where it matters.

Both candidate mechanisms only **NARROW** a deletion, so ADR-0038's `N` permits
either and chooses neither — Table X row **X5** says so explicitly. The manifest
route is refused because it would make the dream gate a writer of
`install-manifest.json` and put the survival of user text behind a file ADR-0038
states it does not authenticate.

## Composition with `WP-scheduler-replay-manifest-independent`

That package (merged as `5b77865f`, `Ready`) lands **first** and rewrites the
same two files. Table W row **W7** names every row of it that must stay
byte-unchanged — its **D2** discovery, **D3** block, **D4**/**D7** snapshot,
**D9** abort, **D5a**/**D5b**, **D11**, **D12**, **D14**, **D15**, **R2** — and
records the ordering rule: the shelf gate runs **before** D9, both aborts delete
nothing so either order is safe, and this one goes first because it reports data
we would **destroy** rather than a deletion we might fail to perform. The two
packages are **disjoint in `manifest.js`** by that package's own Deliverables
note, which fixes `disposeCoreMechanics` as byte-unchanged. Its **D12** vault
pattern is reused where it fits (Table X row **X7**) — the disclose-what-you-
preserve shape and the `ENOENT`-is-absent/anything-else-aborts rule — and not
where it does not (no ambient path, so no `contains`).

`WP-quarantine-only-copy-shelf` (`Ready`) is also in `depends_on`: it owns the
neighbouring clause in both documentation files, and its Table O row **O8**
"may be the only copy" hedge is what the refusal text is written to.

## Verification

```
$ npm run lint
markdownlint: Summary: 0 issues in 0 files
frontmatter check passed: 288 spec(s), 4 agent(s)
lint passed
```

Red-proofs were **not** run (docs-only change; no proofs file exists yet).
The spec's own verification greps were exercised against `5b77865f`:

- `grep -Fn 'canonical core; the vault is always outside it** — is binding on all future' docs/adr/0019-…md` → `53:` (hit)
- `grep -Fn 'function contains(outer, inner) {' src/core/manifest.js` → `1097:`
- `grep -Fn 'contains(dir, vaultPath)' src/core/manifest.js` → `1144:`
- `grep -n '^Status: Accepted$' …0019….md` → `3:` · `grep -n '^Date: 2026-07-06$'` → `4:`
- both guarded **negated** greps are **RED** pre-implementation (the stale
  sentences are present at `GLOSSARY.md:147` and `secret-incident.md:62`) —
  correct, and guarded with `test -f` so the deliverable-absent case reddens too.

RED-proof anchors measured with `grep -Fc` at `5b77865f`: all four pre-measurable
anchors return **1**; two proofs mutate code this package authors and say so.

## Decisions made

- `src/core/dream/promote.js:689` (*"restore from that copy while it is there"*)
  is **not** falsified and is **out of Deliverables** — "while it is there" is
  scoped to the retention cap, which this package does not touch. The stub had
  listed it for re-derivation; Current state item 10 records the sweep.
- Kept as one M package rather than split: the obvious cut would ship an
  uninstall that silently keeps the core with no explanation, and would put the
  ADR amendment in one package and the behaviour it authorizes in another.
- Title changed (id unchanged) — "export/warn step" no longer describes what is
  recommended, and the new title is true under all three of W1's candidate cells.

## Discovered issues

- `disposeCoreMechanics`'s JSDoc at `src/core/manifest.js:1110-1116` ships a
  false claim today (*"none user-authored"* of `state/`). Recorded as Table X row
  **X9** and corrected by this package's own Deliverables, not fixed here.

## Lessons

- WP-adr-0019-quarantine-uninstall-export: a markdown code span may not begin or
  end with a space (MD038), so a `grep -Fc` anchor copied byte-exact from an
  indented source line fails `npm run lint`. Show the anchor un-indented and say
  in the table's preamble that the declaration's `find` string carries the real
  indentation — the measured count is identical either way.
- WP-adr-0019-quarantine-uninstall-export: when two live specs edit the same
  sentence of the same user-facing document, `depends_on` is the tool — declare
  the ordering and name which clause each package owns, rather than writing one
  spec against two possible versions of the text.

**No owner approval is recorded anywhere in this repository for this package**,
and the drafted ADR-0019 amendment carries *"owner signature pending"*.
Do not merge without the design gate.

Generated-by: Claude Opus 5


thegyula:
## Design gate round 1 (Astra) — both findings ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `5084f7c1`; fix at **`f6f27d3e`**.
Dispositions: `docs/specs/logbook/2026-09-18-adr-0019-quarantine-uninstall-export-design-review.md`.

**R1-1 (band A, HEAVY).** Table X row **X1** still selected a recursive `rmSync` of
`state/` from an emptiness snapshot — a dream completing a preserve between the
snapshot and the traversal lost the note, and acceptance criterion 9 inserted its
file *before* the deleter started, so it could not see it. **X1 rewritten:** no
recursive delete ever touches `<state>` or either shelf, no shelf file is ever
deleted, and the shelves and their ancestors come off bottom-up through
non-recursive `fs.rmdirSync` only — `ENOTEMPTY` (or any code but
`ENOENT`/`ENOTDIR`) preserves that directory **and every ancestor** and stops the
climb. The emptiness test and the deletion are now **one syscall**, so there is no
interval. **X2** is snapshot-free (the deleter does not call `quarantineInventory`
at all), **X3** claims the window *closed*, **K7** has no second number to
diverge, and **criterion 9 is an interleaving test** over both live sweeps with a
RED declaration on `quse-carve-out-dropped`, whose mutation now restores this
exact defect.

**R1-2 (band B, reclassified HEAVY — it changes what the user sees).** Table K row
**K2** counted the product-created shelf root directories, so an emptied shelf
refused **forever** and this package's own remedy never terminated. The two roots
now contribute **zero** entries when each is a real directory, with conservative
counting kept for every other entry and for a symlink or file occupying either
root. K3's EMPTY arm, W4's per-root line rule, the worked example and acceptance
criterion 2 (third fixture: both directories present and empty ⇒ proceeds) follow.

**Round-zero LIGHTs applied.** `manifest.js` ranges split correctly (JSDoc
`:1110-1132`, signature `:1133`, loop `:1142-1150`, false sentence `:1112-1116`);
**Table B is 5 anchored + 1 new-code** — the PR body above said 4 + 2 and is
corrected by this comment; the GLOSSARY range overlap with
`WP-quarantine-only-copy-shelf`'s **O8b** (`:141-148` contains `:146-147`) is now
annotated in **W9**, with no content collision.

Not re-opened: owner items 1–3, amend-vs-supersede, refuse-and-report,
code-carve-out-vs-manifest-kind, and Table W row **W7**'s composition rules.

`npm run lint` passes (0 issues, 710 files, 288 specs). `status:` stays `Draft`.
A further round is required on the rewritten X1.

thegyula:
## Design gate round 2 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `85041d69`; fix at **`9e48cde4`**.
Round 1's `rmdirSync` bottom-up sweep and the **K2** shelf-root exclusion **held**.

**R2-1 — the child-by-child sweep followed a symlinked `state`, a regression round
1's own fix introduced.** Table X row **X1** enumerated `paths.state` and
recursively deleted every non-`quarantine` child without first rejecting a
symlink; the shipped `isDir` (`manifest.js:175-181`) uses `statSync` and accepts a
directory symlink, so with `<core>/state` pointing at an external directory of
user files and no `quarantine/`, the gate passes and **that target's children are
deleted** (Astra mocked it — deletion target moves from `state` to
`state/personal-notes`). At `5b77865f` the single recursive `rmSync` removes only
the link, so this was **strictly worse than the tree it replaces**, with a blast
radius wherever the user pointed the link and no cover from the vault guard.

**Fixed by new canonical Table X row X10**, stating the rule once over every path
this package classifies: **every classification is `fs.lstatSync`-based;
`statSync`/`isDir` is never used to decide whether to enumerate or delete.**
**X1 step 0:** a symlinked `paths.state` is `unlinkSync`'d and **never
descended** — no enumeration, nothing below it touched. **X1 step 3:** a shelf
root that `lstat` reports as anything but a directory is **preserved untouched,
reported, never followed, never removed, never climbed past** — consistent with
**K2**, which already counts that object as one of the user's entries.

**One narrowing, stated rather than silently taken.** The recommendation was to
*preserve* the symlink at `state`. **Never-descend is accepted in full**;
preserving the **link itself** is narrowed to **unlinking** it, because that
destroys **no bytes**, reproduces `5b77865f` exactly, and is already how
`disposeCoreMechanics` treats a symlinked **core** (`:1164-1165`) — preserving it
would leave `<core>` permanently non-empty for no safety gain. The shelf roots
keep the recommendation in full.

Also: new Table Y row **Y9**; new acceptance criterion **14** (external
directory's files byte-unchanged, link removed, core emptied, plus both shelf-root
arms); new RED proof **`quse-symlinked-state-descended`** anchored on
`if (!isDir(dir)) continue;` — **1** occurrence at `5b77865f` — whose mutation
reinstates `isDir` and restores this exact defect; the RED-proof criterion moved
**14 → 15** with both of its references. **`isDir` is neither changed nor
removed** (four call sites at `5b77865f`: `:587`, `:616`, `:903`, `:1143`, all
retained; `logs`/`schedules`/`secrets` keep the existing `isDir` + recursive
`rmSync` path and are out of scope).

`npm run lint` passes (0 issues, 288 specs). `status:` stays `Draft`.
A further round is required on X1-with-step-0 and X10.

thegyula:
## Design gate round 3 (Astra) — one finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `a0693065`; fix at **`6fcd68b2`**.
**X10 and X1 step 0 held** and were not re-opened.

**R3-1 (band B, weighted HEAVY — it changes the deletion order the implementer
builds).** `lstat` classifies only a path's **final** component, so X1 step 3's
per-level check applied in the **removal** order reached
`<state>/quarantine/redacted` **through** a symlinked `quarantine`: with
`quarantine` pointing at an external directory holding an empty `redacted/`, the
sweep removed `external/redacted` **before** it detected and preserved the link.
Astra verified the semantics read-only and reproduced the removal by executing X1
in memory. The gate does not cover it — it refuses a *pre-existing* symlinked root,
but `disposeCoreMechanics` is exported and directly callable, and a link introduced
*after* the gate is X3's concurrent class. Criterion 14 asserted surviving **file
bytes**, so it missed the deletion of an **empty** external directory.

**New canonical Table X row X11: validation runs top-down, removal runs bottom-up
— two passes in opposite directions.** X1 step 3 now `lstat`s `<state>`, then
`quarantine`, then `redacted`, **stopping at the first level that is not a real
directory** — that level and everything below it is not accessed, not `lstat`ed
further and not removed — and only then `rmdirSync`s the **validated prefix** leaf
to root. X11 records why the passes cannot be merged: the safety property is about
**ancestors** (downward) while `rmdir`'s emptiness requirement is about
**descendants** (upward), one order serves one of them — the same shape as
`WP-scheduler-replay-manifest-independent`'s Table D row **D5**.

Criterion **14** gains the direct-sweep regression, asserted with
`existsSync`/`lstatSync` and **explicitly not by file bytes**, because an empty
directory has none — the exact gap named. Second RED proof
**`quse-shelf-validated-leaf-first`** reverses X11's validation direction; criterion
14 now carries **two** declarations, since round 2's defect and round 3's are
different failures of the same rule. Totals: **eight declarations over seven
criteria, six anchored + two new-code.**

`npm run lint` passes (0 issues, 288 specs). `status:` stays `Draft`.
One confirming round remains; closes on clean or LIGHT.

thegyula:
## Design gate round 4 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `084142ed`; fix at **`e4962eb6`**,
mirror follow-through at **`9a8dd2ea`**. **X11 held** and was not re-opened.

**R4-1 — the sweep did not recognize its own shelf on a case-insensitive
filesystem.** X1 step 2 recursively deleted every child **not named exactly**
`quarantine`. On APFS/NTFS — two of the three supported platforms — a stored
**`Quarantine`** *is* the destination of `quarantinePreserve`'s lowercase path
while enumeration returns the capitalized name, so byte-equality fails to
recognize our shelf: step 2 deletes it, and with it any preserve completed between
the gate and the sweep, **before** step 3's protected `rmdir` sequence runs. **X10
and X11 do not cover it** — they classify *types* and *ancestors*, not *names*.
Verified on this workspace and reproduced in memory.

**New canonical Table X row X12: shelf identity is decided by ASCII case fold, on
every platform, never by byte-equal name.** A child that folds equal but is **not
byte-equal** is **preserved and reported**, never deleted and never assumed to be
ours — that only ever deletes *less* (ADR-0038's direction) and makes `<state>`
non-empty, so step 3's existing `ENOTEMPTY` rule stops the climb with no new
clause. **The same fold locates the roots for the inventory** (K1 rewritten) so the
plan and the act agree; the **byte-exact lowercase path stays the removal target**.
The accepted set is **closed and is our own good**: our names are pure ASCII with
no combining marks, so their ASCII case variants are exactly the set a
case-insensitive volume could collide with. Explicit ASCII fold required, not
`toLowerCase`.

Criterion **15** (three arms), criterion **9**'s case-fold interleaving arm, RED
proof **`quse-shelf-name-case-sensitive`**, RED-proof criterion moved **15 → 16**.
**Table Y row Y4 was rewritten rather than left standing** — its "exactly two
paths, both literal joins" claim stopped being true once a `readdirSync` name is
joined.

**One unregistered mirror found by my own post-edit claim sweep and registered in
the same pass** (ADR-0031): the Security checklist's untrusted-identifier bullet
was still carrying K1/Y4's superseded claim. Corrected, and a **new mirror class —
"Security-checklist bullets that restate a contract" — added to the Mirrored
Surface Checklist**, naming all five bullets and the rows each defers to.

Totals: **nine declarations over eight criteria, six anchored + three new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 5 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 5 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `59064c78`; fix at **`bbfc4a1d`**.
**X12 held.** First finding of this gate to land on the spec's **Implementation
notes** rather than a contract table — which is itself the lesson: a table row gets
swept for mirrors, a prose bullet did not.

**R5-1 — the blanket "`disposeCoreMechanics` must never throw" breaks uninstall
recovery.** Implemented literally it swallows an `EPERM` removing `secrets/`. The
cited guarantee (`manifest.js:1163`) covers **only** the empty-core removal; the
recursive `rmSync` at `:1148` is **not** wrapped and propagates today. After the
disposer returns, `uninstall.js` deletes the manifest and then `config.yaml`
(`:423-462`), so a swallowed failure leaves a **live OAuth credential on disk**
*and* makes the retry refuse with *"no install manifest found"*. The shipped
mid-sweep recovery test relies on propagation, and the note contradicted this
spec's own "non-quarantine disposal unchanged".

**The claim is WITHDRAWN, not softened.** New canonical **Table X row X13**: the
no-throw promise covers **the shelf levels and the pre-existing empty-core step
only**. Caught and reported: X1 step 3's shelf outcomes (X10/X11 `lstat` or
non-directory level, `ENOTEMPTY`/`EEXIST`, X12's fold ambiguity) → `preservedQuarantine`;
plus the existing empty-core `try/catch`, unchanged. **Still throws byte-for-byte as
at `5b77865f`:** enumerating `paths.state`, any step 2 `rmSync` on a non-shelf
child, and the recursive `rmSync` over `logs/`, `schedules/`, `secrets/`.
**The asymmetry is the stated reason:** preserving at a shelf level can only leave
*more of the user's own text* alive, and it is disclosed; swallowing a mechanics
failure leaves a *credential* alive and destroys the retry — ADR-0019's Decision
(`:47-50`) already says why `secrets/` must go.

**K4's deleter arm narrowed in the same pass** to "at a shelf level only" — it was
the row the blanket note leaned on. New **Table Y row Y10**; new **criterion 16**
measuring the **boundary from both sides** (injected `EPERM` on `secrets/` ⇒ throws,
manifest **and** config still present by existence *and* content, retry succeeds
once cleared; plus the `paths.state` arm; and in the same run a shelf-level
`ENOTEMPTY` that does **not** throw); RED-proof criterion moved **16 → 17**; new RED
proof **`quse-mechanics-failure-swallowed`** anchored on `const mechanics = [` (**1**).

**Blanket-phrase re-scan run as instructed:** four remaining occurrences of
`never throw`/`no-throw`/`crash the uninstall`, all correctly scoped. A new
Security-checklist propagation bullet was added and registered in the mirror class
created at round 4.

Totals: **ten declarations over nine criteria, seven anchored + three new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 6 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 6 (Astra) — one finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `7cfecf2a`; fix at **`53624f18`**.
**X13 held.** Band B, weighted **HEAVY** — the user-visible count changes, and it
was a direct self-contradiction between two of this spec's own contracts.

**R6-1 — removal accounting broke the empty-shelf byte-identity.** X1 required
pushing **every** successful `rmdir` onto the public `removed` array. With both
shelf directories present and **empty** that reports `redacted`, `quarantine` **and**
`state` where the baseline reports **`state` only** — a two-item difference
confirmed by mocked execution. `uninstall.js` sums `removed.length` into
*"Removed N item(s)"* (`:471-473`) **and** renders the same array as the `--dry-run`
mechanics plan (`:337-341`), so both change, contradicting **W6** and criterion 7.

**Fix.** The `rmdir` bookkeeping is **internal** — X1 step 3 pushes nothing onto
`removed` — and **Table X row X4** is rewritten as the public contract:
**`paths.state` appears at most once, and only when the whole of `<state>` is
gone** (final `rmdirSync` succeeded, or step 0 unlinked a symlinked `<state>` per
**X10**); **the intermediate `rmdir`s never appear**; and **when any shelf level was
preserved, `paths.state` does not appear at all**, with `preservedQuarantine`
carrying the detail. X4 now records *why* this is a contract rather than a
formatting choice, citing both caller consumers by line.

**Criterion 7 gains the round-6 fixture** — both shelf directories present and
empty — and compares the *"Removed N item(s)"* line **and** the `--dry-run` plan
**in full**, because the granularity is visible in each and in neither alone.
**Criterion 2** now asserts `removed` holds `paths.state` exactly once and neither
shelf path across all three of its fixtures. **W5** gained the empty-shelf clause,
and an Implementation note names the natural-but-wrong shape (push as you climb) so
it is not re-derived.

Declaration count unchanged: **ten over nine criteria, seven anchored + three
new-code** — `quse-block-printed-on-an-empty-shelf` already mutates exactly the
assertion this finding strengthens.

`npm run lint` passes. `status:` stays `Draft`. Round 7 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 7 (Astra) — one finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `86123c72`; fix at **`796f6652`**,
follow-through at `c08f9c0a` / `9a26b1c4`. **X4 held.** Band B, weighted **HEAVY** —
it changes what the implementer builds for two user-facing surfaces, and it was a
third self-contradiction among this spec's own contracts.

**R7-1 — the algorithm could not satisfy the dry-run contract.** X4 reports
`<state>` removed only after a **successful** `rmdir`/`unlink`, while X1/X2/K7 forbid
the disposer any emptiness test — and in dry-run those syscalls cannot execute.
Skipping them leaves an empty install's plan **incomplete**; assuming success
**mis-reports a populated shelf as removable**. Both `uninstall.js:318` and `:354`
affected.

**New canonical Table X row X15: `dryRun: true` is a READ-ONLY PLANNER**, not a
disposal with the writes skipped. It runs **the same `quarantineInventory`** — same
X12 fold, same X11 three-level top-down validation — performs **no mutating
filesystem call at all**, and returns `removed` as **PREDICTED** removals at X4's
unchanged granularity. **The no-emptiness-test rule is exempted with its reason:**
X2/K7 exist because a snapshot followed by a *deletion* has a raceable interval; a
planner performs no deletion, so being wrong costs a line of output, never a byte.
**The live arm is untouched and stays snapshot-free**, and one implementation means
plan and gate disclosure cannot drift apart. **X2, K5, K7, X4 and W5 were all
re-scoped in the same pass** to say "live arm" where they said "the disposer".

**Named residual `R-dry-run-prediction-drift`, with its direction:** the live
`rmdirSync` fails closed on anything non-empty, so a run can only **preserve more
than predicted, never delete something the plan did not name**. Passing the gate's
snapshot into the disposer was weighed and **not** taken — a caller-controlled input
is exactly what X2 prevents, and the coherence is cosmetic on a command that mutates
nothing.

**Criterion 17**: three shelf states on **both** surfaces, plus **zero mutating `fs`
calls counted through the seam** rather than inferred from the tree afterwards
(which passes vacuously when the fixture had nothing to delete). RED criterion moved
**17 → 18**; new proof **`quse-dry-run-plans-by-assuming`** covers the "assume
success" horn — the other horn is already caught by criterion 7 and
`quse-block-printed-on-an-empty-shelf`, and that is stated rather than left implicit.

**One intra-cell drift found by my own post-edit sweep**: K7's body was re-scoped
but its **row title** and closing sentence still read as universal — the exact
failure `spec-authoring.md` names. Fixed and recorded.

Totals: **eleven declarations over ten criteria, seven anchored + four new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 8 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 8 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `648f225a`; fix at **`45fe40cf`**.
**X15 held.**

**R8-1 — the manifest replay is a second deleter that reaches the shelf.** Table Y
row **Y6** promised immunity to forged manifest entries while the spec left
`reverse()` unchanged. Its `file` handler applies prove-before-delete **only when
the entry carries a hash** (`manifest.js:894`), so a **hash-less**
`{kind:'file', path:'<state>/quarantine/<name>'}` passes `validateEntry`, passes
`withinAllowedRoot` (the shelf is inside `paths.core`) and reaches `fs.rmSync`
(`:899`). Shelf empty at the gate + a preserve completing before the replay ⇒
`reverse()` destroys the new copy **before** the protected disposer runs, and the
**manifest bytes never change**, so the byte-compare does not help. Probed in
memory against the shipped `reverse()`.

**New canonical Table X row X16 — the guard's placement, stated exactly:**
immediately **after** the existing **global deferred-member guard** (`:781`) and
still **before kind dispatch**, following that guard's precedent — it blocks every
path-based route for **every kind** at once, and patching the `file` handler alone
would close the measured instance and leave the rest. **Shelf-resident = any of
three, all preserving:** literal path at/under either shelf root; **resolved** path
at/under it (`.`/`..` and symlink aliases); or **the question cannot be answered**.
Roots located by **X12**'s fold, validated by **X11**'s top-down rule, computed once
beside `schedulerOpts` (`:751-758`). **`contains` is deliberately not used** — its
bare boolean fails closed to *not contained*, which here means *delete* (**X7**).
**Only narrows** (ADR-0038 **N**).

**Table X's heading broadened to "the deletion-side contracts"** — a table governing
one of two deleters is a protection with a documented way around it. **Y6 rewritten,
not patched:** *not being in the manifest was never the same as being safe from it.*

**W7 composition, verified rather than asserted.** Lines added: `quarantineRoots`
beside `schedulerOpts` (`:751-758`) and one guard at the `:781` region — nothing
else in `reverse()`. **No D-row moves:** D5a runs before the entry loop, D5b after
it, the guard inside it before dispatch — three disjoint regions. **Targets are
disjoint by construction:** every D5b target sits under a scheduler root;
`<core>/schedules` and `<core>/state` are sibling subtrees; and **even under a
contrived symlink no shelf file can match R2**, because `quarantinePreserve`
date-prefixes every name (`validate.js:958`) while R2 requires
`^wienerdog-[a-z0-9][a-z0-9-]*\.xml$`. **Landing order both ways:** the scheduler
package is in `depends_on` and lands first, so its "`reverse()` untouched" claim is
true when it ships; **if the order is ever reversed, its implementer must re-pin
that claim.**

**Criterion 18** (pre-existing forged hash-less entry + empty initial inventory +
preserve completed before replay ⇒ bytes survive, entry in `skipped`, manifest bytes
unchanged throughout; plus `.`/`..` and symlink-alias shapes and the
unanswerable-resolution case). RED criterion moved **18 → 19**; new proof
**`quse-reverse-shelf-guard-dropped`** anchored on the global deferred-member guard
line (**1**) — **the one mutation in this package that restores shipped `main`
behaviour**, so a suite green under it asserts nothing this package adds.

Totals: **twelve declarations over eleven criteria, eight anchored + four new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 9 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 9 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `4bd3ea58`; fix at **`21b8c636`**.
**X16's placement held** — the finding is about its *coverage*, and it is the same
family one level up.

**R9-1 — the guard covered targets at or below a shelf, not a recursive deletion of
a shelf ANCESTOR.** With `<core>/app` a symlink to `<core>/state`, a forged
`{kind:'vendored-tree', path:'<core>/state'}` satisfies `reverseVendoredTree`'s only
ownership test (`sameResolvedDir(entry.path, appRoot)`, `manifest.js:588`); its
target lies **outside both shelf subtrees**, so X16 permitted it, and the recursive
`rmSync` at `:593` destroys a preserve completed after the empty-shelf gate. Probed
against the shipped `reverse()`.

**X16 is now symmetric, stated once:** *no manifest-driven deletion may cover a
shelf path **from below** or **from above**.* From below — any mutating kind whose
literal or resolved target is **at or under** a shelf root. From above — any kind
**whose reverser deletes recursively** whose target **contains** one. Unanswerable
resolution preserves in both halves.

**New canonical Table V enumerates the recursive deleters from the shipped code**,
every arm read rather than assumed: **`vendored-tree`** (`:593`) and
**`copied-skill`** (`:645`), and nothing else — `file` is non-recursive, `dir`
removes only a *virtually empty* directory, `symlink` unlinks one link,
`managed-block`/`settings-entry` rewrite one file, `scheduler-entry` removes one
file. `copied-skill` is included **although it is far harder to forge**, because
**the rule is over the deletion shape, not over how hard an arm is to reach**.
**`dir` is excluded deliberately, and that reason is load-bearing:** `init` can
record `{kind:'dir', path:'<state>'}`, and guarding it would emit a `skipped` line
on an **ordinary** uninstall, breaking **W6** and criterion 7's byte-identity.

**Table V is closed only while it is re-measured**, so a **new verification step**
re-runs `grep -n 'recursive: true' src/core/manifest.js` and requires exactly the
three known hits (`:593`, `:645`, `:1148` — the last being `disposeCoreMechanics`,
governed by X1). A fourth hit means a kind was added and Table V is stale.

**Criterion 18** gains the from-above arm as a **different entry shape**, asserting
the recursive `fs.rmSync` is **not invoked** — counted through the seam, because a
skip and a successful-but-empty delete both leave the file present — plus the same
shape for `copied-skill` **and a `{kind:'dir'}` negative control**. **Second RED
proof `quse-reverse-guard-from-below-only`** isolates the new half, which
`quse-reverse-shelf-guard-dropped` cannot: it removes both at once, so a suite
fixturing only at-or-below targets stays green on a from-below-only implementation.

Totals: **thirteen declarations over eleven criteria, nine anchored + four new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 10 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 10 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `878c50bf`; fix at **`1fff91ce`**,
mirror follow-through at `a2bc5c7a`. **Table V and the symmetric X16 held.**

**R10-1 — the guard had no safe absent-shelf contract.** X16 preserved an entry
whenever a shelf resolution *failed*, **without exempting `ENOENT`/`ENOTDIR`** — but
`quarantinePreserve` creates the shelves **lazily** (`validate.js:950`), so on an
install that never quarantined anything **absence is the ordinary case**. Probed
against the shipped `reverse()`: **an ordinary app-tree removal became a skip**, and
`uninstall.js` then deleted the manifest (`:424`), leaving the skipped component
**with no retry ledger**. **And the naive repair is also wrong** — omitting absent
roots would leave a shelf created *during* replay unprotected (X3's window, one
deleter over).

**New canonical Table X row X17 — one resolution rule over every path this package
resolves, explicitly modelled on `WP-scheduler-replay-manifest-independent`'s
Table D row D15:**
1. **Lexical** anchors (`<state>/quarantine`, its `redacted`, plus every X12
   fold-equal name found there) are **always retained, present or not**.
2. **Resolved** anchors are derived through the **nearest validated existing
   ancestor** (`<state>`, else `<core>`) with the missing suffix appended
   **lexically** — so a shelf created **after guard initialisation** is still covered.
3. **`ENOENT`/`ENOTDIR` = ABSENCE**: no anchor, nothing reported, **no entry skipped**.
4. **Any other code = UNANSWERABLE**: entry **preserved and reported**.

**(3) and (4) must stay distinguishable in the implementation and in the tests** —
they are the same syscall failing for opposite reasons, and collapsing them is
exactly what this round caught. **X11, K3 and K4 already carried the rule and are
unchanged**; X16 was the site that lacked it.

**Criterion 18** gains three arms: (a) neither shelf exists ⇒ byte-identical output
and **nothing newly skipped**, asserted over the whole `skipped` array *and* the
complete stdout; (b) a shelf created after guard init but before replay is still
protected; (c) non-`ENOENT` preserves while `ENOENT` on the same level preserves
nothing. New RED proof **`quse-absent-shelf-read-as-unanswerable`** — **the mutation
that turns this package's protection into a denial of service on the majority of
installs**.

**Convergence note (in the spec and the logbook):** resolution semantics are now
**uniform across X11, X12 and X16 under X17**, the same consolidation the scheduler
package reached with D15 after rounds 5–7. **A further finding of this family is
closed by pointing at X17 and adding the site to it, not by writing a fourth rule.**

**Two registered mirrors tightened in the same pass** (Y6 and the Security manifest
bullet still said *"cannot be answered"* — not false, but the phrasing that invites
the collapse).

Totals: **fourteen declarations over eleven criteria, nine anchored + five new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 11 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 11 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `fe6ab693`; fix at **`934b69f2`**.
**X17 held.** The finding is **X16's own rule applied to a site this spec had
exempted**.

**R11-1 — the sweep could delete a quarantined original through a sibling alias.**
With `<state>/quarantine` a **symlink to `<state>/cache`**, X1 step 2 recursively
deletes `cache` — a non-shelf sibling *by name* — **before** step 3 preserves the
link; `quarantinePreserve` writes **through** the alias, so the deleted sibling can
hold **the sole original**. Executed in memory: **the original was deleted and the
surviving, dangling link was reported as *preserved*** — a loss reported as a save.
**Top-down `lstat` validation protects a path from being deleted as itself; it says
nothing about the same bytes being deleted through another name.**

**New Table X row X18 adds no rule** — it applies **X16**'s symmetric containment
check, over **X17**'s anchors, to the recursive deletion the spec had been treating
as ordinary mechanics. **Order, which is the whole fix: protected shelf targets →
step 2 → step 3.** Targets are lexical **and** resolved-through-the-nearest-validated-existing-ancestor,
alias-aware; a target that **contains, equals or is contained by** a protected
resolved target is **preserved and reported**; unanswerable preserves, absent does
not. **Extended beyond the measured site**, because it is the same shape: the
mechanics loop's recursive `rmSync` at **`:1148`** over `logs`/`schedules`/`secrets`
is gated too — `<core>/logs` can be a real directory a shelf symlink resolves into.

**Table V amended:** `:1148` is **governed by X16 via X1/X18, not an exemption**,
and step 2's new per-child delete is listed as a fourth entry. The verification step
now requires **every** `grep -n 'recursive: true'` hit to be **accounted for** in
Table V rather than expecting a fixed count — this package adds one. **W6 is
unaffected** and the row says why.

**Criteria 1 and 9** gain alias arms (direct sweep and post-gate interleaving, with
a `<core>/logs` alias target too) and **assert the alias TARGET's survival before
anything about the link**, because the measured outcome was a loss reported as a
save. New RED proof **`quse-step2-deletes-through-alias`** — the only mutation here
that produces a **dangling link listed in `preservedQuarantine`**, the shape a
reviewer reading the output would be least likely to question.

**Convergence note extended:** two single rows now cover the whole deletion
surface — **X17** for resolution semantics (uniform across X11/X12/X16) and **X16**
for containment — and after this round **every recursive delete in this package, the
replay kinds of Table V *and* the disposer's own two sweeps, is one X16 rule over
X17 anchors. No site is exempt, and Table V no longer lists one.**

Totals: **fifteen declarations over eleven criteria, nine anchored + six new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 12 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 12 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `7cbd4a27`; fix at **`ce1ab255`**.
**X18 held.** The finding lands on **step 0 — the last mutating operation still
outside X16.**

**R12-1 — unlinking a symlinked `<state>` erases the evidence the next sweep needs.**
Step 0 unlinked the alias **before** the protected targets were computed: with
`<core>/state` → `<core>/logs` and an original at `logs/quarantine/…`, the alias goes
and the original does not; X17 then derives anchors under the now-**absent**
`<state>`, X18 sees no overlap, and `logs` is recursively deleted. **Computing the
anchors earlier is not sufficient on its own** — `uninstall.js` calls the disposer
**again at `:467`**, when the alias is already gone — so this had to become a
**retention** rule, not an ordering one.

**New Table X row X19.** Order now begins with the targets: **0a** compute protected
shelf targets (before any mutation at all) → **0b** classify `<state>` → **2**
per-child recursive removal → **3** the `rmdirSync` climb. **A `<state>` symlink
whose resolved target overlaps a protected shelf is preserved and reported, never
unlinked**, so the protection stays discoverable across both sweeps and across a
retry. **One covering no shelf is unlinked exactly as at `5b77865f`** — the ordinary
case criterion 14 already pinned, and X19 is what makes those two arms consistent
rather than contradictory; criterion 14 now says so. The *"step 0 is outside X16"*
language is **gone** from X10, X18 and the security checklist.

**Criteria 1 and 14** gain the arm, asserted across **both** live disposer calls.
New RED proof **`quse-state-alias-unlinked`** exists precisely because **a suite
exercising only the first call stays green** — the loss happens on the second — so
the declaration is what forces the two-call fixture to exist.

## Convergence — the deletion surface is CLOSED

Twelve rounds in, the package's mutating operations are **enumerated**, and the
enumeration is the contract: **(1)** step 0b's `unlinkSync`; **(2)** step 2's
per-child recursive `rmSync`; **(3)** step 3's `rmdirSync` climb; **(4)** the
`mechanics` loop at `:1148`; **(5)** the two Table V replay kinds; **(6)** the
pre-existing, unchanged core removal at `:1151-1170`. **Every one of (1)–(5) is
gated by X16's symmetric check over X17's anchors. No site is exempt, and no table
lists one.** Resolution semantics are uniform across **X11/X12/X16/X19 under X17** —
the shape the scheduler package reached with D15. **A further alias finding is closed
by adding the site to that enumeration, never by writing a new rule**, and an
implementer who finds a mutating operation not on the list has found a spec bug.

Totals: **sixteen declarations over eleven criteria, nine anchored + seven new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 13 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 13 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `52f8be19`; fix at **`6ddae885`**.
**X19 held.** This is the first finding the previous round's convergence note
**predicted its own handling for** — *"a further alias finding is closed by adding
the site to that enumeration, never by writing a new rule"* — and that is how it was
closed.

**R13-1 — symlink replay could erase a shelf alias.** Table V exempted the
`symlink` kind on the reasoning that unlinking a link deletes no bytes. With
`<state>` → `<claude>/skills/wienerdog-test` → `<core>/logs` and an original at
`logs/quarantine/…`, a `symlink` entry naming the **intermediate** alias resolves
**above** the shelf, so X16 permitted it; `reverse()` unlinks it, the disposer meets
a **dangling** `<state>`, X17 rebuilds anchors **lexically**, X18 no longer sees
`<core>/logs` overlapping anything, and the recursive delete takes the original.

**Closed by the rule.** The site joins the enumeration as **item (7)** —
`reverse()`'s `symlink` reverser's `unlinkSync` (`manifest.js:314`). **New Table X
row X20** states the unified sentence: ***no operation of this uninstall removes a
link on the resolution chain of a protected shelf*** — the disposer's step 0b by
**X19**, `reverse()`'s `symlink` kind by **X20**, **one rule seen from two sides**.
A `symlink` entry is preserved when **its own target, or any intermediate alias on
the chain from a lexical anchor to its resolved anchor**, overlaps a protected
target.

**The conceptual correction, recorded because it is the reusable part:** Table V's
YES column meant *"deletes bytes recursively"*; the `symlink` kind is governed for a
**different** reason — **it deletes REACHABILITY, not bytes**, and the bytes go one
step later. The column is re-read as *"the pre-dispatch guard applies"*, each row
naming its clause. **The `dir` exemption is kept and is now provable:** its reverser
removes only a virtually-empty directory and can remove **neither a file nor a
link** — and guarding it would break **W6**.

**Criterion 18** gains the resolution-chain arm — **the only arm spanning `reverse()`
*and* both disposer calls *and* a retry** — asserting the **alias's** survival before
the original's, because the alias is what makes the original protectable next pass.
New RED proof **`quse-symlink-entry-breaks-chain`** on the `reverseSymlink` dispatch
line (**1**): **every other declaration here mutates a deleter; this one mutates an
exemption**, and the loss happens **two steps downstream**, so a suite checking the
shelf right after `reverse()` stays green.

**Enumeration still closed, now seven items** — (1) step 0b's unlink, (2) step 2's
per-child `rmSync`, (3) step 3's `rmdirSync` climb, (4) the `:1148` loop, (5) the two
recursive Table V kinds, (6) the pre-existing core removal, **(7) the `symlink`
reverser's unlink**. All of (1)–(5) and (7) gated by X16 over X17's anchors; (1) and
(7) are one sentence. `dir` is the single, provable exemption.

Totals: **seventeen declarations over eleven criteria, ten anchored + seven new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 14 follows; closes on clean or LIGHT.

thegyula:
## Design gate round 14 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL — and the alias family is now BOUNDED

Raw + focus committed **before** adjudication at `e33c9666`; fix at **`89b2d919`**.
**X20 held.** The finding is a **definition gap**, not a new site.

**R14-1.** X18 checked deletion targets against the shelf **anchors** but not
against the **locations** of intermediate links on the chain. With
`<state>/quarantine` → `<state>/cache/link` → `<core>/logs/recovery`, step 2 may
recursively delete `<state>/cache` — it overlaps **neither** anchor — removing the
intermediate link; the **first** sweep still protects `logs` from cached anchors,
the **second** rebuilds anchors through the now-dangling shelf and deletes the
original.

**Fixed ONCE, as a definition — X17 (1a):** the **PROTECTED SET is the closure of
each shelf's resolution chain**, walked component by component with
`lstat`/`readlink` — **each link's own LOCATION**, each intermediate target, and the
final resolved target. **X16, X18, X19 and X20 all test against that set**, so a
recursive deleter preserves any directory that **contains a chain link**, not merely
one overlapping an anchor. **Criterion 1** gains the two-hop arm across **both**
disposer calls **and a retry**, asserting the **intermediate link first**, then
`cache`, then the bytes — the order the failure cascades in. New RED proof
**`quse-protected-set-is-anchors-only`**: **the first call still passes under it**,
which is what forces the two-call fixture to exist.

## Bounding the family — new Table X row X21

Rounds 11–14 each produced **one more alias shape**, and the runbook's rule is that
the loop converges **by freezing surface, not by patience**.

- **In scope:** links at any depth **inside the resolved chain closure as computed
  at protected-set-computation time**.
- **Out of scope — named residual `R-alias-outside-closure`:** any alias not on a
  chain at that moment — concretely, **a link created after the set is computed that
  redirects a chain component**.
- **Why accepted:** the shelves are `0700` under a `0700` core owned by the user
  running the uninstall, so such a link is **the same user's own act on their own
  files** — the boundary ADR-0035 and ADR-0034 already draw. **Closing it needs a
  filesystem transaction no platform offers**, and the resident process that could
  hold a lock is forbidden by **ADR-0004**. Every alternative enumerates the
  adversary's options, which this repo has twice paid for.
- **Overrule path — new owner item 4**, drafted with its cost: refuse the uninstall
  whenever **any** symlink exists under `<core>`. **An ordinary install with a
  symlinked vault or app directory — a supported layout — could then not uninstall
  at all.** Recommendation: **no**.

**Binding on later rounds: a further alias finding inside the residual's definition
is CLOSED BY CITING THE RESIDUAL, not by a fifteenth revision of the deletion
contracts.**

Owner items are now **1, 2, 3 and 4**, all OPEN. Totals: **eighteen declarations over
eleven criteria, ten anchored + eight new-code.** `npm run lint` passes. `status:`
stays `Draft`. Round 15 follows; closes on clean, LIGHT, **or a finding inside the
residual**.

thegyula:
## Design gate round 15 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `69f49bb4`; fix at **`039d0483`**.
**`R-alias-outside-closure` HELD** — round 14's residual bounded the family as
intended and **nothing was reported inside it**. What landed was the **mirror-image**
failure: round 14's single-class closure **over-protected**.

**R15-1.** For the supported layout `<core>` → `/data/wienerdog`, X17 (1a) put that
node in the protected set, and X18 then preserved **every** target beneath it —
`app`, `logs`, **`secrets`** — **with both shelves absent** — while `uninstall.js:424`
deleted the manifest anyway, **stranding credentials with no retry**. The shipped
symlinked-core tests require those removed.

**Fixed as a two-class definition in X17 (1a):**
- **Class (i) — SHELF SUBTREES** (the shelf roots' resolved targets): protect the
  node **and every descendant**; containment **both ways**.
- **Class (ii) — CHAIN NODES** (every link **location** and intermediate target):
  protect the node **and its ANCESTORS**, **not** their unrelated descendants — the
  target must not **equal or contain** the node, with no containment the other way.

So `/data/wienerdog/secrets` is a **descendant** of a class (ii) node and is removed
exactly as today, while a recursive delete of `/data/wienerdog` **itself** is still
refused. **Verified in the row against rounds 11–14's measured cases**, because a
rule that regresses one of them is not a fix.

**The round's second question, answered rather than confirmed.** `:424`'s
unconditional manifest delete after a **preserving** sweep is **not** governed by
W3's refusal — W3 is the **pre-plan gate** and has already passed. New **Table W row
W10**: when the live sweep returns a non-empty `preservedQuarantine`, the run
**stops before `:424`**, leaving the manifest and `config.yaml` in place, in the shape
`:428-432` already ships. **Same lesson as X13, one step later: preserving the user's
text must never cost them the ability to finish.**

**Criterion 2** gains the symlinked-core arms (absent **and** empty shelves; `app`,
`logs`, `secrets` gone; manifest removed **only** after a clean sweep). **New
criterion 19** measures W10's boundary from both sides. Two new RED proofs:
**`quse-chain-node-over-protects`** — **the only declaration in this package that
proves something is still REMOVED** rather than preserved, the direction a protection
package stops testing — and **`quse-manifest-deleted-after-preserve`** (anchor
`fs.rmSync(paths.manifest, { force: true });`, **1**).

The security checklist now carries the counterweight explicitly: **a protection that
strands a credential has failed in the other direction** (ADR-0019 `:47-50`).

Totals: **twenty declarations over thirteen criteria, eleven anchored + nine new-code.**
`npm run lint` passes. `status:` stays `Draft`. Round 16 follows; closes on clean,
LIGHT, or a finding inside the residual.

thegyula:
## Design gate round 16 (Astra) — two band-B findings, both weighted HEAVY, both ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `905d8ac3`; fix at **`c45fd13f`**.
The two-class rule held, and **nothing landed inside `R-alias-outside-closure` for
the second consecutive round.**

**R16-1 — W10 checked only the FIRST sweep.** That sweep can return nothing, a
concurrent dream then recreates the shelf through `quarantinePreserve`'s recursive
`mkdirSync`, `:424` deletes the manifest and config, and the `:467` sweep preserves
the new copy — after which a retry fails with *"no install manifest found"* **even
once the user clears the shelf**. **W10's check now sits immediately before `:424`**
and consults **every** sweep's `preservedQuarantine` **plus a fresh
`quarantineInventory` read**. **The `:467` call is NOT moved, and the row says
precisely why** (the round asked): its only remaining job is removing the
**now-empty core**, which the manifest and config must be gone for (`:463-466`).
What moves is the check, not the call.

**The window is narrowed, not closed — and both remainders are NAMED** in new
**Table W row W11**: `R-post-ledger-preserve` (between the check and the `rmSync`)
and `R-post-uninstall-preserve` (after the last sweep;
`mkdirSync(…,{recursive:true})` recreates the path so the copy **succeeds**).
**Both are leftovers, neither is a loss**, recovery is one `rm -rf`, and both require
a dream **already running** since the scheduler entries are reversed earlier.
**Because recovery is manual, W8 must now name the preserved directory LITERALLY**
rather than count it. Criterion 9 gains the after-the-first-sweep arm; new RED proof
**`quse-w10-checks-first-sweep-only`**, whose point is that **every earlier
interleaving fixture stays green under it** — they all seed *before* the first sweep.

**R16-2 — X17 and X19 contradicted each other.** For `<state>` → `/data/personal`
with **no** `quarantine/`, hypothetical anchors made the alias look like it contained
a shelf, the unlink branch became unreachable, and with W10 the uninstall **stopped
forever with nothing to clear**. **X17 (1a) now distinguishes HYPOTHETICAL anchors**
— class (ii) semantics only, **never** triggering X19's retention — **from EXISTING
chains**. X19 retains a `<state>` alias **only when an existing shelf's chain passes
through it**; an absent-shelf alias is unlinked exactly as at `5b77865f`.

**And a shelf created after that unlink is still protected — established, not
asserted, so it is NOT residual territory:** `fs.mkdirSync(qdir, {recursive:true})`
(`validate.js:953`) recreates `<core>/state/quarantine/` even with `<state>` gone, and
the live disposer holds **no snapshot across calls** (X2), so the next sweep computes
its set afresh and preserves it. **Criterion 14 is reconciled and strengthened:** it
now asserts the command **completes** on that layout — exit 0, link gone, core gone,
manifest gone — and that a second run refuses for the **ordinary** reason.

Totals: **twenty-one declarations over thirteen criteria, twelve anchored + nine
new-code.** `npm run lint` passes. `status:` stays `Draft`. Round 17 follows; closes
on clean, LIGHT, or a finding inside `R-alias-outside-closure`,
`R-post-ledger-preserve` or `R-post-uninstall-preserve`.

thegyula:
## Design gate round 17 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `dd500826`; fix at **`a6874cfa`**.
**W10/W11 held, and nothing landed inside the three named residuals** — the third
consecutive round in which the bound held.

**R17-1 — round 16's fix went one notch too far.** X17 gave absent shelves **class
(ii)** protection, which excludes descendants, while X16 computes its anchors **once
per `reverse()` call**. With a **stable** `<core>` → `/data/wienerdog` and initially
absent shelves, a pre-existing **hash-less** `{kind:'file'}` entry naming the
**physical** path `/data/wienerdog/state/quarantine/2026-09-18-note.md` bypassed the
lexical anchors **and** — once the dream created that file after anchor computation
and before replay — the hypothetical resolved ones too. **No alias is created or
redirected after computation, so `R-alias-outside-closure` did not cover it**: a rule
defect, not residual territory.

**Corrected as a narrowing of round 16's narrowing:** a **hypothetical shelf ROOT
keeps FULL class (i) SUBTREE protection — descendants included — in every deletion
guard (X16/X18/X20)**, and **the ONLY thing conditioned on an EXISTING shelf is
X19's alias-retention decision**. X19 now says so in its own cell, so the distinction
lives in **exactly one row**.

**Both neighbouring findings re-checked in the same pass**, because this is the third
consecutive round to move this definition. **Round 15 stays closed:** `secrets`,
`logs` and `app` neither equal, contain, nor sit inside the hypothetical root, so
they are **still removed**. **Round 16's contradiction stays removed**, because it
lived entirely in X19 — with `<state>` → `/data/personal` and no `quarantine/`, no
*existing* chain passes through the alias, so it is unlinked as at `5b77865f`.
**Acceptance criterion 14 was re-checked explicitly and holds unchanged.**

**Criterion 18(b)** gains the round-17 arm: stable symlinked core, shelves initially
absent, a hash-less entry naming the **PHYSICAL** path, preserve completing **after
`reverse()` initialises its anchors** — the physical path is *required*, because a
`<core>`-relative one is caught by the lexical anchors and would make the arm
vacuous. New RED proof **`quse-hypothetical-root-not-subtree`**: **nothing else in
this package reddens under it**, so the declaration is what forces both fixture
properties to exist.

**On the bound:** rounds 15, 16 and 17 each landed on the **rule** — over-protection,
a self-contradiction, over-narrowing — and none on a residual. What has moved three
times is the hypothetical/existing definition, now confined to one row **precisely so
a fourth move has one place to happen**.

Totals: **twenty-two declarations over thirteen criteria, twelve anchored + ten
new-code.** `npm run lint` passes. `status:` stays `Draft`. Round 18 follows.

thegyula:
## Design gate round 18 (Astra) — one band-A HEAVY + one band-C LIGHT, both ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `b80dbb46`; fix at **`dc4a277e`**.
Nothing inside the three named residuals for the **fourth** consecutive round.

**R18-1 (HEAVY) — absent-shelf alias removal lost protection across retries.** With
a **pre-existing** `<state>` → `<core>/logs` alias and absent shelves, X19 permitted
the unlink; a dream completing `logs/quarantine/note` **after the absence decision
and before the unlink** was preserved on the first run (cached set, W10 kept the
manifest) — but **on the retry** `<state>` was gone, the inventory missed the copy,
the recomputed anchors no longer protected `logs`, and its recursive deletion
destroyed the original. **No alias is created or redirected after computation, so
`R-alias-outside-closure` did not cover it.**

**X19 gains clause (b):** retain a `<state>` alias when **(a)** an existing shelf's
chain passes through it **or (b)** its **resolved target lies inside any directory
this uninstall sweeps**. The **swept set is enumerated**: `<core>/logs`,
`<core>/schedules`, `<core>/secrets`, every child of `<state>` that X1 step 2
removes, every Table V replay target, and **`<core>` itself** — the last included
conservatively, because `reverse()` can delete a manifest-recorded path anywhere
beneath it. **The blanket "never unlink an absent-shelf alias" was refused and the
refusal recorded**: it re-opens round 16's contradiction and criterion 14. **Clause
(b) is sufficient by construction** — nothing this command sweeps lives outside the
swept set, so a late copy through such a name is unreachable on this run or a retry.

**Criterion 9** gains the before-the-unlink arm — **the only arm spanning a full
RETRY**; **criterion 14**'s fixture must now place its target **outside every swept
directory**, because that (not the shelf's absence) is what licenses its unlink. New
RED proof **`quse-alias-into-swept-dir-unlinked`**: **the first run still passes
under it**, so the declaration forces the fixture to re-run the whole command.

**R18-2 (LIGHT) — and it is mine to own.** X17 (1a)'s concluding sentence still
carried round 16's withdrawn *"class (ii) only"* clause, which **I left in place at
round 17** by rewriting the cell around its ending. That is precisely the intra-cell
drift `spec-authoring.md` names, and the remedy it prescribes — **re-read the whole
cell after a rewrite** — is the one I did not perform. Replaced; **existence now
matters to X19 alone**; criterion 18(b) re-verified against the single definition;
whole-file re-scan returns zero. **Two further X19 mirrors** (X10's step-0b clause
and the security bullet) were found by my own sweep and updated in the same commit.

**Process note recorded in the logbook:** this is the second intra-cell drift in this
gate (K7's title was the first, round 7). Both were introduced by my own revision and
are invisible to the Mirrored Surface Checklist by construction; from here the
whole-cell re-read runs on every cell the round touches, not only the changed clause.

Totals: **twenty-three declarations over thirteen criteria, twelve anchored + eleven
new-code.** `npm run lint` passes. `status:` stays `Draft`. Round 19 follows.

thegyula:
## Design gate round 19 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `e6503c6e`; fix at **`6f0b8cf2`**,
sweep follow-through at `46e25779`. Nothing inside the residuals — **fifth**
consecutive round.

**R19-1 — clause (b) tested only one direction.** With `<state>` → `~/.claude`,
absent shelves, and a sparse manifest holding a **hash-less** `{kind:'file'}` entry
for `~/.claude/quarantine/2026-09-18-note.md`, the target is **outside `<core>`** and
an **ANCESTOR** of the replay target, so the alias was unlinked; a copy before the
unlink, a `secrets` `EPERM` retaining the manifest under X13, and a retry then let the
`file` entry delete the original. **No named residual covers it.**

**Closed as a CLOSURE move, not a third enumeration.** Clause (b) now retains a
`<state>` alias whenever its resolved target **OVERLAPS — equals, contains, or is
contained by — ANY of `withinAllowedRoot`'s roots**:
`[paths.core, paths.claudeDir, paths.codexDir, <home>/.local/bin]` (`manifest.js:742`).
**That object is closed where a list is not:** the gate covers **every** mutating
replay kind (`:872-883`) and the disposer's sweeps all lie under `<core>`, so **every
path this command can delete sits under one of those four roots by construction** —
including an individual **file beneath** the alias target, which is exactly what this
round measured. **The round-18 swept-directory enumeration is kept only as the
derivation and is no longer the test**; the spec stops maintaining a list of targets
it does not own.

**Criterion 9** gains the ancestor arm — target outside `<core>` and an ancestor of a
replay target, copy before the unlink, an **injected `secrets` failure** so X13
retains the manifest, then a **full retry** — with all three properties called
load-bearing (no failure ⇒ no retry; no retry ⇒ the first run passes). **Criterion
14**'s fixture must be **disjoint from all four roots**. New RED proof
**`quse-alias-ancestor-of-target-unlinked`** anchored on the shipped `allowedRoots`
line (**1**): **no earlier fixture reaches it**, since they all put the alias target
inside `<core>` where the one-way test already retained.

**The whole-cell re-read adopted at round 18 paid for itself in the round it was
adopted.** Running it on X19 right after this rewrite caught **two stale tail
sentences** left from round 18, both contradicting the new clause (b) **inside the
cell that defines it**; neither was reported by the reviewer. The same sweep then
found a third — the DoD's round-18 *history* paragraph — which now carries a forward
pointer rather than being rewritten. Two registered mirrors and criterion 14's fixture
requirement were updated in the same pass.

Totals: **twenty-four declarations over thirteen criteria, thirteen anchored + eleven
new-code.** `npm run lint` passes. `status:` stays `Draft`. Round 20 follows.

thegyula:
## Design gate round 20 (Astra) — one band-A HEAVY finding, ACCEPTED IN FULL

Raw + focus committed **before** adjudication at `8be6ce17`; fix at **`14706933`**.
X19(b) held; nothing inside the residuals — **sixth** consecutive round. **First
finding since round 10 in a different family:** rounds 11–19 were about *what gets
deleted*, this one about *whether the user can finish afterwards*.

**R20-1.** X16/X17 have `reverse()` **skip** an entry when a shelf resolution is
unanswerable, recording it **only in `skipped`**, while W10 consulted just the
disposer's preservation and a fresh inventory. A **transient `EIO`** during live
replay skips `app` and hook entries; later reads succeed and find the shelves empty;
the manifest is deleted; the components remain and the retry refuses *"no install
manifest found"*. **No alias change, no late file — no named residual covered it.**

**New Table X row X22, shaped on K4 and the scheduler package's D9, with two
deliberately different rules:**
- **SET-LEVEL — abort.** An unanswerable protected set at **`reverse()`'s guard
  initialisation** **aborts the live replay before any mutation**, in the disposer's
  own unreadable-case shape (name the directory and its `code`, nothing removed),
  leaving the manifest and `config.yaml` untouched. **`--dry-run` reports instead.**
- **PER-ENTRY — preserve, report, and PROPAGATE.** One entry whose own path fails to
  resolve stays preserved and reported, and now reaches W10 via a new
  **`shelfGuarded: string[]`**. **W10 consults three inputs.**

**The asymmetry is recorded rather than left implicit:** aborting on the per-entry
case too would make a damaged tree **un-uninstallable** — precisely the failure owner
item 4 is declined for. **X16's scope note was corrected in the same pass**: it had
promised "no return-shape change", which this field breaks; `skipped` is unchanged and
`shelfGuarded` is empty on an ordinary install, so **W6** holds.

**Criterion 19** gains both arms, the replay-guard one requiring the failure to be
**confined to the replay window**. New RED proof
**`quse-shelf-guard-skip-not-propagated`**: **every other W10 fixture in this package
makes the disposer preserve something**, so all of them stay green under it — only a
confined-failure fixture can see it.

**The whole-cell re-read caught two more before the reviewer:** W10's tail still named
**two** inputs inside the cell that had just gained a **third**, and the
security-checklist insertion left a dangling fragment. That is the third and fourth
intra-cell drift caught by process rather than by a round.

Totals: **twenty-five declarations over thirteen criteria, fourteen anchored + eleven
new-code.** `npm run lint` passes. `status:` stays `Draft`. Round 21 follows.

thegyula:
## Design gate round 21 (Astra) — one finding, ACCEPTED IN FULL — plus the sizing answer

Raw + focus committed **before** adjudication at `c32aaea2`; fix at **`131040ca`**.
X22 held; nothing inside the residuals — **seventh** consecutive round.

**R21-1 (band B, weighted HEAVY — the command blocks forever).** W10 assumed every
`shelfGuarded` entry was an artifact still on disk. But **X16 guards literal shelf
paths BEFORE `reverse()`'s already-gone check** (`manifest.js:865-869`) and X17 keeps
hypothetical anchors when the shelves are absent — so after the user cleared
criterion 18's preserved file **without editing the manifest**, that nonexistent path
was still added to `shelfGuarded` and W10 refused on **every** retry. **The prescribed
remedy could not resolve it.**

**The guard is left alone** — hypothetical anchors must keep protecting a late-arriving
file — **and the fix is in the decision**: *W10 blocks on what still exists or cannot
be checked, never on what is confirmed gone.* Each `shelfGuarded` entry is `lstat`ed
**at decision time** and classified with **X17's own vocabulary** — that rule at a
**third site**, not a new one. **The filter is scoped to `shelfGuarded` alone**, and
the row says why the other two inputs need none: `preservedQuarantine` is computed by
this run's sweep, and the fresh read *is* the current state.

**Criterion 19** gains the remedy-terminates arm (stop → clear the file → re-run →
**complete**) plus complementary assertions, so the filter is measured **as a filter**.
New RED proof **`quse-w10-blocks-on-absent-guarded-path`**: **every other W10 fixture
asserts a run STOPS**, so all stay green under a mutation that only blocks more — it is
the single declaration guarding the **refusing-forever** direction, now recorded in the
Security checklist as a **third** failure direction beside destroying the user's text
and stranding a credential. (Round 1's **K2** was the first instance of that same
direction.)

## Sizing — answered in the logbook, as asked

**Measured:** ~**1,950 lines**, 8 deliverables, **six** canonical tables (X alone has
**22** rows), **26** RED declarations, **13 criteria carrying ~45 arms**.
`spec-authoring.md` warns about exactly this shape, and CLAUDE.md sizes a WP at one
session and one PR. **This is no longer one session.**

**Recommendation: split into TWO, when the gate CLOSES — not before.**
- **A — the gate (S/M):** `quarantineInventory`, the pre-plan refusal and disclosure,
  the two doc clauses, the ADR-0019 amendment, owner items 1–3. **Touches no deleter**,
  is **safe alone and strictly better than `main`**, and is precisely the 2026-07-27
  mandate. Residual: X3's window, closed by B.
- **B — the deletion-side contracts (M):** Tables X and V, W7/W8/W10/W11, owner item 4.
  **B cannot be subdivided further without shipping a measured defect** — rounds 11–21
  each found a path the previous round's partial rule left open.

**Cost of splitting:** two gates (B inherits 21 rounds of settled contracts), a
four-long `depends_on` chain, line numbers re-pinned twice, and A's amendment must be
true without B (it is — it says uninstall *stops*). **Cost of not splitting:** one
session lands 8 files, 22 contract rows, 26 declarations and ~45 arms in one PR.
**Why not now:** splitting mid-gate restarts both gates and orphans 21 rounds of
dispositions keyed to row ids in this document. It is a mechanical extraction after a
clean return; done now it is a redesign.

Totals: **twenty-six declarations over thirteen criteria, fifteen anchored + eleven
new-code.** `npm run lint` passes. `status:` stays `Draft`. Round 22 follows.

thegyula:
## Round 22 (Astra) — `approve`, no findings. **THE DESIGN GATE IS CLOSED**, and the split is done

Raw + focus committed **before** adjudication at `4c5fd818`. Split at **`8485ae0a`**.

**Closing tally.** 22 rounds. Rounds 1–21 produced **17 band-A/HEAVY and 4 LIGHT
findings, every one accepted in full — none dispositioned away**. All 22 raws committed
before adjudication: `5084f7c1` · `85041d69` · `a0693065` · `084142ed` · `59064c78` ·
`7cfecf2a` · `86123c72` · `648f225a` · `4bd3ea58` · `878c50bf` · `fe6ab693` · `7cbd4a27` ·
`52f8be19` · `e33c9666` · `69f49bb4` · `905d8ac3` · `dd500826` · `b80dbb46` · `e6503c6e` ·
`8be6ce17` · `c32aaea2` · `4c5fd818`.

## The two packages

| Spec | Status | Size | Lines | Carries |
|---|---|---|---|---|
| `WP-adr-0019-quarantine-uninstall-gate` | **Ready** | **M** | 891 | `quarantineInventory`, the pre-plan refusal, the ADR-0019 amendment, the two doc sentences. Tables **K** (+**K8**), **W1–W7**/**W9**, **Y1–Y5**/**Y7**/**Y8**; owner items **1–3**; **6** RED declarations |
| `WP-uninstall-shelf-deletion-guards` | **Ready** | **M** | 917 | the carve-out in `disposeCoreMechanics`, the shelf guard in `reverse()`. Tables **X** and **V**, **W8**/**W10**/**W11**, **Y6**/**Y9**/**Y10**; owner item **4**; the three residuals; **20** RED declarations. `depends_on` the gate |

The gate **changes no deleter** — safe alone, strictly better than `main`, and exactly
the owner's 2026-07-27 *"C now + B as follow-on"*. The deletion package **cannot be
subdivided further without shipping a measured defect**.

**Row ids are unchanged**, including the gap where **X14** never existed, so all 22
rounds of dispositions still resolve. Verified mechanically: **no row missing, none
duplicated** across the two files (23 rows in A, 27 in B, zero overlap).

**Two allocations deviate from the sketch and are recorded rather than made silently.**
**W7** → the **gate**, because the gate is the first package to touch
`src/cli/uninstall.js` after `WP-scheduler-replay-manifest-independent` and must carry
that constraint where it applies. **W8** → the **deletion guards**, because it renders
`preservedQuarantine`, a field the gate does not introduce; in the gate it would be
unreachable code referencing a field that does not exist. **W5 and W6 are split by half,
not by row**, with the deletion package's halves deferring to the gate's cells.

The original spec is now a **64-line `Superseded` pointer** with the split rationale. The
dispositions logbook gained the round-22 closure, the 22-SHA list, the **row-id map**, the
**criteria map**, the RED-declaration split, and three process lessons — chief among them
the **whole-cell re-read**, which caught **four of this gate's five intra-cell drifts**.

`npm run lint` passes (290 specs). **A review gate is not owner approval:** nothing here
records the owner approving, accepting, ratifying or signing any of this, and the
ADR-0019 amendment carries *"owner signature pending"*. **Owner items 1–4 remain open.**
Not merged.

PR #303, opened 2026-09-18
fix(secret-scan): scan the whole value before the field cap at the three truncate-then-redact sinks (WP-secret-sink-redact-before-truncate)

thegyula:
Spec: docs/specs/WP-secret-sink-redact-before-truncate.md

## Deliverables checklist

### AC1/AC2 — Table A required/old expressions

```
ok: A1 required expression = 1
ok: A2 required expression = 1
ok: A3 required expression = 1
ok: A1 old expression gone = 0
ok: A2 old expression gone = 0
ok: A3 old expression gone = 0
```

### AC3 — converted probes, old names/DEFECT_MSG gone

```
ok: probe present exactly once: sink-probe: alerts — a labelled secret straddling MAX_FIELD_CHARS is redacted in alerts.jsonl = 1
ok: probe passes: sink-probe: alerts — a labelled secret straddling MAX_FIELD_CHARS is redacted in alerts.jsonl = 0
ok: probe present exactly once: sink-probe: run-evidence — a labelled secret straddling the argv cap is redacted in run-evidence.jsonl = 1
ok: probe passes: sink-probe: run-evidence — a labelled secret straddling the argv cap is redacted in run-evidence.jsonl = 0
ok: probe present exactly once: sink-probe: run-evidence — a labelled secret straddling the scalar-field cap is redacted in run-evidence.jsonl = 1
ok: probe passes: sink-probe: run-evidence — a labelled secret straddling the scalar-field cap is redacted in run-evidence.jsonl = 0
ok: failing tests in the two files = 0
ok: old defect probe names gone = 0
ok: DEFECT_MSG removed from both files = 0
ok: WD-SINK-TRUNC-ALERTS kept as a regression id = 1
ok: WD-SINK-TRUNC-RUNEV-ARGV kept as a regression id = 1
ok: WD-SINK-TRUNC-RUNEV-FIELD kept as a regression id = 1
```

### AC4 — declaration files shape/content

```
ok: tests/red-proofs/secret-sink-wiring-probes-alerts.proofs.json removed
ok: tests/red-proofs/secret-sink-wiring-probes-run-evidence.proofs.json removed
ok: tests/red-proofs/secret-sink-redact-before-truncate-alerts.proofs.json — suite tests/unit/alerts.test.js, 1 proof(s)
ok: tests/red-proofs/secret-sink-redact-before-truncate-run-evidence.proofs.json — suite tests/unit/run-evidence.test.js, 2 proof(s)
```

### AC4 — bare unfiltered `npm run red-proofs` (full run, ~106 declared proofs, this WP's slice)

```
106 declared proof(s), 106 selected
...
PROVEN       alerts-truncate-before-redact  (WP-secret-sink-redact-before-truncate criterion AC4)
PROVEN       run-evidence-argv-truncate-before-redact  (WP-secret-sink-redact-before-truncate criterion AC4)
PROVEN       run-evidence-field-truncate-before-redact  (WP-secret-sink-redact-before-truncate criterion AC4)
...
PROVEN       WP-secret-sink-redact-before-truncate criterion AC4 — alerts-truncate-before-redact=PROVEN; run-evidence-argv-truncate-before-redact=PROVEN; run-evidence-field-truncate-before-redact=PROVEN

RUN: PROVEN
```

Every other WP's declared proof in the repo-wide run also reported PROVEN; no `ERROR`/`GATE FAIL`/`not ok` line anywhere in the run's output.

### Hand-measured RED mutations (BEFORE writing the declarations, on a scratch copy, restored afterward)

| Mutation | Targeted suite / pattern | Reddened | Exact assertion error |
|---|---|---|---|
| C1 — A1 inverse (`secret-sink-redact-before-truncate-alerts.proofs.json`) | `tests/unit/alerts.test.js` `--test-name-pattern="straddling MAX_FIELD_CHARS"` | Exactly 1 test, the converted B1 probe | `ERR_ASSERTION`: `WD-SINK-TRUNC-ALERTS: this sink must scan the whole value before it applies the field cap. ...` |
| C2 — A2 inverse (`secret-sink-redact-before-truncate-run-evidence.proofs.json`) | `tests/unit/run-evidence.test.js` `--test-name-pattern="straddling the argv cap"` | Exactly 1 test, the converted B2 probe | `ERR_ASSERTION`: `WD-SINK-TRUNC-RUNEV-ARGV: this sink must scan the whole value before it applies the field cap. ...` |
| C3 — A3 inverse (same file as C2) | `tests/unit/run-evidence.test.js` `--test-name-pattern="straddling the scalar-field cap"` | Exactly 1 test, the converted B3 probe | `ERR_ASSERTION`: `WD-SINK-TRUNC-RUNEV-FIELD: this sink must scan the whole value before it applies the field cap. ...` |

Each mutation was applied by hand to a scratch copy of the file, run in isolation, confirmed to redden only its declared test with an `ERR_ASSERTION` carrying its `signal` substring, then the file was restored and the full two-file suite re-run green (32/32) before moving to the next mutation. The declared `expectRed`/`signal` values were written from these measurements, not predicted.

### AC5/AC6 — `npm test`

```
ℹ tests 2908
ℹ suites 0
ℹ pass 2896
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
```

### AC7 — `npm run lint`

```
--- markdownlint ---
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed
```

### AC8 — permission boundary

```
ok: files outside the permission boundary = 0
```

`node scripts/boundary-check.js docs/specs/WP-secret-sink-redact-before-truncate.md $(git diff --name-only origin/main...HEAD)` → exit 0.

### AC9 — idempotency

N/A per spec: this WP changes three expressions and their tests; it ships no command and writes nothing outside the repo.

### `git status --porcelain tests/golden/`

Empty (no golden fixture changed — spec did not call for one).

### AC5 — both-sides evidence (measured for PR gate round 1)

Pattern used: `--test-name-pattern="is recorded as the oversized marker"` (unique to the AC5 test; a plain `"oversized"` pattern also matches a pre-existing, unrelated compaction test).

| State | How produced | Result |
|---|---|---|
| **absent** | Ran the pattern against `origin/main`'s `tests/unit/alerts.test.js` (no AC5 test exists there) | `TAP version 13` / `1..0` — zero subtests match; the file-level pseudo-test reports `pass 1` but proves nothing (unproven, not green) |
| **compliant** | Ran the pattern against this branch's shipped code | `ok 1 - sink-probe: alerts — a field over the detector scan bound is recorded as the oversized marker in alerts.jsonl`; `tests 1 / pass 1 / fail 0` |
| **violating** | Hand-reverted A1 to its pre-fix form (`redactOnly(String(v == null ? '' : v).slice(0, MAX_FIELD_CHARS))`) on a scratch copy, re-ran the same pattern, then restored | `not ok 1` — `code: 'ERR_ASSERTION'`, `expected: '[wienerdog: oversized content withheld from secret scan]'`, `actual: 'FFFF…F'` (the 2000-char pre-cut value, never reaching the detector's oversized threshold because the pre-fix order caps to 2000 chars before scanning) |

### PR gate round 1 (wd-reviewer) — paperwork fixes, this push

- Added `docs/specs/logbook/2026-09-19-secret-sink-trunc-probe-flips.md`, recording the P2/P4/P6 Table P DEFECT→CORRECT flips and the R1–R3→C1–C3 declaration replacement, per the spec's Out-of-scope instruction not to edit the done spec directly.
- Rewrote the three stale "Cut before scan … keeps exactly `PROBE_HEAD`" comments this PR's own diff falsified (`tests/unit/alerts.test.js`, `tests/unit/run-evidence.test.js` ×2) to describe the shipped scan-before-cut order and the resulting `'F'×1976 + [REDACTED:anthropic-key]` shape; the `Table S row S<n>` provenance lines above each `test(` call are untouched.
- Re-wrapped `src/core/alerts.js`'s `sanitizeAlert` doc comment at ~80 columns (content unchanged); the fix expression (A1) now sits at `:47` instead of `:48`.
- Re-ran the full verification suite after these edits: `npm test` → `2908/2896/0 fail/12 skipped` (unchanged); `npm run lint` → pass (713 files, +1 for the new logbook entry); the spec's verification script → all `ok`; `node scripts/boundary-check.js ...` → exit 0. RED proofs were not re-proven — no change to `alerts.js:47` or `run-evidence.js:64`/`:78` themselves, only to comments and a doc-comment line-wrap.
- New tip: `17070f40` (was `9f5b0b6a`).

## Decisions made

- The two comment sentences the spec flags as now-false were corrected exactly where cited and nowhere else: `src/core/alerts.js`'s `sanitizeAlert` doc comment ("each length-capped and then secret-scrubbed... the cap bounds the scan input, then `redactOnly` guarantees...") now reads "each secret-scrubbed and then length-capped... `redactOnly` scans the whole value, then the cap bounds the stored length"; `src/core/run-evidence.js`'s `sanitizeRecord` doc comment ("bounded + scrubbed") now reads "scrubbed + bounded", reflecting the new order. No other prose changed.
- "Existing capped-field tests may move" (spec, Implementation notes): checked both test files for any length/content assertion pinning the old truncate-then-redact order. The one plain-truncation test in `tests/unit/alerts.test.js` (`'r'.repeat(MAX_FIELD_CHARS + 500)`) contains no secret-shaped content, so `redactOnly` is a no-op on it either way — its length/content assertions were unaffected and needed no re-derivation. No other capped-field assertion in either file depends on the cut/scan order. Simpler option: leave both untouched rather than speculatively touching assertions the swap does not affect.
- AC5's new test reads the sink's public output for the `reason` field and asserts it equals the detector's fixed 56-character oversized marker string, using `ScanLimits.SCAN_MAX_BYTES` (imported, read-only) to size the oversized input rather than hardcoding `262144`. Simpler option chosen over asserting on `findings` or importing/duplicating the marker string as a new constant, since the spec is explicit that `src/core/secret-scan.js` stays untouched and the marker is not exported.
- Table C's `expectRed[].signal` field: the spec's literal instruction ("use `<id>: this sink must scan the whole value before it applies the field cap`") is read as the regression id the test's `LEAK_MSG(id)` call is invoked with (`WD-SINK-TRUNC-ALERTS` / `-RUNEV-ARGV` / `-RUNEV-FIELD`), not the declaration's own kebab-slug `id` field — that is the string that actually appears in the reddened assertion's message, which the Rules paragraph requires. Measured directly off each hand-triggered `ERR_ASSERTION` before writing the declaration; all three matched.
- Branch name: the dispatch message gave `wp/redact-before-truncate`, while the spec's Definition of done 1(d) mandates `wp/secret-sink-redact-before-truncate`. This is the orchestrator's error, not the implementer's; the branch was not renamed, per the coordinator's instruction that this be recorded and accepted rather than fixed here.

- The four `WD-SINK-CHUNK-*` defects (`-BRAIN-STDOUT`, `-BRAIN-STDERR`, `-RUNJOB-STDOUT`, `-RUNJOB-STDERR`) are untouched and remain fully open at this PR's merge; their probes stay green-because-broken. `WP-secret-sink-chunk-fix` is the successor package. This is also recorded in the new logbook entry, which additionally notes that `WP-secret-sink-wiring-probes`'s Table P rows for P2, P4, P6 are now stale (flip DEFECT→CORRECT) and are left uncorrected in that done spec, per its own Out-of-scope instruction — `WP-secret-sink-chunk-fix` corrects Table P itself once all seven rows have moved.
- No spec-prose was falsified by a measured value; the spec's own 2000-character / 24-byte-leak / "occurs exactly once" claims all matched on re-derivation, and the red-proofs run reddened exactly the declared set with no scope creep.
- `PROBE_TAIL` (declared, unused) survives in both test files, as the spec's Current-state section already notes; not touched, per Deliverables scope.
- `WP-secret-stream-safe-cut-redactor`, running in parallel in a sibling worktree, touches `src/core/secret-scan.js`; this WP's Deliverables never required touching that file and none of it was touched here.
- The round-1 doc-comment re-wrap moved Table A site A1's fix expression from `src/core/alerts.js:48` to `:47` (one fewer line above it). No other Table A site line number changed. Flagging it here since the done-flip logbook entry and this PR's own history both cite `:48` for the pre-round-1 state.

## Lessons

- WP-secret-sink-redact-before-truncate: the `LEAK_MSG`/`expectRed.signal` indirection is worth flagging for future ADR-0042 conversions — the spec's literal `<id>: ...` template describes the string the *test's own call site* produces (the regression id passed into the shared message-builder), not the declaration's kebab-slug `id` field; the two are easy to conflate since both are called "id" in adjacent tables. Measuring the mutation by hand before writing the declaration is what caught this — the assertion's actual printed message is the only reliable source of `signal`, never an assumption from the id column.
- WP-secret-sink-redact-before-truncate: worktree dispatch-time re-verification (`git diff --stat <pin>..origin/main -- src tests skills templates`) is cheap insurance — `origin/main` had already moved past the dispatch note's own snapshot (one more doc-only merge, `423a0f9e`→`c357b281`) by the time the worktree was created, and re-confirming it touched none of the eight Deliverables files (and that the two extra `tests/red-proofs/*.proofs.json` files bringing the count from 24 to 26 were exactly the ones the dispatch note named) avoided building on a stale citation.
- WP-secret-sink-redact-before-truncate: a code-change spec's own worked examples (test comments explaining a fixture's shape) are load-bearing prose too, not just the doc comments the spec names explicitly — swapping an order-dependent expression falsifies any comment describing the *old* order at every call site that constructs a fixture around it, not only the two doc comments the spec pointed at. Worth a grep for the mutation's own keywords ("cut before scan", "truncate", the old variable-order language) across the touched test files before opening the PR, rather than relying on the reviewer to find them.

---
Generated-by: Claude Sonnet 5


thegyula:
## Gate round 2 on tip `17070f40` — fidelity gate clean, independent gate raised one finding, adjudicated below; CI 7/7; merging

**wd-reviewer (spec fidelity), round 2: APPROVE.** Delta `9f5b0b6a..17070f40` mechanically confirmed as comments + the logbook file only (`git diff -U0 -- src/ tests/` minus comment lines is empty). Fresh detached worktree: `npm test` 2908 / 2896 pass / 0 fail / 12 skipped, `npm run lint` 0, boundary-check 0, spec verification script `ALL GATES PASSED`; the three Table C `find` anchors still occur exactly once each. Round-1 findings 1–5 closed.

**Codex plugin `review`, `gpt-6-astra`, round 2 (orchestrator-run, detached worktree, porcelain identical), verbatim:**

> The change fixes the labelled-token truncation leaks but introduces a reproducible regression in entropy-based redaction. Previously withheld token bytes can now reach durable records.
>
> - [P2] Preserve entropy-based redaction after applying the cap — `src/core/run-evidence.js:64`. When an entropy-detected token ends near the cap and is followed by low-entropy alphanumeric padding, scanning the whole value dilutes its entropy and now persists a token previously redacted. For example, `'.'.repeat(1968) + '0123456789abcdef0123456789abcdef' + 'F'.repeat(5000)` writes the complete 32-character token through `recordRunEvidence`, whereas the previous implementation redacted it. The same regression affects both scalar scrubbers. Keep the whole-value scan to fix straddling credentials, but also sanitize the capped result while preserving the storage bound, and add a regression covering this entropy-dilution case.

**Adjudication — reproduced, then bounded with controls (both trees, `recordRunEvidence` through the public entry point, artifact read from disk):**

| value | main `06ccdb95` | this tip |
|---|---|---|
| `'.'×1968 + TOKEN + 'F'×5000` (Astra's case) | redacted | **leaks** |
| `'.'×1968 + TOKEN` (ends at the cap, no padding) | redacted | redacted |
| `TOKEN + 'F'×5000` (same padding, no cap involved) | **leaks** | **leaks** |
| `TOKEN + 'F'×100` (132 chars, far under the cap) | **leaks** | **leaks** |

The finding is real and exactly as narrow as the controls show. A 32-hex token glued to same-alphabet padding is **not a candidate by the scanner's own definition** — `entropyPass` measures Shannon bits over the maximal `[A-Za-z0-9+=/]` run, and the padding dilutes it below 3.5 bits/char — so it leaks on `main` today at every position except one: when the 2000-character cap happens to fall inside the padding, the pre-fix order severed the padding and the truncated run passed the entropy floor by accident. That accidental, cut-position-dependent redaction is what the tip loses; it was never a guarantee, and the same token 100 bytes earlier was never protected.

**Disposition: not taken in this WP.** (1) The proposed remedy — a second `redactOnly` over the capped result — contradicts the frozen contract (Table A's three byte-exact expressions, "the same move … nothing else changes", and verification step A2/A3 greps for those exact expressions); it would also only restore the cap-position accident, not the class (`TOKEN + 'F'×100` stays a leak), and a second pass can lengthen a field past the bound (a short `key=value` rewritten to `key=[REDACTED:generic-secret]`), which then needs a third cut. (2) The defect belongs to the scanner's entropy rule, not to sink order: **entropy dilution by adjacent same-alphabet padding** is pre-existing on `main`. Recorded as **accepted residual 3** for the done-flip (with this table as its threat-model row) and named in HANDOVER as a backlog candidate for a scanner-side spec — a run-boundary or windowed-entropy rule — which is the owner's call, not this WP's. Net: three deterministic labelled-token leaks closed against one accidental entropy catch lost.

**Errata for the done-flip** (spec-side, unchanged from round 1 minus closed items): `:89` "24 files" in `tests/red-proofs/` (measured 26); Table A row A1 and the Current-state bullet cite `alerts.js:48` (tip `:47`); plus the residual above. For `WP-secret-sink-chunk-fix`'s one-dated-block sweep of the done wiring-probes spec: its `:485` count sentence (nine CORRECT / seven DEFECT → now twelve / four) and its `:892` verification-script hardcode of R1's old `expectRed` test name are both falsified by this PR's flips and are not in the logbook entry.

CI on `17070f40`: seven checks pass. Branch name `wp/redact-before-truncate` accepted-and-noted as an orchestrator dispatch error (spec DoD says `wp/secret-sink-redact-before-truncate`); template fixed for the remaining dispatches.

PR #305, opened 2026-09-18
test(dream): pin that the redacted shelf holds only-copies and name the cap's loss (WP-quarantine-only-copy-shelf)

thegyula:
Spec: docs/specs/WP-quarantine-only-copy-shelf.md

## Deliverables checklist

Table O settles what the bounded `state/quarantine/redacted/` shelf owes a copy
that is the only surviving one of a note's pre-scrub content: candidate (a) —
the cap deletes it anyway — with the loss made **legible**, not prevented.
**No `src/` change**, which is the package's central decision (rows O5, O6, O7).

| Deliverable | What landed |
|---|---|
| `tests/unit/dream-validate.test.js` | four tests at the end of the AC-14 block: `[OC-1]` (class A, end to end), `[OC-2]`/`[OC-3]` (both class-C2 entrances), `[OC-4]` (class C1's decay into a sole survivor). No new helper; the whole-tree walk is a local `const` in each of the two tests that needs it |
| `tests/red-proofs/quarantine-only-copy-shelf.proofs.json` | one declaration, `prune-keeps-a-copy-of-what-it-evicts`, `expectRed` = two one-element identity paths |
| `docs/runbooks/secret-incident.md` | one clause inside the existing "keeps roughly the 50 most recent copies" sentence (row O8a) |
| `docs/GLOSSARY.md` | one clause beside *"disposable"* in the `secret quarantine` entry (row O8b) |

**Dispatch-time re-verification.** `git diff --stat c05a575b..origin/main -- src
tests skills templates` is **empty**, and `docs/runbooks/secret-incident.md`,
`docs/GLOSSARY.md` and `scripts/red-proofs.js` are unchanged since the pin too,
so every line number the spec cites was re-derived at its pinned location:
`REDACTED_RETENTION_CAP = 50` (`:663`), `quarantinePreserve` (`:936`),
`pruneRedactedOriginals` (`:1176-1201`), the redact arm (`:1411-1416`), the
identity-gated delete (`:1489-1504`). Base: `c357b281`.

**The RED mutation, measured by hand before the declaration was written**
(the spec requires the implementer to re-measure, because the four tests did not
exist when the architect predicted the red set):

| run | result |
|---|---|
| baseline, four tests | 4 pass / 0 fail |
| mutated (`copyFileSync` aside, then the same `rmSync`), four tests | `[OC-1]` RED, `[OC-2]` green, `[OC-3]` green, `[OC-4]` RED |
| mutated, **whole `npm test`** | `tests 2911 / pass 2897 / fail 2` — the two declared reds and nothing else |
| baseline, whole `npm test` | `tests 2911 / pass 2899 / fail 0 / skipped 12` |

Both reds are `ERR_ASSERTION` and both carry their declared signal **in the
failing assertion's message**:

```
AssertionError [ERR_ASSERTION]: [O2-the-prune-destroys-a-sole-surviving-copy] the evicted copy was the only one of those bytes: nothing under the core holds them any more (walked 51 files)
AssertionError [ERR_ASSERTION]: [O2-C1-decayed-copy-is-destroyed] and its bytes exist nowhere under the core: the prune, which performs no identity check at all, destroyed the last copy (walked 51 files)
```

`[OC-2]`/`[OC-3]` stay green for the reason the spec gives: neither seeds the
shelf to the cap, so neither prunes, so the mutated line never executes in them.
**The measured red set equals the declared one**; nothing was widened.

## Verification output

```
# 1 — the four new tests, by name
$ npm test -- --test-name-pattern OC-1   (… and OC-2, OC-3, OC-4 — identical shape)
ℹ tests 117
ℹ suites 0
ℹ pass 117
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1985.282375

# 1b — the same four, run directly over the one file so the names are VISIBLE
$ TMPDIR=<scratch> WIENERDOG_TEST_NO_REAL_SCHEDULER=1 node --test --test-name-pattern "OC-[1-4]" tests/unit/dream-validate.test.js
✔ EP2 retention [OC-1]: the prune destroys a SOLE-SURVIVING copy — a class-A original is evicted and its bytes then exist nowhere under the core (Table O rows O1, O2) (182.538208ms)
✔ EP2 retention [OC-2]: identity UNPROVEN because the buffers DIFFER — the redacted/ copy is kept, recorded, and is the only copy of its version (Table O row O1, class C2) (82.841667ms)
✔ EP2 retention [OC-3]: identity UNPROVEN because the comparison READ THREW — the redacted/ copy is kept and recorded (Table O row O1, class C2) (85.494334ms)
✔ EP2 retention [OC-4]: a C1 duplicate DECAYS into a sole survivor — the announced twin is deleted, and a later prune destroys the last copy (Table O row O1, class C1; row O2) (121.716ms)
ℹ tests 4
ℹ pass 4
ℹ fail 0
ℹ skipped 0

# 2, 3, 5, 6
$ grep -c 'EP2 retention \[OC-1\]: the prune destroys a SOLE-SURVIVING copy' tests/unit/dream-validate.test.js
1

$ node -e "...expectRed/pathLen..."
tests/unit/dream-validate.test.js 1 WP-quarantine-only-copy-shelf expectRed=2 pathLen=1

$ grep -n 'only copy' docs/runbooks/secret-incident.md
59:     that is the only copy of the original, which cannot be recovered once it
$ grep -n 'only surviving copy' docs/GLOSSARY.md
147:  there is usually the only surviving copy of that note's pre-scrub text, so

$ git diff --stat origin/main...HEAD
 docs/GLOSSARY.md                                   |   5 +-
 docs/runbooks/secret-incident.md                   |   6 +-
 docs/specs/WP-quarantine-only-copy-shelf.md        |   2 +-
 .../quarantine-only-copy-shelf.proofs.json         |  30 +++
 tests/unit/dream-validate.test.js                  | 287 +++++++++++++++++++++
 5 files changed, 326 insertions(+), 4 deletions(-)
$ git diff --name-only origin/main...HEAD | grep -c '^src/'
0

# 4 — the RED lane, BARE and unfiltered (npm run red-proofs)
RED proofs — root /Users/gyulafeher/Documents/Claude_Projects/wienerdog/.claude/worktrees/wp-only-copy-shelf
107 declared proof(s), 107 selected
snapshot: 1742 entries under the declared domain (`.git/` and `node_modules/` excluded)
…
PROVEN       prune-keeps-a-copy-of-what-it-evicts  (WP-quarantine-only-copy-shelf criterion 1)
             why: row O2 is that the prune DESTROYS the bytes for classes A and B, not merely that it unlinks a name from redacted/. A prune that copies the file aside before unlinking satisfies every shipped retention assertion — the shelf still returns to the cap, the evicted basename is still gone, and the rmSync deletion still happens, so the AC-14 tests and the Table N row N2 deletion-count spy all stay green — while the secret-bearing bytes remain on disk. [OC-1]'s whole-tree byte-absence walk is the only assertion in the repo that can see it, which is exactly why it exists
…
PROVEN       WP-quarantine-only-copy-shelf criterion 1 — prune-keeps-a-copy-of-what-it-evicts=PROVEN

RUN: PROVEN

# 7 — the whole suite and the lint pipeline
$ npm test
✔ vendor: an attended sync carries containment forward or refuses — no DATA-shaped A7 write moves it (326.612208ms)
ℹ tests 2911
ℹ suites 0
ℹ pass 2899
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 110069.573333
$ npm run lint
Linting: 712 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed
```

## Decisions made

1. **The declaration's `id` is the Exact-contracts one.** The Deliverables table
   still calls it `shelf-copy-gains-a-withheld-twin` (and cites row O9); the
   frozen block under **Exact contracts** calls it
   `prune-keeps-a-copy-of-what-it-evicts`. AC-6 says the file must be
   "byte-for-byte the declaration under **Exact contracts**", so that block won.
2. **`expectRed` has two entries**, per the round-4 Exact-contracts block and
   the prose beside it ("**`expectRed` has exactly TWO entries**"), not the one
   that AC-6 and verification step 3's comment still say. Measured: two reds.
3. **`[OC-4]`'s fixture removes the refused note from the vault before the
   second run.** The gate refuses that note, so it stays in the vault holding the
   pre-scrub bytes; a later redacting run would preserve *those same bytes* again
   and the byte-absence walk could never pass (row O1's dedup caveat, `OC-P5`).
   Deleting it is what `docs/runbooks/secret-incident.md` step 3 tells the owner
   to do anyway ("delete the whole note if that's simpler"), and it is the
   simpler of the two ways to make the fixture sound.
4. **Every content assertion compares buffers and asserts a boolean or a
   basename.** The security checklist forbids putting shelf bytes in an assertion
   message, and `assert.equal` prints both sides on failure — so even the vault's
   scrubbed form is compared with `Buffer.compare(...) === 0` rather than
   `assert.equal(text, REDACT_SCRUBBED)`.
5. **The three `fs`-patching tests count their injection.** `[OC-2]`, `[OC-3]`
   and `[OC-4]` assert `fired >= 1` before anything else, so a patch that stopped
   matching the path would fail loudly instead of silently turning the test into
   a restatement of the default behaviour.
6. **Branch `wp/only-copy-shelf`** as dispatched, rather than the
   `wp/quarantine-only-copy-shelf` the spec's Definition of done names. Both
   satisfy CLAUDE.md's `wp/<slug>`; nothing reads the branch name.

1. **Owner item 2 — the prune's overlapping-run selection defect — is measured,
   untouched and still open.** A lock-stealing overlapping run evicts another
   still-running invocation's **newest** shelf copy iff the pruning run's own
   `created` set has reached the cap (QD-P3/QD-P4: false at 0/10/48/49, true at
   50/51/60), and at 51 and above the eviction does not even reach the cap, so it
   is pure loss while the other run's report names an artifact that is already
   gone. Table O row O9 recommends the guard and parks it, because it amends
   Table N rows N3 and N5 — contract rows of a `Done` package. Recommended home:
   a new `S` package, `WP-ep2-prune-overlap-guard`. Nothing here touches it.
2. **Two spec statements were falsified by the measurement, both about the
   verification machinery rather than the product.** (a) Acceptance criterion 6
   says `expectRed` holds "**exactly one** entry" and names "`[OC-1]` as the sole
   red"; the frozen declaration it points at has two, and `[OC-4]` reddens under
   the same mutation. The prose two paragraphs above it already says "**exactly
   TWO entries**", so this is a stale criterion, not an ambiguity. (b)
   Verification step 3's comment ("holds exactly **ONE** expectRed entry") is
   stale for the same reason — its command prints `expectRed=2 pathLen=1`, which
   is what the contract requires.
3. **Verification step 1 is a weak check as written.** `npm test --
   --test-name-pattern 'OC-1'` forwards the pattern to the per-file `node --test`
   children, but `tests/run.js` reports one result **per file**, and a file with
   no matching test still passes — so all four commands print `tests 117 / pass
   117` whether or not the tests exist. The targeted per-file run that actually
   names them is included above.
4. **`docs/specs/WP-quarantine-only-copy-shelf.md` has an orphaned fragment at
   lines 477-479**: the tail of an Implementation-notes bullet ("fails there,
   before the lane." plus the "Ambiguity →" bullet) sits directly after the two
   doc-clause paragraphs, duplicating the same two bullets that appear correctly
   under **Implementation notes & constraints**. Cosmetic; not touched.
5. **`docs/GLOSSARY.md` and `docs/runbooks/secret-incident.md` are shared with
   `WP-adr-0019-quarantine-uninstall-gate`**, which edits the adjacent
   `wienerdog uninstall` clause in both. This PR leaves that clause byte-unchanged
   in both files; the GLOSSARY hunk reflows the line it starts on, so expect a
   small textual overlap if the two land close together.

## Lessons

- WP-quarantine-only-copy-shelf: `npm test -- --test-name-pattern X` cannot fail
  when the named test is absent — `tests/run.js` counts one "test" per FILE and a
  file with no match still passes. A spec that verifies a new test by name needs
  either the `grep -c` companion (this one has it) or a direct
  `node --test --test-name-pattern` over the single file.
- WP-quarantine-only-copy-shelf: `tests/with-temp-root.js` takes a **script path**
  and spawns `process.execPath [script, ...args]`, so there is no way to pass
  `--test` through it. To run one unit file directly, set `TMPDIR`/`TMP`/`TEMP`
  at the scratch dir by hand and add `WIENERDOG_TEST_NO_REAL_SCHEDULER=1`.
- WP-quarantine-only-copy-shelf: a byte-absence assertion over a fixture is only
  as good as the fixture's uniqueness. The redacted shelf does **not**
  deduplicate, and a *refused* note stays in the vault holding the very bytes you
  are about to prove gone — so any later redacting run in the same test will
  preserve them again. Neutralize the source note, and use a different secret
  token for the run that drives the prune.
- WP-quarantine-only-copy-shelf: `quarantinePreserve` verifies its artifact
  through the descriptor it already holds (`readSync`), not by pathname, so
  patching `fs.readFileSync` for a quarantine path hits **only** the identity
  comparison in the refuse fall-through. That is what makes the two class-C2
  tests a one-line injection with no collateral effect on either preserve.
- WP-quarantine-only-copy-shelf: when a spec's acceptance criteria and its
  "Exact contracts" block disagree after a late design round, the frozen block is
  the contract — but say so in the PR, because the criterion is what a reviewer
  reads first.

---
Generated-by: Claude Opus 5


thegyula:
## Review gates — both run on tip `5152129c`, both clean

Per `docs/runbooks/codex-review.md` ("When it runs", item 2), both gates ran on the
same tip. Neither wrote into the reviewed worktree: `git status --porcelain` there
was empty before and after both runs.

### Gate 1 — Codex PR review (`/codex:review` backend, native rubric), verbatim

```
# Codex Review

Target: branch diff against main

No actionable defects found. The four regression tests, mutation proof, and documentation changes match the specified behavior without changing production code. Syntax and diff checks passed; runtime tests were not run because the environment is read-only.
```

**Disclosure required by the runbook's "what was EXECUTED" rule:** this verdict is a
**reading**, not an execution — Codex states it did not run the suite. The suite was
run by the implementer and independently re-run by gate 2 (below) and by CI.

### Gate 2 — wd-reviewer (spec-fidelity)

**VERDICT: APPROVE.** Reviewed tip `5152129c`, base `c357b281`. Executed in a throwaway
copy, never in the worktree: the four tests by name (4/4), full `npm test`
(`2911 / pass 2899 / fail 0 / skipped 12` — matches the PR body), `npm run lint`
(passed), `tests/unit/red-proofs.test.js` (139/139), `boundary-check.js` (exit 0), plus
byte-comparisons of the declaration and of all four test names against the spec's frozen
blocks (**identical**). It ran the `--wp`-filtered lane (`PROVEN` for this criterion,
overall `RUN: FILTERED` by construction) rather than the bare lane, and says so — the
bare unfiltered `RUN: PROVEN` in the PR body is the implementer's, not independently
reproduced.

**No band-A findings. No band-B findings against the implementation.** Its nine findings:

| # | band | what | disposition |
|---|---|---|---|
| 1 | B | Deliverables cell at spec `:268` names a declaration id (`shelf-copy-gains-a-withheld-twin`) and a Table O row (O9) that the round-4 frozen block falsifies — registered mirror drift | **fix in the spec (wd-architect)**, not in this PR |
| 2 | B | AC-6 and verification steps 2–3 say ONE `expectRed` entry; the frozen block and its prose say TWO. Reviewer agrees with the implementer's reading and calls AC-6 self-defeating as written | **fix in the spec (wd-architect)**, not in this PR |
| 3 | C | row O8a / AC-7 say "**may be** the only copy"; the Exact-contracts instruction (which the shipped runbook clause follows) says "for most files … the **only** copy" | **residual** — shipped text satisfies all three of row O8's prohibitions; wd-architect to reconcile the two wordings |
| 4 | C | orphaned fragment at spec `:476-479` (pre-existing on `main`) | drop / sweep |
| 5 | C | `walkFor` duplicated in `[OC-1]` and `[OC-4]` — **spec-mandated** (no sixth helper; "a local `const` inside the test") | drop |
| 6 | C | `[OC-2]`/`[OC-3]` express "the verdict is `{refuse:true}`" as `withheld === 1` + `kept() === false`, because the fixture adapter exposes no verdict object | drop — faithful translation |
| 7 | C | branch `wp/only-copy-shelf` vs the DoD's `wp/quarantine-only-copy-shelf` | drop |
| 8 | C | a short dangling line in the runbook bullet (markdownlint clean) | drop |
| 9 | C | merge-order overlap with `WP-adr-0019-quarantine-uninstall-gate` in the same two doc files | drop — maintainer sequencing |

Findings 1–4 are **spec-side**; the reviewer recommends routing a **Table O mirror-sync
pass** to wd-architect (the Deliverables proofs-file cell, AC-6, verification steps 2–3,
the O8a wording, and the `:476-479` orphan) rather than five point edits, because this is
the second round of mirror drift on the same contract family. **None of them blocks this
PR**, and nothing is routed back to the implementer.

CI on this tip: 7/7 green (boundary, lint, pr-title, smoke ×2, test ×2).

thegyula:
## Gate round 1 on tip `5152129c` — both gates clean (implementer-run fidelity gate + orchestrator-run independent gate), CI 7/7, merging

**Codex plugin `review`, `gpt-6-astra`, run by the orchestrator (detached worktree, porcelain identical):** *"The added tests, mutation proof, and documentation match the specified retention behavior; no actionable defects were identified."* (A second, implementer-run plugin review on the same tip returned the same reading — posted above.)

**wd-reviewer (spec fidelity, run by the implementer on the same tip, verdict posted above): APPROVE** — suite, lint, boundary re-run; the declaration and all four test names byte-compared against the frozen Exact-contracts block; no band-A/B finding against the implementation. Orchestrator spot-check: the implementer's own unfiltered `npm run red-proofs` log reads `RUN: PROVEN` (107/107), and the copy-then-delete mutation hand-measured reddens exactly `[OC-1]` and `[OC-4]` as declared.

**Adjudication.** No finding against the diff; `src/` untouched as the spec requires. Errata for the done-flip (spec-side, Table O mirror drift — second drift on the same family; recommend one mirror-sync pass rather than point edits): AC-6 and verification step 3 still say `expectRed` has exactly ONE entry / `[OC-1]` as the sole red (the frozen block says two); the Deliverables proofs-file cell carries the wrong id and row; verification step 1 cannot fail (`tests/run.js` counts one test per file); an orphaned fragment at `:476-479`; row O8a "may be the only copy" vs the Exact-contracts instruction "for most files … the only copy" (the shipped runbook clause follows the latter). Owner item 2 (the prune's overlapping-run selection) remains open and untouched. `docs/GLOSSARY.md`/`secret-incident.md` are shared with `WP-adr-0019-quarantine-uninstall-gate`, whose clause is byte-unchanged here.

CI on `5152129c`: seven checks pass.

PR #307, opened 2026-09-18
fix(dream): give the hardened parser one bounded second look at parse-threw quarantines (WP-ledger-retry-parse-threw-on-upgrade)

thegyula:
Spec: docs/specs/WP-ledger-retry-parse-threw-on-upgrade.md

`parse-threw` is the one quarantine reason whose cause is **our own parser**
rather than the file. When the parser is fixed the file has not changed, the
fingerprint still matches, `selectState` still answers `skip-quarantined`, and
the fixed parser never sees the transcript again. This package converts each
such quarantine record into a `deferred` one — **once per install**, gated by a
new optional top-level ledger marker — so the run that converts them is also the
run that reconsiders them.

The record is **converted, never deleted** (Table A rows A1/A9): deleting it puts
the file on `selectState`'s no-record path, which consults `baseline_mtime`
first and answers `skip-processed` for any file at or below it. `selectState`
and `recordQuarantined` are byte-identical to the merge base; no new record
field and no new selection branch exist.

## Deliverables checklist

## Verification output

> **Re-run in full on the amended tip `eedde8ab` after review round 1** (the
> round-1 fixes touched `src/core/dream/ledger.js`, so nothing below is carried
> over from `562f40ea`): `npm test` 2924 tests / **0 fail**, `npm run lint`
> exit 0, the byte-identity gate `SELECTSTATE AND RECORDQUARANTINED UNCHANGED
> OK`, the upgrade gate `PARSE-THREW RETRY OK`, `boundary-check` exit 0,
> `git status --porcelain tests/golden/` empty, and the **UNFILTERED**
> `npm run red-proofs` → `RUN: PROVEN` (186 declarations, zero
> FILTERED/VACUOUS/UNCONTROLLED/FAILED/ERROR, `pointer-derivation-doctor`
> still PROVEN). All five `expectRed` sets were **re-measured** after the code
> change and came back **identical**.

### `npm test` — exit 0
```
ℹ tests 2924
ℹ suites 0
ℹ pass 2912
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 63050.116083
```

### `npm run lint` — exit 0
```
> wienerdog@0.14.0 lint
> node scripts/lint.js

--- markdownlint ---
Linting: 714 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed
```

### `npm run red-proofs` — UNFILTERED, exit 0
```
PROVEN       rpt-deletion-instead-of-deferral  (WP-ledger-retry-parse-threw-on-upgrade criterion 1)
PROVEN       rpt-reason-filter-widened  (WP-ledger-retry-parse-threw-on-upgrade criterion 2)
PROVEN       rpt-gate-never-set  (WP-ledger-retry-parse-threw-on-upgrade criterion 4)
PROVEN       rpt-gate-not-persisted  (WP-ledger-retry-parse-threw-on-upgrade criterion 4)
PROVEN       rpt-reason-dropped-on-conversion  (WP-ledger-retry-parse-threw-on-upgrade criterion 3)
PROVEN       WP-ledger-retry-parse-threw-on-upgrade criterion 1 — rpt-deletion-instead-of-deferral=PROVEN
PROVEN       WP-ledger-retry-parse-threw-on-upgrade criterion 2 — rpt-reason-filter-widened=PROVEN
PROVEN       WP-ledger-retry-parse-threw-on-upgrade criterion 4 — rpt-gate-never-set=PROVEN; rpt-gate-not-persisted=PROVEN
PROVEN       WP-ledger-retry-parse-threw-on-upgrade criterion 3 — rpt-reason-dropped-on-conversion=PROVEN
RUN: PROVEN

186 declarations reported PROVEN; zero FILTERED / VACUOUS / UNCONTROLLED / FAILED / ERROR verdicts.
```

### Table A row A8 — `selectState` and `recordQuarantined` byte-identical to the merge base
```
SELECTSTATE AND RECORDQUARANTINED UNCHANGED OK
```

### The upgrade gate (NEW verification step), observed RED before it was trusted

The gate is a `node -e` script run from the repo root; it seeds a real ledger
file (baseline_mtime.codex = 2000) holding two `parse-threw` quarantines — one
at mtime 1000, i.e. AT OR BELOW the baseline — plus a sibling record for every
reason row A2 protects, round-trips it through `writeLedger`/`readLedger`, and
checks Table A rows A1, A2, A3, A6, A9 and A10 in five cases.

```
$ # (a) on the merge base — case 1 must fail, because no retry exists
PARSE-THREW RETRY GATE FAILED — case 1: this tree has no one-time parse-threw retry
exit=1

$ # (b) with the reason equality widened to match any quarantined record
PARSE-THREW RETRY GATE FAILED — case 2: expected exactly 2 conversions, got 8
exit=1

$ # (c) on the finished change
PARSE-THREW RETRY OK
exit=0
```

<details><summary>the gate script</summary>

```bash
#!/usr/bin/env bash
# The upgrade gate: Table A rows A1, A2, A3, A6, A9 and A10 over a REAL ledger
# file round-trip, including a file at or below the baseline.
node -e '
const fs = require("node:fs"), os = require("node:os"), path = require("node:path");
const L = require("./src/core/dream/ledger");
const fail = (m) => { console.error("PARSE-THREW RETRY GATE FAILED — " + m); process.exit(1); };
const state = fs.mkdtempSync(path.join(os.tmpdir(), "wd-rpt-gate-"));
const BASE = 2000;
const d = (n, mt) => ({ harness: "codex", path: "/tmp/wd-rpt-gate/" + n + ".jsonl", mtimeMs: mt, size: 64, dev: 7, ino: 42 });
const q = (x, reason) => ({ fingerprint: L.fingerprint(x), outcome: "quarantined", reason, updated_at: "2026-09-01T00:00:00.000Z", harness: x.harness });
const below = d("below-baseline", BASE - 1000), above = d("above-baseline", BASE + 1000);
const files = {};
files[L.foldKey(below.path)] = q(below, "parse-threw");
files[L.foldKey(above.path)] = q(above, "parse-threw");
for (const r of ["over-ceiling", "too-many-lines", "read-error", "secret-revert", "secret-revert-exhausted"]) {
  const x = d("sib-" + r, BASE + 1000);
  files[L.foldKey(x.path)] = q(x, r);
}
const nr = d("no-reason", BASE + 1000);
files[L.foldKey(nr.path)] = { fingerprint: L.fingerprint(nr), outcome: "quarantined", updated_at: "2026-09-01T00:00:00.000Z", harness: "codex" };
const done = d("processed", BASE + 1000);
files[L.foldKey(done.path)] = { fingerprint: L.fingerprint(done), outcome: "processed", updated_at: "2026-09-01T00:00:00.000Z", harness: "codex" };
const seed = { version: 1, baseline_mtime: { claude: null, codex: BASE }, files,
  oversizedExtracts: { "/tmp/wd-rpt-gate/m.jsonl": { fingerprint: "1:2:3:4", appVersion: "0.14.0", extractBytes: 9000000 } } };
L.writeLedger(state, seed);
const seedObj = JSON.parse(fs.readFileSync(L.ledgerPath(state), "utf8"));

// CASE 1 (rows A1, A9): both parse-threw quarantines are reconsidered after a
// real writeLedger -> readLedger round-trip, INCLUDING the one at or below the
// baseline — the file a deletion-based retry answers skip-processed for.
if (typeof L.retryParseThrewOnce !== "function") fail("case 1: this tree has no one-time parse-threw retry");
const r1 = L.retryParseThrewOnce(L.readLedger(state));
L.writeLedger(state, r1.ledger);
const after = L.readLedger(state);
if (L.selectState(after, above) !== "select") fail("case 1: the above-baseline parse-threw file was not reconsidered");
if (L.selectState(after, below) !== "select") fail("case 1: the parse-threw file AT OR BELOW baseline_mtime answered " + L.selectState(after, below));

// CASE 2 (row A2): exactly the two parse-threw quarantines converted; every
// sibling record is byte-identical, and the baseline and memo are untouched.
if (r1.converted !== 2) fail("case 2: expected exactly 2 conversions, got " + r1.converted);
for (const k of Object.keys(seedObj.files)) {
  if (k === L.foldKey(below.path) || k === L.foldKey(above.path)) continue;
  if (JSON.stringify(after.files[k]) !== JSON.stringify(seedObj.files[k])) fail("case 2: " + k + " changed but is not a parse-threw quarantine");
}
if (Object.keys(after.files).length !== Object.keys(seedObj.files).length) fail("case 2: a record was added or removed");
if (JSON.stringify(after.baseline_mtime) !== JSON.stringify(seedObj.baseline_mtime)) fail("case 2: baseline_mtime changed");
if (JSON.stringify(after.oversizedExtracts) !== JSON.stringify(seedObj.oversizedExtracts)) fail("case 2: oversizedExtracts changed");

// CASE 3 (rows A1, A10): the converted record is a well-formed deferral and
// spends none of the secret-revert budget.
const rec = after.files[L.foldKey(above.path)];
if (rec.outcome !== "deferred" || rec.reason !== "parse-threw") fail("case 3: the converted record is not a parse-threw deferral");
if (rec.fingerprint !== L.fingerprint(above)) fail("case 3: the fingerprint was not carried through unchanged");
if ("deferrals" in rec) fail("case 3: the converted record invented a deferrals counter");
if (L.secretDeferralCount(after, above) !== 0) fail("case 3: the converted record spends secret-revert deferrals");

// CASE 4 (rows A4, A6): the marker survived the write, so the sweep is one-shot.
if (L.retryParseThrewOnce(after).converted !== 0) fail("case 4: the retry re-ran after a round-trip through the ledger file");

// CASE 5 (row A3): a file that still fails is re-quarantined and stays skipped.
L.writeLedger(state, L.recordQuarantined(after, above, "parse-threw"));
const back = L.readLedger(state);
if (L.selectState(back, above) !== "skip-quarantined") fail("case 5: a file that still fails was not re-quarantined");
if (L.retryParseThrewOnce(back).converted !== 0) fail("case 5: the retry un-quarantined it a second time");

fs.rmSync(state, { recursive: true, force: true });
console.log("PARSE-THREW RETRY OK");
'
```

</details>

## RED mutations — MEASURED, not predicted

Every set below was measured by hand-applying the declared exact-substring
mutation to `src/core/dream/ledger.js` and running
`node --test --test-reporter=tap --test-name-pattern='\[LRP-' tests/unit/ledger.test.js`,
**before** the declarations were written. All four `[LRP-*]` tests run under the
pattern in every case; every red is an `ERR_ASSERTION` carrying its test's
signal in the assertion message.

| Proof id | Mutation | MEASURED reds (of the 4 tests run) |
|---|---|---|
| `rpt-deletion-instead-of-deferral` | the conversion becomes `delete files[key]` | `[LRP-1]`, `[LRP-2]`, `[LRP-3]` |
| `rpt-reason-filter-widened` | the reason equality drops, matching any quarantine | `[LRP-1]`, `[LRP-2]`, `[LRP-4]` |
| `rpt-gate-never-set` | the returned ledger no longer carries the marker | `[LRP-3]`, `[LRP-4]` |
| `rpt-gate-not-persisted` | the marker is dropped from `writeLedger`'s literal | `[LRP-4]` |
| `rpt-reason-dropped-on-conversion` | `reason: 'parse-threw'` is dropped from the written record | `[LRP-3]` |

Test identities: `[LRP-1]` selection incl. at-or-below-baseline · `[LRP-2]`
nothing else un-quarantined · `[LRP-3]` converted-record shape + `secretDeferralCount`
+ re-quarantine · `[LRP-4]` the one-shot gate across a `writeLedger`→`readLedger`
round-trip.

**`occurrences` sweep before committing** — every `tests/red-proofs/*.proofs.json`
declaration whose `file` is `src/core/dream/ledger.js` still finds its `find`
exactly its declared number of times, including the sibling package's
`pointer-derivation-doctor` (`occurrences: 1`) that this package's file edits
could have disarmed:

```
OK    ledger-retry-parse-threw.proofs.json [rpt-deletion-instead-of-deferral] declared 1, found 1
OK    ledger-retry-parse-threw.proofs.json [rpt-reason-filter-widened] declared 1, found 1
OK    ledger-retry-parse-threw.proofs.json [rpt-gate-never-set] declared 1, found 1
OK    ledger-retry-parse-threw.proofs.json [rpt-gate-not-persisted] declared 1, found 1
OK    ledger-retry-parse-threw.proofs.json [rpt-reason-dropped-on-conversion] declared 1, found 1
OK    quarantine-banner-location-doctor.proofs.json [pointer-derivation-doctor] declared 1, found 1
OK    quarantine-banner-location.proofs.json [banner-shelf-claim-restored] declared 1, found 1
OK    quarantine-banner-location.proofs.json [summary-shelf-claim-restored] declared 1, found 1
OK    quarantine-banner-location.proofs.json [pointer-derivation-ledger-suite] declared 1, found 1
```

## Decisions made

1. **The retry's persistence is conditioned on `converted > 0`, exactly like the
   `migrateFromWatermarks` precedent Table A row A5 says to copy.** The first
   shape written here conditioned it on *the sweep having run* (a new ledger
   object), so the gate would also persist on an install with nothing to
   convert. **Measured: that reddened six pre-existing tests** — four in
   `tests/integration/dream.test.js` ("records no ledger outcome", "advances no
   ledger") and two in `tests/unit/dream-pipeline.test.js` (`[L7]`, and "new size
   evidence is the only ledger write before brain failure") — because a run that
   fails must leave the ledger file untouched. Those files are outside the
   Deliverables, and the invariant they pin is real, so the guard was narrowed to
   match the precedent. The gate is not weakened in practice: the marker rides
   the in-memory ledger from the sweep onward, so **every** later `writeLedger`
   in the run persists it too — an install whose first run processes anything at
   all has the gate set from that run. The only residue is an install that never
   writes its ledger at all (nothing is ever processed), where a no-op sweep over
   a `parse-threw`-free ledger repeats at zero I/O cost until there is something
   to convert.
2. **Marker naming.** The spec fixes that the gate is "one new top-level ledger
   key" whose value is "a marker for this retry", not the name. Chose the key
   `parse_threw_retry` (snake, beside `baseline_mtime`) and the value
   `parse-threw:harden-text-values`, which names the parser revision the retry
   belongs to. Both are exported so tests and future readers cannot re-spell them.
3. **The key is optional in the serializer**, mirroring
   `optionalOversizedExtracts`: a ledger written by code that has not run the
   sweep stays byte-identical to what version-1 code wrote. **Corrected in
   review round 1:** an earlier wording here and in the ADR claimed a ledger
   that "never needed the retry" does not gain the key — false, because
   `retryParseThrewOnce` sets the marker unconditionally (including when
   `converted === 0`) and the first later write carries it. A non-string or
   empty value is *no* marker and the sweep runs — the fail-safe direction (one
   extra reconsideration beats a lost session).
4. **`'parse-threw'` stays an inline literal inside `retryParseThrewOnce`**
   rather than being hoisted into a shared constant, because Table A row A8 lists
   `INFORMATIONAL_QUARANTINE_REASONS` among the things this package must not
   change and that is where the literal already lives.
5. **The user-facing line.** Criterion 5 requires the dry run to "report what it
   would convert" but fixes no text. One line, printed only when
   `converted > 0`, count only, no path, no basename — in the two voices the
   surrounding quarantine lines already use.
6. **ADR-0023 gains Amendment 4** — owner item 1's recorded recommendation,
   implemented under the standing process, carrying
   `Status: **ACCEPTED under standing authorization 2026-09-18 — owner signature pending.**`
   Nothing in this PR records the owner approving, accepting, ratifying or
   signing anything. `docs/adr/README.md` is untouched: its `0023` row carries the
   ADR's own status and does not enumerate Amendments 1–3 either.

## Review round 1 — dispositions (tip `eedde8ab`)

Both PR review gates ran on `562f40ea` and agreed on the same two blocking
findings: **ADR-0023 Amendment 4 contained two sentences the code in the same
commit falsifies.** That is precisely the drift owner item 1's overrule-cost
paragraph exists to prevent, so both are FIXED, not waived.

| # | Gate | Finding | Disposition |
|---|------|---------|-------------|
| 1 | both | Amendment 4: *"A ledger that has never needed the retry does not gain the key"* is false — `retryParseThrewOnce` sets the marker unconditionally, so the first later write carries it on every install | **FIXED.** Rewritten to the true claim: the key is optional *in the serializer*, so a ledger written by code that has not run the retry stays byte-identical |
| 2 | both | Amendment 4: *"older code that drops the key causes at most one extra reconsideration"* understates it — each marker-erasing rewrite authorizes another retry | **FIXED.** Now states the bound is per marker, not global |
| 3 | both | Amendment 4: *"The retry never runs on a preview"* is false — the sweep runs in memory on `--dry-run` (that is how the plan is truthful); only the write is suppressed | **FIXED.** Rewritten to say the retry runs in memory and the preview persists neither marker nor conversions |
| 4 | spec-fidelity | ADR-0031: Amendment 4's prose is a NEW, unregistered mirror of Table A rows A1/A2/A4/A5/A6/A9/A10 | **FIXED.** Registered as a bullet in the spec's Mirrored Surface Checklist, in the same commit as findings 1–3 |
| 5 | spec-fidelity | `Status:` date `2026-09-18` vs the logbook section dated 2026-09-19 (LOW) | **NOT CHANGED — and the finding was WITHDRAWN by its author in round 2.** The two dates are different facts: `Amendment 4 (2026-09-19)` is the amendment's date, `standing authorization 2026-09-18` names when the standing authorization was granted. That is the house form (`docs/adr/0025-…:484`, `docs/adr/0043-…:3`, `docs/adr/README.md:51`). Nothing is owed to wd-architect here |
| 6 | spec-fidelity | `src/core/dream/ledger.js`: a `__proto__` ledger key was silently DROPPED — the rebuilt map used assignment on an ordinary object, so the record vanished (LOW, reviewer did not block) | **FIXED anyway.** A vanished record is exactly row A9's hazard, in the one function whose thesis is "converted, never removed". The map is now `Object.create(null)`; measured before/after, and all five `expectRed` sets re-measured UNCHANGED |
| 7 | both | Row A5's `converted > 0` persistence guard | **NO CHANGE — explicitly cleared.** The spec-fidelity gate traced it end to end and found no gap in row A4: a converting run persists the marker and the conversions in the SAME atomic write, and the marker rides `withRecord` so the write that records a fresh `parse-threw` quarantine persists it too. The residual is an idle install paying one `Object.entries` pass — zero I/O, zero re-parse, which is not the cost row A4 names. Copying `migrateFromWatermarks`'s guard is literally what row A5 instructs |
| 8 | spec-fidelity | Criterion 6 says "three" Table B declarations; Table B declares five. Table B's `rpt-deletion-instead-of-deferral` prediction is narrower than the measured set | **SPEC BUGS — routed to wd-architect** as dated errata; already under Discovered issues below. Non-blocking |

The spec-fidelity gate's verdict on the code itself: *"The code is right. Rows
A1, A2, A4, A5, A9 and A10 are implemented faithfully."* Every blocking finding
was in prose this PR introduced.

## Review round 2 — both gates clean on tip `eedde8ab`

Both PR review gates re-ran on the amended tip and **both approve**.

- **Spec-fidelity gate: APPROVE.** It re-ran `npm test` (2924 / 0 fail), `npm run lint`,
  the byte-identity gate, the upgrade gate and `boundary-check` itself rather than
  trusting this body, verified each round-1 finding's disposition, and **measured**
  the `__proto__` fix end to end (`keys kept = ["__proto__", …]`, `converted: 2`,
  marker persisted, second sweep `converted: 0`). On `Object.create(null)` it
  confirmed the one real hazard was absent: both `isPlainObject` implementations
  that gate `files` (`ledger.js:125-127`, `warnings.js:127-129`) are
  prototype-agnostic, no consumer uses `files.hasOwnProperty(…)` or `key in files`,
  and `composeWarnings` / `activeQuarantines` / `quarantineBannerLine` /
  `secretDeferralCount` / `selectState` produce **byte-identical** output against a
  null-prototype vs an ordinary `files`. Closed-Contract Drift Check: no drift.
- **Codex gate: APPROVE.** Both rewritten ADR paragraphs verified TRUE against the
  code, read as whole paragraphs; a consumer-by-consumer table found no downstream
  break from `Object.create(null)`; no new defect in either `42bd4836..eedde8ab` or
  the round-2 delta `562f40ea..eedde8ab`.

**Accepted residuals (both gates marked non-blocking; recorded, not changed):**

1. A future test comparing a post-sweep `ledger.files` against an object literal
   with `assert.deepStrictEqual` would fail on the prototype. The current suite is
   unaffected (2924 / 0). Left as-is rather than re-spinning a verified tip for a
   comment; noted here so the next test author meets it.
2. `retryParseThrewOnce` overwrites a *foreign* value under `parse_threw_retry`
   with its own marker. Bounded and symmetric with the "per marker, not global"
   bound Amendment 4 now states; Table A row A4 defers "a later retry needs a later
   marker and a deliberate decision" to that future decision, so this is a note for
   whoever writes the second retry.
3. **Process recommendation, routed to wd-architect** (the spec-fidelity gate rates
   it above the errata): when a spec carries a conditional ADR-amendment
   deliverable, **pin the amendment's byte-exact text in the spec**, as sibling
   packages do, so the implementer copies a block instead of authoring a new,
   unregistered mirror of the spec's own contract table. Every blocking finding in
   round 1 came from that one gap.

1. **Spec prose falsified by measurement — Table B / criterion 6 disagree on the
   count.** Acceptance criterion 6 reads "reports `RUN: PROVEN` for all **three**
   Table B declarations"; Table B declares **five**. Five were written, measured
   and proven.
2. **Spec prose falsified by measurement — `rpt-deletion-instead-of-deferral`'s
   prediction.** Table B says "only the at-or-below-baseline case reddens" under
   that mutation. That is true *inside* `[LRP-1]` (its above-baseline assertion
   stays green and the at-or-below one fails), which is the property the row
   exists to establish — but `[LRP-2]` and `[LRP-3]` also redden, because they
   observe the record's continued existence and its written shape, both of which
   a deletion destroys. The declaration carries the measured set and its `why`
   was corrected to say so.
3. **Stale line cites in the spec's Current state.** `src/core/dream/ledger.js`
   is byte-identical to the pinned base `08de2bc3`, so these were already off at
   authoring rather than moved by the dependency: `readLedger` is `:140-158` (spec:
   `:139-157`), `writeLedger` `:162-178` (`:160-177`), `selectState` `:235-264`
   (`:235-263`), its baseline check `:261-262` (`:260-261`),
   `secretDeferralCount`'s reason guard `:289` (`:288-289`).
   `recordQuarantined :326-334` is exact. Every construct was re-derived by hand
   and all are where the spec says they are; only the numbers drifted.

## Lessons

- WP-ledger-retry-parse-threw-on-upgrade: a new unconditional `writeLedger` in
  `src/cli/dream.js` is never free — four integration tests and two pipeline
  tests pin "a run that fails writes no ledger". Copy the guard of the call you
  are told to imitate (`if (mig.migrated && !dryRun)`) rather than inventing a
  stronger trigger, and if a stronger one seems necessary, check the existing
  pins first.
- WP-ledger-retry-parse-threw-on-upgrade: a gate that must be durable does not
  need its own write to become durable — a marker placed on the in-memory ledger
  is persisted by every later `writeLedger` in the same run. Ask what already
  writes before adding a write.
- WP-ledger-retry-parse-threw-on-upgrade: a RED mutation's measured set is
  routinely *wider* than the spec's prediction, because a spec predicts against
  the criterion while the runner measures against every test the
  `testNamePattern` selects. Measure first, then write the declaration and
  correct the `why` — the runner rejects any undeclared red in its own body.
- WP-ledger-retry-parse-threw-on-upgrade: `node --test` needs
  `--test-reporter=tap` before its `not ok` lines can be parsed; the default
  reporter's ✖ lines silently yield an empty measured set, which reads exactly
  like "the mutation reddens nothing".
- WP-ledger-retry-parse-threw-on-upgrade: writing an ADR amendment freehand
  makes you the author of a NEW mirror of your own contract table, and both
  review gates caught the same two sentences in it as false against the code in
  the same commit. Sibling packages that pinned the amendment text byte-exactly
  in the spec did not have this failure mode. Check every sentence of an
  amendment against the diff before committing it, and register it as a mirror.
- WP-ledger-retry-parse-threw-on-upgrade: rebuilding a map keyed by strings
  from disk with `const m = {}; m[key] = v` silently DROPS the key
  `__proto__` — the inherited setter runs instead of a property being defined.
  `Object.create(null)` (or `Object.fromEntries`) fixes it in one line; the
  surrounding `withRecord` was already safe because it used a spread with a
  computed key.

---
Generated-by: Claude Opus 5


thegyula:
## Gate round on tip `eedde8ab` — both orchestrator-run gates clean, CI 7/7, merging

(The implementer reports having run both gates itself over two rounds; none of that is posted here, so both gates were re-run independently on the same tip.)

**Codex plugin `review`, `gpt-6-astra` (detached worktree, porcelain identical), verbatim:** *"No actionable regressions found. The retry preserves record identity, leaves unrelated quarantines untouched, and persists its one-shot marker without changing dry-run state. The three focused in-memory tests passed; filesystem-dependent tests were not run in the read-only environment."*

**wd-reviewer (spec fidelity, executed in a detached worktree): APPROVE.** `npm test` 2924 / 2912 pass / 0 fail / 12 skipped; lint 0; boundary-check 0 (7 files = 6 Deliverables rows + spec); `SELECTSTATE AND RECORDQUARANTINED UNCHANGED OK` re-diffed by hand; the PR's upgrade gate reproduced in all three states (base → `no one-time parse-threw retry`; widened reason → `expected exactly 2 conversions, got 8`; head → `PARSE-THREW RETRY OK`); all five RED declarations hand-applied — measured sets identical to the declarations, every red `ERR_ASSERTION` with the signal; 21 sibling declarations over `ledger.js`/`dream.js` still find their `find` at declared count. Brief-specific: the `converted > 0 && !dryRun` guard traced — conversions and marker share one ledger object and one serializer, so no state persists conversions without the marker; a crash before the write re-converts idempotently; a `__proto__` fingerprint key survives read→sweep→write (measured, seeded via raw JSON); Amendment 4 checked sentence by sentence against the code with no false sentence remaining; Status line byte-exact; no owner-approval claim in the diff; version-1 ledgers serialize byte-identically (6 shapes measured).

**Adjudication.** No finding against the diff. Owner item 1 (ADR-0023 Amendment 4) shipped under the standing process — **owner signature pending**, as the Status line says.

**Errata for the done-flip** (spec-side): criterion 6 "three" Table B declarations (five); Table B's `rpt-deletion-instead-of-deferral` "only the at-or-below-baseline case reddens" (measured LRP-1/2/3 — third package in a row where a Table B row predicts per-criterion while the runner measures everything `testNamePattern` selects; author rows as "at least" or per-test); stale `ledger.js` cites at authoring time (`readLedger` 140-158, `writeLedger` 162-178, `selectState` 235-264, baseline `:262`); Deliverables cell for `dream-pipeline.test.js` says "sweep observed through a real run" while criterion 5 is the dry-run criterion; Mirrored-Surface bullet mislabels round 1's false sentence (it was the key-acquisition claim, not byte-compatibility). Mirror-only contract facts to fold into Table A row A4: absent/empty/non-string marker = no marker (retry runs); the bound is per marker, not global. Process: pin ADR amendment text byte-exact in the spec (both round-1 blockers were in freehand amendment prose).

CI on `eedde8ab`: seven checks pass.

PR #308, opened 2026-09-18
feat(uninstall): replay the scheduler from disk, not the ledger (WP-scheduler-replay-manifest-independent)

thegyula:
Spec: docs/specs/WP-scheduler-replay-manifest-independent.md

## Deliverables checklist

## Verification output

```
$ npm test
ℹ tests 2968
ℹ suites 0
ℹ pass 2956
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 65341.83675
(exit 0)

$ npm run lint
--- markdownlint ---
Linting: 713 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed
(exit 0)

# The recognizer is defined once and CALLED from the manifest (Table R row R1).
$ test -f src/scheduler/generators.js && grep -n 'function recognizeScheduleBasename' src/scheduler/generators.js
144:function recognizeScheduleBasename(basename) {

$ test -f src/core/manifest.js && grep -n 'recognizeScheduleBasename' src/core/manifest.js
745:      if (!gen.recognizeScheduleBasename(name)) continue;
1301:    if (gen.recognizeScheduleBasename(path.basename(p)) === 'launchd') {

# withinSchedulerRoot still carries its own loose patterns (Table R row R3).
$ test -f src/core/manifest.js && grep -Fn '/^ai\.wienerdog\..*\.plist$/' src/core/manifest.js
558:  return /^ai\.wienerdog\..*\.plist$/.test(base) || /^wienerdog-.*\.(timer|service|xml)$/.test(base);

# reverse() does not discover on its own (Table D row D4).
$ test -f src/core/manifest.js && grep -n 'discoveredSchedules = \[\]' src/core/manifest.js
951:function reverse(paths, manifest, { dryRun = false, discoveredSchedules = [] } = {}) {

# ADR-0041: the amendment landed (criterion 8).
$ test -f docs/adr/0041-real-scheduler-mutation-is-opt-in.md && grep -n 'R-stripped-manifest-orphan is closed' docs/adr/0041-real-scheduler-mutation-is-opt-in.md
334:## Amendment (2026-09-19) — R-stripped-manifest-orphan is closed

$ test -f docs/adr/0041-real-scheduler-mutation-is-opt-in.md && grep -n 'ACCEPTED under standing authorization 2026-09-18' docs/adr/0041-real-scheduler-mutation-is-opt-in.md
336:Status: **ACCEPTED under standing authorization 2026-09-18 — owner signature pending.**

# ...and its owner signature header did not move or change (criterion 8, ADR-0035).
$ grep -n '^Status: Accepted$' docs/adr/0041-real-scheduler-mutation-is-opt-in.md
3:Status: Accepted
$ grep -n '^OWNER-SIGNED 2026-08-31$' docs/adr/0041-real-scheduler-mutation-is-opt-in.md
4:OWNER-SIGNED 2026-08-31

$ npm run red-proofs          # UNFILTERED (criterion 19)
PROVEN       srm-widened-set-never-passed  (WP-scheduler-replay-manifest-independent criterion 1)
PROVEN       srm-not-disclosed  (WP-scheduler-replay-manifest-independent criterion 3)
PROVEN       srm-recognition-widened  (WP-scheduler-replay-manifest-independent criterion 2)
PROVEN       srm-act-time-recheck-dropped  (WP-scheduler-replay-manifest-independent criterion 3)
PROVEN       srm-disposition-flipped  (WP-scheduler-replay-manifest-independent criterion 4)
PROVEN       srm-unreadable-root-read-as-empty  (WP-scheduler-replay-manifest-independent criterion 9)
PROVEN       srm-vault-exclusion-removed  (WP-scheduler-replay-manifest-independent criterion 11)
PROVEN       srm-unload-moved-after-loop  (WP-scheduler-replay-manifest-independent criterion 12)
PROVEN       srm-record-suppresses-unload  (WP-scheduler-replay-manifest-independent criterion 13)
PROVEN       srm-external-root-discovered  (WP-scheduler-replay-manifest-independent criterion 15)
PROVEN       srm-resolution-failure-read-as-external  (WP-scheduler-replay-manifest-independent criterion 17)
...
PROVEN       WP-scheduler-replay-manifest-independent criterion 1 — srm-widened-set-never-passed=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 3 — srm-not-disclosed=PROVEN; srm-act-time-recheck-dropped=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 2 — srm-recognition-widened=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 4 — srm-disposition-flipped=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 9 — srm-unreadable-root-read-as-empty=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 11 — srm-vault-exclusion-removed=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 12 — srm-unload-moved-after-loop=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 13 — srm-record-suppresses-unload=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 15 — srm-external-root-discovered=PROVEN
PROVEN       WP-scheduler-replay-manifest-independent criterion 17 — srm-resolution-failure-read-as-external=PROVEN

RUN: PROVEN
(no FILTERED, VACUOUS, UNCONTROLLED or FAILED verdict anywhere in the run)

$ node scripts/boundary-check.js docs/specs/WP-scheduler-replay-manifest-independent.md $(git diff --name-only origin/main...HEAD)
(exit 0)

(empty)
```

### Each new check observed in all three states

`spec-authoring.md` requires absent / compliant / violating. Run on a scratch copy
of the three deliverables (`exit=<code> out=[grep -c output]`):

```
===== STATE: COMPLIANT =====
exit=0 hits=1   test -f src/scheduler/generators.js && grep -c 'function recognizeScheduleBasename' …
exit=0 hits=2   test -f src/core/manifest.js && grep -c 'recognizeScheduleBasename' …
exit=0 hits=1   test -f src/core/manifest.js && grep -cF '/^ai\.wienerdog\..*\.plist$/' …
exit=0 hits=1   test -f src/core/manifest.js && grep -c 'discoveredSchedules = \[\]' …
exit=0 hits=1   test -f docs/adr/0041-… && grep -c 'R-stripped-manifest-orphan is closed' …
exit=0 hits=1   test -f docs/adr/0041-… && grep -c 'ACCEPTED under standing authorization 2026-09-18' …
exit=0 hits=1   grep -c '^Status: Accepted$' …
exit=0 hits=1   grep -c '^OWNER-SIGNED 2026-08-31$' …

===== STATE: ABSENT (the three deliverables deleted) =====
exit=1 out=[]   (× 6 — the `test -f` guard reddens before grep is reached)
exit=2 out=[ugrep: warning: docs/adr/0041-…: No such file or directory]   (× 2 — the two
                 header greps are unguarded in the spec, and still exit non-zero)

===== STATE: VIOLATING (recognizer renamed, loose regex tightened, default removed,
                        amendment reverted, header lines reformatted) =====
exit=1 out=[0]  (× 8 — every check reddens)
```

## Review rounds — what changed (architect rulings, design gate round 11 + PR-gate round 2)

Both gates ran on `13eaa703` (round 1) and `085d2da1` (round 2). **Every ruling
below is an architect ruling made on a PR review round. Nothing in this
repository records the owner approving, accepting, ratifying or signing any of
them.** They live in the spec under `### Design gate round 11 (PR #308 review,
2026-09-19) — amendments to Tables D and R`, with `*Amended by round 11 …*`
pointers on cells D3, D6 and D11.

**Round 2 — the two rulings that were themselves the defect, and are amended:**

- **R-C′ (replaces R-C): identity, not containment.** R-C had phase D5b re-derive
  the D14 discovery roots and re-check containment. Astra reproduced the hole: a
  scheduler ROOT replaced by a symlink **into the vault** re-derives as contained,
  because the vault is inside `HOME` — the candidate lstats as a regular file,
  containment passes, and `rmSync(item.path)` deletes the vault's same-named file.
  Re-asserting containment re-asks a question the attacker can move the answer to.
  Each discovery item now carries `real`, its D15-resolved path **at discovery
  time**, and D5b removes only when `lstat` reports a regular file, D15 resolution
  succeeds, and `res.real === item.real`. That one comparison implies
  discovery-time containment *and* discovery-time vault exclusion, and closes
  every swap class — final component, parent, root — without re-deriving anything.
  The `removalRoots`/`resolvedContains` re-check is gone. Tests: root→vault-symlink
  swap (the vault file survives byte-identical, notice printed, unload still
  spawned), parent-directory swap, plus the three round-1 swap tests.
- **R-B′ (replaces R-B *and* R-A): ownership is read off the plan.** R-B excluded
  *every* discovered path from the manifest-derived `remove` lines, including a
  `remove:false` plist that its own `scheduler-entry` reverser deletes. Measured:
  one `scheduler-entry`, plist present → `--dry-run: 0 item(s) would be removed`
  plus a `keep …dream.plist` line, while `--yes` removed it. The plan under-stated
  a deletion on every normal macOS install. Now only the block's own removals
  (`remove === true` under `unload-and-remove`) are excluded, and a `remove:false`
  item is disclosed as either `removed by its own manifest entry (listed above)` —
  when the plan itself reports that path in `removed` — or `keep <path> (another
  manifest entry owns this file)` plus R9's warning for a launchd basename. The
  `ownedByScheduler` flag R-A introduced is **deleted**: reading the fate off the
  plan is what makes a record naming a **symlink alias** warn correctly (the alias
  is unlinked, the plist survives — Astra's P2), with no ownership flag to get
  wrong. Item shape is now `{path, real, remove}`.

**Round 2 fidelity findings, all applied:** D5b's notices are gated on `!dryRun`
(D8 / ruling R-D — a plan pass prints nothing); `fs.rmSync` is wrapped so a
directory swapped in after the `isFile` check cannot throw `ERR_FS_EISDIR` out of
the uninstall; the no-discovery path is asserted to add **zero** `fs` calls by
instrumenting `realpathSync`/`lstatSync`/`readdirSync` and comparing the omitted
and explicit-`[]` call traces byte-for-byte; the disclosure test now runs F5's
four-item corpus (unrecorded orphan, `scheduler-entry`-recorded plist,
`file`-recorded plist, `scheduler-entry` recording a symlink alias) and checks
each plan line against the file's actual fate after `--yes`; `uninstall.js`'s
JSDoc carries the `{path, real, remove}` shape.

**Round 1 rulings that still stand:** R-C′ supersedes R-C's containment clause but
keeps its ordering and its "remove the DISCLOSED path, never the resolved one";
R-D (D8 reads "prints nothing") is unchanged; the linux-CI fix (every argv
assertion injects its platform via `withDerivedPlatform`) and fidelity F3/F4 from
round 1 are unchanged. `test (ubuntu-latest)` has been green since round 2.

## Decisions made

1. **Who prints the disk-derived block.** Table D row D3 specifies the block's exact
   shape (header, `would run:`, `remove`/`keep`, R9 warning, vault line) and says the
   plan may show the same unregister command **twice** — once from the ledger's
   `scheduler-entry` and once from this block. That count only works if
   `uninstall.js` renders the whole block and phase **D5a prints nothing**; a third
   copy would appear if `reverse()` also emitted a `would run:` line in dry-run.
   So D8's "prints and removes nothing" is implemented as *prints nothing, removes
   nothing* for phase D5a. `reverseSchedulerEntry`'s own dry-run print is untouched.
2. **`discoverSchedulesOnDisk`'s `schedules` element type.** The Exact-contracts JSDoc
   says `schedules: string[]`; the worked example immediately below it and Table D row
   D4 both say `{path, remove}` items. Took the two agreeing sources; the shipped
   JSDoc says `Array<{path:string, remove:boolean}>` (see Discovered issues).
3. **Where an unresolvable vault path is reported.** D12/D15 say it is *unreadable*
   and must abort, but `unreadable` is typed `{root, code}`. Reported as
   `{root: <the vault path>, code}` so D9's message still names the directory and its
   code, rather than inventing a second field.
4. **`remove` attribution when a manifest entry's own path cannot be resolved.**
   D11 matches a validated entry against a discovered path "after realpath
   resolution". Ownership is established by lexical equality **or** resolved equality,
   so a record whose path currently fails to resolve still withholds the deletion of
   the file it literally names — the preservation direction.
5. **Phase D5a counts every disclosed item.** `reverse()`'s returned
   `discoveredSchedules` includes items whose `deriveUnloadArgv` yields `null`
   (a `.service`, a uid-less darwin). Criterion 3 requires the returned set to *equal*
   the disclosed set, and R6 already says `null` means nothing is unregistered while
   the disposition still applies.
6. **One shared `discoveryRoots()` helper.** Its vault resolution runs before the
   root pass, so a vault path that IS a scheduler root is reported once, not twice.
   Round 3 note: R-C′ removed its act-time caller, so it now has exactly one
   consumer (discovery) — kept as a named function because D14 is a named rule.
6b. **The three act-time skip notices share one sentence** — "it is no longer the
   file that was shown to you", with the distinguishing detail in parentheses.
   R-C′ makes them one condition, so one sentence is the honest phrasing.
7. **Disposition and R9's warning text live at one site each.** `DISCOVERED_DISPOSITION`
   and `R9_LOGIN_RELOAD_WARNING` are exported from `src/core/manifest.js` so the
   `uninstall.js` disclosure mirrors the cell rather than restating it — reversing
   owner item 1 stays a one-token edit (Table R row R4, "applied at one site").

## RED mutations — measured, not predicted

Every `expectRed` set below was produced by hand-applying the mutation and running
`tests/unit/uninstall.test.js` **before** the declaration was written. Each new test
carries a unique signal token in *every* assertion message, so the runner can tie a
mutation to the assertion that catches it. All eleven are `ERR_ASSERTION`.

| Proof | Tests that reddened (round 3 — re-measured after the R-B′/R-C′ rewrites) |
|---|---|
| `srm-widened-set-never-passed` | AC1, AC2, AC3/AC17, AC4, AC12, AC13, AC10/R-B′ |
| `srm-not-disclosed` | AC3, AC11-disclosure, AC10/R-B′ |
| `srm-recognition-widened` | AC2, AC3/AC17 |
| `srm-act-time-recheck-dropped` | AC3/AC17 |
| `srm-disposition-flipped` | AC1, AC2, AC4, AC10/R-B′ |
| `srm-unreadable-root-read-as-empty` | AC9 |
| `srm-vault-exclusion-removed` | AC11, AC11-disclosure |
| `srm-unload-moved-after-loop` | AC1, AC2, AC3/AC17, AC4, AC12, AC13 |
| `srm-record-suppresses-unload` | AC12, AC13, AC10/R-B′ |
| `srm-external-root-discovered` | AC15/AC16 |
| `srm-resolution-failure-read-as-external` | AC17 |

31 `expectRed` entries, every one `ERR_ASSERTION`, every set measured by
hand-applying the mutation against the round-3 tree before the declaration was
written. Two finds moved again this round: `srm-act-time-recheck-dropped` now
spans R-C′'s three identity checks, and `srm-not-disclosed` follows
`printDiscoveredSchedules`'s new `planRemoved` argument.

1. **Round 9 falsified Table D row D5's "why the split is not cosmetic" as a
   *measurable* claim, and with it Table B's `srm-unload-moved-after-loop` mutation.**
   The sentence *"One pass cannot be both 'after the trusted entries' and 'before they
   destroy the evidence'; two phases can"* was written when the unload had an act-time
   precondition. Round 9's D6 made phase D5a **filesystem-free**: the argv is derived
   from `path.basename(item.path)` alone, so an unload phase placed *after* the entry
   loop derives and spawns exactly the same argv for a file the loop deleted. I
   measured this directly — relocating the phase changes no observable. The ordering
   is still the right shape (it is what makes the phase order *sufficient* rather than
   accidental, and it is what a future act-time precondition would need), but it is no
   longer observable on its own, so the declared mutation removes phase D5a's input
   instead of moving it. Table D row D5 and Table B's "Mutation" cell should be
   reconciled with D6 by the architect.
2. **Exact-contracts JSDoc contradicts the worked example and Table D row D4** on
   `discoverSchedulesOnDisk`'s `schedules` element type (`string[]` vs `{path, remove}`).
   Shipped the latter; the spec's `@returns` line needs the same fix.
3. **The worked example under "Exact contracts" predates round 8.** Its plan block
   shows two launchd `keep` lines with no R9 warning after either, while Table D row D3
   and Table R row R9 make that warning mandatory on every launchd `keep` line. Shipped
   per D3/R9; the example is a registered mirror and is now stale.
4. **"Implementation notes & constraints" points the agreement test at
   `tests/unit/scheduler.test.js`**, which does not exist on this tree. The Deliverables
   table names `tests/unit/scheduler-generators.test.js`, which is where it landed.
5. ~~**A widened removal is disclosed twice in the pre-confirm plan.**~~ **Fixed**
   by ruling R-B, then corrected by **R-B′** in round 2 — only the block's own
   removals are excluded, so a record-owned path keeps its own reverser's line.
5b. **`opts.platform` on `discoverSchedulesOnDisk` is declared and unused** (fidelity
   F6). Left in place: the JSDoc is the spec's byte-exact Exact-contracts block.
   Recorded as an erratum for the done-flip.
6. **`tests/unit/uninstall.test.js` cannot be run with a bare `node --test`.** The
   pre-existing `Table T forgotten-seam determinism` test asserts
   `WIENERDOG_TEST_NO_REAL_SCHEDULER === '1'`, which only `tests/run.js` sets. Not
   touched; noted because it costs every implementer one confused minute.
7. **The new red-proof anchors are darwin-only.** The eleven reddening tests skip off
   darwin (the launchd argv is the executable arm per the spec's Platform-scope table),
   and a skipped test is not a terminal PASS for the runner's BASELINE phase. `npm run
   red-proofs` is not in CI, so this does not break the pipeline, but the lane is
   host-dependent for this declaration file.

## Lessons

- WP-scheduler-replay-manifest-independent: `scripts/red-proofs.js` allows exactly one
  `suite` per declaration file, so a spec that mandates ONE proofs file and declares
  mutations in two source files forces every reddening test into one suite. Worth
  deciding in the spec which suite that is, rather than discovering it at implementation.
- WP-scheduler-replay-manifest-independent: appending helpers to a long existing test
  file silently overwrites same-named ones — a second `function captureStderr` at the
  bottom of `tests/unit/manifest.test.js` hoisted over the first and broke 21 unrelated
  symlink tests. Grep for every helper name you are about to introduce.
- WP-scheduler-replay-manifest-independent: `execFileSync` inherits the child's stderr
  by default, so a test that asserts on an interactive prompt (`confirm` writes to
  stderr) sees neither the prompt nor its ordering. Assert on what lands in stdout.
- WP-scheduler-replay-manifest-independent: give every assertion in a red-proof-anchored
  test the SAME unique signal token. The runner requires the signal in whichever
  assertion the mutation happens to redden, and different mutations redden different
  ones.
- WP-scheduler-replay-manifest-independent: when a design gate's later round changes a
  contract (round 9 making the unload filesystem-free), earlier rounds' RED declarations
  can quietly become unmeasurable. Re-derive each declared mutation against the contract
  as it finally stands, not against the round that proposed it.

---
Generated-by: Claude Opus 5


thegyula:
## Gate round 1 on tip `13eaa703` — independent gate raised two P1s, CI 6/7 (ubuntu `test` fails); fidelity gate still running — back to the implementer

> The widened removal can delete an undisclosed symlink target outside its allowed roots. Two new tests also deterministically fail on the existing Linux CI platform.
>
> - [P1] Revalidate containment before deleting a discovered file — `src/core/manifest.js:1263`. If a discovered schedule file is replaced with a symlink while the confirmation prompt is open, `resolveOrReport()` resolves its new target and this line deletes that target—even outside the scheduler roots or inside the vault. Discovery's earlier regular-file check does not protect this later deletion. Recheck the non-symlink and containment constraints before removal, while keeping the unload unconditional. This violates the root-bounded deletion invariant in Table S1–S2.
> - [P1] Make the new unload assertions portable to Linux — `tests/unit/manifest.test.js:2561-2563`. On Linux, this test supplies only a `.plist` and an `.xml`, so `deriveUnloadArgv()` returns `null` for both and the asserted spawn never happens. The following D15/R9 test likewise expects one spawn from a plist without injecting Darwin derivation. Both therefore fail the existing Ubuntu `npm test` CI job. Use platform-appropriate fixtures or the existing `withDerivedPlatform()` helper for these assertions.

**Adjudication (orchestrator, against the spec as frozen).** Both HEAVY.

1. The act-time removal at `manifest.js:1263` calls `resolveOrReport(item.path)` and then `fs.rmSync(res.real)`. Table D row **D15** says the act-time re-check applies the *same* rule as discovery — "only a SUCCESSFUL resolution may classify a path as external or contained" — and rows **S1/S2** make the roots, not the name, the deletion boundary. The shipped re-check resolves but never *classifies*: it neither re-asserts D1's `lstat` regular-file requirement nor re-asserts containment in this run's discovery roots, and it deletes the **resolved target** rather than the disclosed path. So a symlink planted between disclosure and the act deletes whatever it points at. This is a fidelity gap against D6/D15/S1/S2, not a spec gap: the re-check must (a) `lstat` the disclosed path and require a regular file, (b) resolve it through D15's rule and require containment in one of this run's discovery roots (D14), skipping with a notice — and R9's warning where the basename is launchd — on anything else, and (c) remove the **disclosed path** `item.path`, never `res.real`. Removal stays a subset of the disclosed set (D7); the unload stays unconditional (D6).
2. The two new `manifest.test.js` assertions derive an unload argv from `.plist`/`.xml` fixtures without pinning the platform, so on the ubuntu runner `deriveUnloadArgv` returns `null` and the spawn never happens — CI red. Use the file's existing `withDerivedPlatform()` helper (or platform-appropriate fixtures) so the assertions are platform-independent.

The fidelity reviewer's findings are appended below when they arrive; the implementer has both items now.

thegyula:
## Gate round 1, part 2 — fidelity gate REQUEST-CHANGES; architect rulings for the two canonical contradictions

**wd-reviewer (spec fidelity, executed in a detached worktree under a redirected temp HOME + `XDG_CONFIG_HOME` + `WIENERDOG_LOADER_NOOP`):** `npm test` 2953 / 2941 / 0 fail; lint 0; boundary-check 0 (10 files, all Deliverables); all 8 spec greps hit; ADR-0041 amendment **byte-identical** to the drafted block, `Status: Accepted` / `OWNER-SIGNED 2026-08-31` unmoved at lines 3–4; all 11 RED mutations hand-applied — each reddens exactly its declared set; no owner claim beyond the mandated standing-authorization Status line. Findings: (1) D5b's `fs.rmSync(res.real)` is unguarded — reproduced: a directory at the disclosed path crashes `reverse()` with `ERR_FS_EISDIR` mid-uninstall; an out-of-root symlink swapped in during the prompt deletes its target (same defect as Astra's P1, now measured); (2) the R9 login-reload warning prints for every launchd `keep`, including `scheduler-entry`-owned plists whose own reverser deletes the file — false on every normal macOS install; Table D row D3 ("every `keep` line") and Table R row R9 ("NON-scheduler kind") state incompatible scopes; (3) criterion 10's disclosure half (keep line + warning + vault line) has no test; (4) criterion 1's linux arm is unasserted; (5) D5b's deletions enter `reverse().removed`, so a discovered path is disclosed twice in the plan and the `--dry-run` headline is no longer independent of R4, contradicting D3; (6) `discoverSchedulesOnDisk`'s `opts.platform` is unused. Brief items: the substituted `srm-unload-moved-after-loop` mutation is right (relocating D5a after the loop changes no observable once D6 made it filesystem-free — D5/Table B/S9 must be reconciled with D6 at the done-flip); Decisions 1 and 4 are ambiguity resolutions; nothing in the diff can reach `launchctl` or `~/Library/LaunchAgents` under a correctly redirected HOME; the worked example is the stale mirror (predates R9).

**Architect rulings (design gate round 11, recorded in the spec on this branch under "Design gate round 11 (PR #308 review, 2026-09-19)"):**
- **R-A** D3 narrows, R9 unchanged: discovery items carry `{path, remove, ownedByScheduler}`; the R9 warning follows a `keep` line only when the owning record is not a `scheduler-entry`. D5b's unreadable-skip notice keeps its warning (the file is left standalone by this run).
- **R-B** `reverse().removed` keeps every D5b deletion (post-action record); the plan and `--dry-run` disclose a discovered path exactly once, in the disk block — `uninstall.js` excludes the discovered set from the manifest-derived `remove` lines and the headline count.
- **R-C** act-time re-check: `lstat` regular file, D15 resolution succeeds, resolved path contained in this run's discovery roots; absent = silent skip; anything else = skip with notice (+R9 where launchd); `fs.rmSync(item.path)` on the disclosed path, never the resolved one. Unload unconditional; removal ⊆ disclosed.
- **R-D** D8 wording; worked example rewritten to the shipped output, stating the record kinds it assumes.

Implementer has both messages; round 2 runs both gates on the new tip. Done-flip errata (spec-side, not for this PR): D5/Table B/S9 vs D6; Mirrored-Surface "criterion 10" ×3; DoD 6a "CLOSED … open"; `scheduler.test.js` cite; `opts.platform` unused; PR Discovered issues 6–7 to the logbook.

thegyula:
## Gate round 2 on tip `085d2da1` — independent gate raised one P1 and one P2; CI 7/7; fidelity gate still running — round 3 with two amended rulings

> The removal recheck can delete vault files after a scheduler-directory symlink replacement. Scheduler symlink aliases also incorrectly suppress the warning for surviving reloadable plists; both cases were reproduced with mocked filesystem operations.
>
> - [P1] Recheck vault containment before deleting discovered files — `src/core/manifest.js:1331-1335`. If a scheduler directory is replaced with a symlink into the vault while confirmation is open, `removalRoots` accepts its new location whenever the vault is inside HOME. The candidate still passes `lstatSync().isFile()` and this containment check, so `rmSync(item.path)` deletes the vault's same-named file. Discovery's vault exclusion is never reapplied during removal. Carry the accepted vault boundary into `reverse()` and exclude vault-resident targets again before deletion.
> - [P2] Keep the reload warning for scheduler symlink targets — `src/core/manifest.js:785-786`. When a scheduler-entry records a symlink to a discovered plist, canonical equality sets `ownedByScheduler: true` and `remove: false`. However, `reverseSchedulerEntry()` unlinks only the recorded symlink, leaving the target plist in LaunchAgents. The new flag suppresses the login-reload warning even though that surviving plist will restart the job against the deleted core. Distinguish ownership through a final-component symlink from a record whose reverser actually removes the candidate before suppressing the warning.

**Adjudication (orchestrator/architect).** Both real, both against rulings I wrote in round 1 — so the rulings are amended, not the implementer's reading of them.

- **R-C′ (replaces R-C's containment re-check with identity).** Re-asserting *containment* at act time re-derives the roots, and a root swapped for a symlink into the vault re-derives as contained (the vault is inside HOME). The act-time check must instead assert **identity with what was disclosed**: each discovery item carries `real` — its D15-resolved path at discovery time — and phase D5b removes only when `lstat` reports a regular file, D15 resolution succeeds, and `res.real === item.real`. Any other outcome skips with the notice (+R9 where launchd). Identity implies discovery-time containment and discovery-time vault exclusion in one comparison, and closes every swap class (final-component symlink, parent or root swapped, hardlink to elsewhere) without re-deriving anything. Item shape becomes `{path, real, remove, ownedByScheduler}`.
- **R-A′ (ownership for the warning is lexical only).** `ownedByScheduler` is true only when a validated `scheduler-entry`'s recorded path equals the discovered path **lexically** — that record's reverser removes exactly `entry.path`. A resolved-only match (the record names a symlink alias) keeps the warning: the alias is unlinked and the plist survives, which is R9's residual. `remove: false` keeps D11's lexical-or-resolved rule (preservation direction).

Round 3 runs both gates on the new tip. The fidelity reviewer's round-2 findings are appended when they arrive.

thegyula:
## Gate round 2, part 2 — fidelity gate REQUEST-CHANGES: ruling R-B was the defect; R-B′ supersedes both R-B and R-A′

**wd-reviewer (spec fidelity, round 2, executed under a redirected temp HOME):** `npm test` 2967 / 0 fail; lint 0; boundary 0; ADR-0041 amendment byte-identical; all 32 `expectRed` entries reddened exactly as declared; all three R-C swap reproductions closed, plus a symlinked root (S5) and an in-root hardlink handled correctly; no owner claim; CI 7/7 incl. macOS. **Finding 1 (HEAVY, spec-side):** with R-B as written, `uninstall.js` excludes *every* discovered path from the manifest-derived `remove` lines — including `remove:false` items owned by a `scheduler-entry`, whose own reverser deletes the file. Measured on a minimal fixture (one `scheduler-entry`, its plist present): `--dry-run` says `0 item(s) would be removed` and the disk block says `keep …dream.plist`; `--yes` then removes it (`Removed 2 item(s)`). The plan under-states deletions on every normal macOS install — consent obtained for less than is done, and a regression against `c73676bf` (which said `1 item(s)`). The implementer followed R-B literally and the worked example encodes the same output: the ruling is the defect. Other findings: (2) `removalRoots` is computed unconditionally → 5 extra `realpathSync` per `reverse()` on the no-discovery path, violating criterion 6's "no new I/O"; (3) `keepDiscovered` prints on dry-run plan passes, against D8/R-D; (4) `fs.rmSync` unwrapped — a directory swapped in after the `isFile` check still throws `ERR_FS_EISDIR`; (5) the R-B test plants only an unrecorded orphan, so the breaking case is never exercised; (6) `uninstall.js:270` JSDoc omits the third field.

**Architect ruling R-B′ (supersedes R-B and R-A′; the warning and the block line are decided by the plan, not by record kind):**
- The manifest-derived `remove` lines and the `--dry-run` headline exclude only the paths the disk block itself discloses with a `remove` line — items with `remove === true` under `unload-and-remove` (phase D5b's own deletions). A `remove:false` item's fate belongs to its owning record's reverser, and that reverser's own line stays.
- For a `remove:false` item the disk block prints, after `would run:`, **one** of: `removed by its own manifest entry (listed above)` when the dry-run `reverse()` reports the path in `removed`; otherwise `keep <path> (another manifest entry owns this file)` followed by R9's plain-language warning when the basename recognizes as launchd. This is data-driven from the same plan the user is shown, so the alias case (record names a symlink; the alias is unlinked, the target survives) warns correctly without any ownership flag.
- `ownedByScheduler` is therefore dropped; the item shape is `{path, real, remove}` (R-C′ stands).
- Every discovered path appears in the plan exactly once as a `remove`/`removed by …`/`keep` statement, and the sum of what the plan says will be removed equals what the run removes.

Finding 2 falls away with R-C′ (no root re-derivation at act time) — the criterion-6 test is extended to count `fs` calls. Findings 3, 4, 6 applied as written. Finding 5: the plan test's corpus gets an unrecorded orphan, a `scheduler-entry`-recorded plist, a `file`-recorded plist and a symlink-alias record, asserting each plan line against the file's fate after `--yes`. Round 3 runs both gates.

**Done-flip:** this family ("what the plan discloses about a discovered path") has now produced findings in two consecutive rounds across five surfaces with no primary (D3, D11, R-A, R-B, `@returns`) — the canonical-extraction pass the reviewer describes (one table keyed by recorded? × record kind × R4 cell, columns: manifest-derived line · disk-block line · warning · headline · actual fate) is the right done-flip move. Plus errata: D11 "D5b prints keep"; worked example; `opts.platform`; `scheduler.test.js` cite; Mirrored-Surface "criterion 10" ×3 and missing 9/18/19; D5 + Table B `srm-unload-moved-after-loop` and `srm-not-disclosed` anchors vs shipped declarations; DoD 6a contradiction; D3 line cites and the vault-line wording; criterion 3 gains the suppression clause.

thegyula:
## Gate round 3 on tip `fcfa9fb0` — both gates clean, CI 7/7 (macOS + ubuntu), merging

**Codex plugin `review`, `gpt-6-astra` (orchestrator-run, detached worktree, porcelain identical), verbatim:** *"No actionable regressions were identified in the changed code. The implementation follows the specified discovery, disclosure, and removal safeguards. Syntax checks passed; the test suite could not be run in the read-only sandbox."*

**wd-reviewer (spec fidelity, round 3, executed under a redirected temp HOME + `XDG_CONFIG_HOME` + `WIENERDOG_LOADER_NOOP`): APPROVE, no blocking findings.** `npm test` 2968 / 2956 / 0 fail; boundary exact (9 Deliverables + spec); all 8 spec greps; ADR-0041 amendment byte-identical, header lines 3–4 unmoved; eleven out-of-scope functions byte-unchanged against base; the spec's worked-example plan block byte-identical to the shipped output; all 11 RED declarations (31 entries) hand-applied — every declared test reddened with its signal, nothing undeclared; no owner claim beyond the mandated Status line. **Plan-vs-fate fixture (R-B′):** unrecorded orphan → `remove`, gone; `scheduler-entry` plist → `removed by its own manifest entry (listed above)` + its manifest-derived line, gone; `file`-recorded plist → `keep` + R9 warning, present; `scheduler-entry` naming a symlink alias → alias removed, target `keep` + warning, present; foreign plist absent from the plan, present. Each discovered path once; the headline equals base (`2 item(s)`) — the round-2 under-statement is gone; plan removal statements sum to the 5 files the run removed. **R-C′:** root→vault-symlink swap, parent-dir swap, out-of-root symlink, directory swap, hardlink, same-target-different-route (still removed, correct), different-file-same-path (removed — path identity, recorded as erratum E10). Criterion 6: zero added fs calls on the no-discovery path (instrumented, three configurations). D8/R-D: no D5b stderr on either dry-run pass. EISDIR: patched `lstat` + real directory → run completes with the notice. D9 abort and AC15 external-XDG survival re-executed.

**Adjudication.** No finding against the diff. Owner item 1 shipped as `unload-and-remove` under the standing process; ADR-0041 amendment — **owner signature pending**. The round-11 rulings (R-B′, R-C′, R-D; R-A/R-B/R-C withdrawn on measurement) are architect rulings recorded in the spec.

**Done-flip:** the canonical-extraction pass for "what the plan discloses about a discovered path" (one table keyed recorded? × record kind × R4 cell; D3/D11/R4/R9/`@returns`/worked example demoted to mirrors), plus errata E1–E12 from the round-3 report (D3 headline sentence and three-way enumeration; D11 "D5b prints keep"; Mirrored-Surface criterion numbering; DoD 6a contradiction; `scheduler.test.js` cite; `opts.platform`; `@returns` wording "every disclosed item D5a processed"; D5/Table B vs D6; R-C′ = realpath-string identity; shifted `uninstall.js` cites; Discovered issues 6–7 → logbook). Non-blocking observations: D5b's re-check runs on dry-run passes (over-states in the safe direction); `removed by its own manifest entry (listed above)` names no path (worked example pins it — architect call).

PR #309, opened 2026-09-19
feat(uninstall): refuse to destroy the secret quarantine (WP-adr-0019-quarantine-uninstall-gate)

thegyula:
Spec: docs/specs/WP-adr-0019-quarantine-uninstall-gate.md

## Deliverables checklist

**What landed.** `quarantineInventory` (Table K) in `src/core/manifest.js` — a
read-only, non-following walk of `<state>/quarantine` and
`<state>/quarantine/redacted` that locates the shelves by **ASCII case fold**
(K8), never opens a file (K6), never counts a shelf root (K2), and reports an
enumeration or stat failure in `unreadable` rather than swallowing it (K3/K4).
`src/cli/uninstall.js` calls it **once, before the first `console.log`** and
before Table D row D2's discovery (K5/W7), and **refuses** — identically with
and without `--yes` (W2/Y7) — while either shelf holds an entry or its state
cannot be determined. `--dry-run` prints the same disclosure block, says a real
run stops there, and continues (W5). With no shelf entries the complete output
is byte-identical to the pre-gate build (W6).

**This package changes no deleter.** `disposeCoreMechanics`, `reverse()`,
`contains`, `withinSchedulerRoot`, `withinAllowedRoot`, `validateEntry` and
`isDir` are byte-unchanged — the whole `src/` diff is additive apart from one
name added to `module.exports`.

**Dispatch-time re-verification.** Every cite was re-derived against
`origin/main` (`86f8669a`), not the spec's `5b77865f` pin. The constructs moved
but are otherwise unchanged: `disposeCoreMechanics` `:1133`→`:1470` (JSDoc
`:1447-1469`, mechanics `:1473-1478`, sweep `:1479-1487`), `contains`
`:1097`→`:1434`, `paths.state` `paths.js:70`→`:68`,
`quarantinePreserve`'s `mkdirSync` `validate.js:950`→`:953`;
`quarantinePreserve` `:936` and `REDACTED_SUBDIR` `:657` unmoved. The four
`disposeCoreMechanics` call sites are now `:408`, `:452`, `:510`, `:569`; the
manifest headline is `:400`; `readVaultPath(paths.config)` is `:360`;
`if (!yes)` is `:440`. ADR-0019's `Status:`/`Date:`/invariant paragraph are
exactly at `:3`, `:4` and `:52-54` as the spec says.

**RED criterion numbers re-derived.** Table B carries the UNSPLIT document's
numbers. Using the mapping table in
`docs/specs/logbook/2026-09-18-adr-0019-quarantine-uninstall-export-design-review.md`
("Acceptance criteria — old number → new": unsplit `3,4,5,6,7,8,10,11,12,13,15,20`
→ gate `1..12`), the declarations carry: `quse-refusal-not-raised` 3→**1**,
`quse-yes-skips-the-refusal` 4→**2**, `quse-unreadable-shelf-read-as-empty`
8→**6**, `quse-block-printed-on-an-empty-shelf` 7→**5**,
`quse-dry-run-block-suppressed` 6→**4**, `quse-shelf-name-case-sensitive`
15→**11** (its second, unsplit-9 criterion is the deletion package's, so it is
not carried here).

## Verification output

All commands run in the worktree at the commit below. Every check was also
observed in all three states — absent → red, compliant → green, violating →
red — and that observation is pasted after the commands.

```
$ npm test
ℹ tests 2990
ℹ suites 0
ℹ pass 2978
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 50833.923958

$ npm run lint
--- markdownlint ---
Linting: 714 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed

$ node scripts/boundary-check.js docs/specs/WP-adr-0019-quarantine-uninstall-gate.md $(git diff --name-only origin/main...HEAD)
(no output; exit 0)

(empty — no golden fixture changed)

# The shelves are known to the inventory and to the CLI (Table K row K1, W3).
$ test -f src/core/manifest.js && grep -n 'quarantine' src/core/manifest.js
1447:/** The two secret-quarantine shelf directory names, as `quarantinePreserve`
1448: *  writes them (`src/core/dream/validate.js`): `<state>/quarantine` for a
1449: *  withheld note and `<state>/quarantine/redacted` for a pre-scrub original.
1451:const QUARANTINE_DIRNAME = 'quarantine';
1505: * `<state>/quarantine`, a real directory whose name folds to `redacted` becomes
1510: * @param {boolean} top true only for `<state>/quarantine` itself
1547: * Inventory the secret quarantine under this core (Table K). READ-ONLY: a
1549: * <paths.state>/quarantine. It NEVER opens a file and never reads a byte of one
1558: *     order [<state>/quarantine, <state>/quarantine/redacted], each with the
1566:function quarantineInventory(paths) {
1671:module.exports = { load, record, save, reverse, disposeCoreMechanics, quarantineInventory, discoverSchedulesOnDisk, DISCOVERED_DISPOSITION, R9_LOGIN_RELOAD_WARNING, reverseSchedulerEntry, reverseVendoredTree, reverseCopiedSkill, reverseSymlink, hashDir, insertionAnchor, linkIdentity, sha256File, validateEntry, withinAllowedRoot, withinSchedulerRoot };

$ test -f src/cli/uninstall.js && grep -n 'quarantine' src/cli/uninstall.js
309: *  `WP-quarantine-only-copy-shelf`'s Table O row **O8** verified it against all
316: * The secret-quarantine disclosure block (Table W row **W4**), printed
335:function quarantineBlock(inv, paths) {
339:  const target = inv.roots.length > 0 ? inv.roots[0].dir : path.join(paths.state, 'quarantine');
361:  lines.push(`  mv "${target}" ~/wienerdog-quarantine`);
433:  const quarantine = manifestLib.quarantineInventory(paths);
434:  if (quarantine.entries > 0 || quarantine.unreadable.length > 0) {
439:      console.log(quarantineBlock(quarantine, paths));
449:        `wienerdog uninstall stopped — nothing was removed.\n\n${quarantineBlock(quarantine, paths)}`

$ test -f src/core/manifest.js && grep -Fn 'if (!dryRun) fs.rmSync(dir, { recursive: true, force: true });' src/core/manifest.js
1645:    if (!dryRun) fs.rmSync(dir, { recursive: true, force: true });
[exit 0]

$ test -f src/core/manifest.js && grep -Fn 'contains(dir, vaultPath)' src/core/manifest.js
1641:    if (vaultPath && contains(dir, vaultPath)) {
[exit 0]

$ test -f src/core/manifest.js && ! grep -Fn 'preservedQuarantine' src/core/manifest.js
[exit 0]

$ test -f docs/adr/0019-uninstall-disposes-core-mechanics.md && grep -n 'the secret quarantine is a preserved kind' docs/adr/0019-uninstall-disposes-core-mechanics.md
86:## Amendment (2026-09-19) — the secret quarantine is a preserved kind, and the invariant is narrowed to what the core already holds
[exit 0]

$ test -f docs/adr/0019-uninstall-disposes-core-mechanics.md && grep -n 'ACCEPTED under standing authorization 2026-09-18' docs/adr/0019-uninstall-disposes-core-mechanics.md
88:Status: **ACCEPTED under standing authorization 2026-09-18 — owner signature pending.**
[exit 0]

$ grep -n '^Status: Accepted$' docs/adr/0019-uninstall-disposes-core-mechanics.md
3:Status: Accepted
[exit 0]

$ grep -n '^Date: 2026-07-06$' docs/adr/0019-uninstall-disposes-core-mechanics.md
4:Date: 2026-07-06
[exit 0]

$ grep -Fn 'canonical core; the vault is always outside it** — is binding on all future' docs/adr/0019-uninstall-disposes-core-mechanics.md
53:canonical core; the vault is always outside it** — is binding on all future
[exit 0]

$ test -f docs/GLOSSARY.md && ! grep -Fn 'removes it with everything else Wienerdog keeps' docs/GLOSSARY.md
[exit 0]

$ test -f docs/runbooks/secret-incident.md && ! grep -Fn 'folder along with everything else Wienerdog keeps' docs/runbooks/secret-incident.md
[exit 0]


$ npm run red-proofs     # UNFILTERED
RED proofs — root /Users/gyulafeher/Documents/Claude_Projects/wienerdog/.claude/worktrees/wp-adr-0019-quarantine-uninstall-gate
133 declared proof(s), 133 selected
snapshot: 1749 entries under the declared domain (`.git/` and `node_modules/` excluded)

PROVEN       quse-refusal-not-raised  (WP-adr-0019-quarantine-uninstall-gate criterion 1)
             why: criterion 1 asserts an ABORT, and an abort assertion goes green whenever the run fails for any reason at all — a bad fixture, a missing manifest, a throw in unrelated setup. Same shape, same reason as WP-scheduler-replay-manifest-independent+s srm-unreadable-root-read-as-empty. The mutation makes the gate read every NON-EMPTY shelf as EMPTY while leaving UNREADABLE alone, so the run proceeds past the shelf and deletes it.
PROVEN       quse-yes-skips-the-refusal  (WP-adr-0019-quarantine-uninstall-gate criterion 2)
             why: criterion 2 is the one the whole of Table W row W2 rests on, and it is invisible to any suite that only ever drives the interactive path. A --yes suite that asserts "the run refused" passes under the mutation if it never actually passed --yes. The mutation gates the refusal on !yes, so a --yes run proceeds; note that [QU-1] (no --yes) and [QU-5] (empty shelf) stay green, which is what makes the declaration a measurement of the --yes arm specifically.
PROVEN       quse-unreadable-shelf-read-as-empty  (WP-adr-0019-quarantine-uninstall-gate criterion 6)
             why: criterion 6 asserts an abort AND zero removals, and both halves go green on a fixture that never made the shelf unreadable. This is Table K row K4 and Table Y row Y5, and it is the one failure that silently destroys the data rather than merely failing to protect it: the mutation restores the fail-open shape exactly, swallowing the error so the walk returns unreadable: [] and rmSync({force:true}) takes a tree nothing could see.
PROVEN       quse-block-printed-on-an-empty-shelf  (WP-adr-0019-quarantine-uninstall-gate criterion 5)
             why: criterion 5 asserts BYTE-IDENTITY with the pre-gate output for an install with no shelf. It is the criterion that keeps this package from changing every existing uninstall+s output, and a byte-identity assertion is trivially satisfiable by a fixture that compares the wrong stream or an empty one. The mutation prints the block unconditionally, so an install with no shelf gets a 0 file(s), 0 bytes block.
PROVEN       quse-dry-run-block-suppressed  (WP-adr-0019-quarantine-uninstall-gate criterion 4)
             why: criterion 4 observes DISCLOSURE, not the refusal. Without this declaration a suite can assert the refusal (which is loud) and never notice that --dry-run — the surface a user consults precisely to find out what an uninstall will do — says nothing about the shelf at all. The mutation drops the block from the --dry-run arm while leaving the refusal intact.
PROVEN       quse-shelf-name-case-sensitive  (WP-adr-0019-quarantine-uninstall-gate criterion 11)
             why: Round 4. Criterion 11 asserts that a capitalized shelf and the bytes inside it SURVIVE. It is a survival assertion, and it goes green on any fixture that never created the capitalized directory, created it on the wrong volume, or left it empty. The defect is invisible to every other assertion in this package: the rest classify types, codes and counts, not names, so nothing else already declared reddens under this mutation. The mutation replaces Table X row X12+s ASCII fold in the inventory+s name test with byte equality (name === +quarantine+), restoring the round-4 defect exactly.
PROVEN       extra-classes-dropped-helper  (WP-audit-d-code-derived-recipients criterion 8)
  … (every other package's declarations follow; none regressed) …
RUN: PROVEN
```

### Three-state observation of every check

```
CHECK A: grep -n 'quarantine' src/core/manifest.js
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)

CHECK B: grep -n 'quarantine' src/cli/uninstall.js
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)

CHECK C: ! grep -Fn 'preservedQuarantine' src/core/manifest.js  (a hit means package B leaked in)
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)

CHECK: grep -n 'the secret quarantine is a preserved kind' $ADR
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)

CHECK: grep -n 'ACCEPTED under standing authorization 2026-09-18' $ADR
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)

CHECK: grep -n '^Status: Accepted$' $ADR   (ADR-0035: no agent moves or reformats an authority line)
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 2)

CHECK: grep -n '^Date: 2026-07-06$' $ADR   (ADR-0035: no agent moves or reformats an authority line)
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 2)

CHECK: grep -Fn '<invariant paragraph>' $ADR
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 2)

CHECK: ! grep -Fn 'removes it with everything else Wienerdog keeps' docs/GLOSSARY.md
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)

CHECK: ! grep -Fn 'folder along with everything else Wienerdog keeps' $RUNBOOK
  compliant    -> GREEN (exit 0)
  violating    -> RED (exit 1)
  absent       -> RED (exit 1)
```

Note: the three ADR-0019 header / invariant greps are the only ones the spec
does not guard with `test -f`. They still go red when the file is absent —
with grep's exit 2 rather than 1 — so the intent holds either way.

## RED proofs — measured, not predicted

Every `expectRed` set below was MEASURED by hand-applying the mutation and
running `node --test --test-name-pattern='\[QU-' tests/unit/uninstall.test.js`,
before the declaration was written. The suite is `tests/unit/uninstall.test.js`
and each declaration is scoped with `testNamePattern: "\\[QU-"`, so the observed
set is this package's own assertions.

| Proof id | Criterion | File | Mutation | Tests that reddened (measured) |
|---|---|---|---|---|
| `quse-refusal-not-raised` | 1 | `src/cli/uninstall.js` | beside `const vaultPath = readVaultPath(paths.config) \|\| paths.vault;`, make the gate read every NON-EMPTY shelf as EMPTY (UNREADABLE untouched) | **QU-1, QU-2, QU-3, QU-4, QU-7, QU-8, QU-9** |
| `quse-yes-skips-the-refusal` | 2 | `src/cli/uninstall.js` | beside `const yes = argv.includes('--yes');`, gate the refusal on `!yes` | **QU-2, QU-3, QU-6, QU-7, QU-8, QU-9** — and NOT QU-1 (no `--yes`) or QU-4/QU-5, which is what makes it a measurement of the `--yes` arm |
| `quse-unreadable-shelf-read-as-empty` | 6 | `src/core/manifest.js` | the walk's `catch` swallows the error, restoring `unreadable: []` exactly | **QU-6** |
| `quse-block-printed-on-an-empty-shelf` | 5 | `src/cli/uninstall.js` | print the block unconditionally, so an install with no shelf gets a `0 file(s), 0 bytes` block | **QU-5** |
| `quse-dry-run-block-suppressed` | 4 | `src/cli/uninstall.js` | drop the block from the `--dry-run` arm, refusal intact | **QU-4, QU-6** |
| `quse-shelf-name-case-sensitive` | 11 | `src/core/manifest.js` | replace X12's ASCII fold in the inventory's name test with byte equality (`name === 'quarantine'`) | **QU-9** |

Two measured facts worth recording against Table B's predictions:

- `quse-dry-run-block-suppressed` also reddens **QU-6**, not QU-4 alone: QU-6's
  `--dry-run` arm asserts the unreadable shelf is *reported*, and the block is
  what reports it. Expected in hindsight; declared as measured.
- `quse-refusal-not-raised` and `quse-yes-skips-the-refusal` each redden six or
  seven tests, not one. The blast radius is real — the refusal is the single
  gate every other assertion sits behind — and the declaration records the whole
  observed set rather than the convenient subset.
## PR round 1 — three findings from the independent gate, all fixed

Round 1 on `25e9e91e`: fidelity gate APPROVE, independent gate three P2s. All
three are fixed on this tip; the spec is untouched.

1. **Y1 leak (HEAVY).** A failing `lstat` on a quarantined **file** put that
   note's basename in `unreadable[].dir`, and `quarantineBlock` prints every
   `unreadable[].dir` verbatim in both the refusal and `--dry-run` — so the one
   thing Y1 forbids reached the surface through the error path. Fixed at the
   source: `pushUnreadable()` is now the single site every unreadable level
   flows through, it takes a **directory** and the entry-level catch passes the
   **containing** directory, and the list is de-duplicated so one damaged
   permission prints one line rather than one per file. New `[QU-10]` plants a
   shelf file whose basename carries a unique token, injects an `lstat` EACCES
   on it, and asserts the token appears nowhere in the complete refusal surface
   **or** in `--dry-run`, while `<shelf> (EACCES)` still does. Two
   `manifest.test.js` tests cover the return value and the de-duplication.
2. **A non-directory shelf entry left the remedy naming a path that does not
   exist.** A file or symlink at `state/Quarantine` counted as a blocking entry
   (K2) but had no path in the result, so the remedy named the canonical
   lowercase `state/quarantine` — on a case-sensitive volume that path does not
   exist, the printed `rm -rf` succeeded silently, and the uninstall refused
   forever. `quarantineInventory` now returns **`blockers: string[]`** carrying
   the actual matched path; the block lists each one (`<path> — not a folder;
   something else is sitting where the quarantine folder goes`) and the remedy
   targets `roots[0].dir || blockers[0] || <canonical join>`. W4's line rules
   are intact: a `blockers` entry is not a shelf directory, so it takes no
   `roots` row, and an existing-but-empty shelf still contributes no line. New
   `[QU-11]` asserts the remedy names `state/Quarantine` and that the named path
   exists.
3. **Remedy lines were double-quoted.** Inside `"` a shell still expands `$`,
   `` ` `` and `\`, so a core under a home directory containing any of them
   produced a copyable `rm -rf` addressing a different path — or executing
   embedded text. Now POSIX single-quoted (`'` → `'\''`) via `shQuote()`. W4
   pins the **content** of the five items, not the quoting, so this is
   implementation. New `[QU-12]` builds a temp core under
   `a $x b ``id`` c "d" e`, takes the printed token off the `rm -rf` line, hands
   it to `/bin/sh -c 'printf %s <token>'`, and asserts the shell parses it back
   to the literal path — and that the token is single-quoted. Verified
   non-vacuous: reverting `shQuote` to double quotes reddens it.

**Re-measured after the fixes** (the `quse-unreadable-shelf-read-as-empty`
anchor moved into `pushUnreadable`, so its `find` was re-derived; the other five
anchors are unchanged). New declared sets, all measured:

| Proof id | Tests that reddened (measured, round 2) | Δ vs round 1 |
|---|---|---|
| `quse-refusal-not-raised` | QU-1, 2, 3, 4, 7, 8, 9, **11**, **12** | + the two new refusal-path tests |
| `quse-yes-skips-the-refusal` | QU-2, 3, 6, 7, 8, 9, **10**, **11**, **12** | + the three new tests |
| `quse-unreadable-shelf-read-as-empty` | QU-6, **10** | + QU-10; the mutation now empties `pushUnreadable` itself, so **both** the enumeration failure and the entry-level `lstat` failure fail open exactly as before the rule existed |
| `quse-block-printed-on-an-empty-shelf` | QU-5 | unchanged |
| `quse-dry-run-block-suppressed` | QU-4, 6, **10** | + QU-10 (its `--dry-run` arm also reads the block) |
| `quse-shelf-name-case-sensitive` | QU-9, **11** | + QU-11: under byte equality a file named `Quarantine` stops being recognised at all |

**Re-verified on this tip:** `npm test` 3016 tests / 0 fail · `npm run lint`
passed · `boundary-check` exit 0 · `git status --porcelain tests/golden/` empty ·
every spec grep green · every check re-observed absent→red, compliant→green,
violating→red · **UNFILTERED `npm run red-proofs` → `RUN: PROVEN`**, all six
declarations PROVEN, no FILTERED/VACUOUS/UNCONTROLLED/FAILED. Rebased onto
`1d8d4743` (PR #306 landed in `src/core/secret-scan.js` while round 1 ran).

**Round-1 additions to Decisions made.** (a) `blockers` is a new return field
rather than a `roots` row with `entries: 1`, because a `roots` row would
describe a file as a shelf directory "holding 1 file" and W4's per-directory
line rules are written about shelf directories. (b) When both a real shelf and a
blocker exist, the remedy targets the shelf and the blocker still gets its own
listed line, so nothing the user must deal with is invisible. (c)
`unreadable[].dir` for a shelf-root candidate stays the candidate path
(`<state>/Quarantine`): its basename already matched the closed ASCII case-fold
set, so it carries no user-derived text (Y4).

**Round-1 addition to Discovered issues.** Item 8 of the original list — "a
non-directory occupying a shelf root path is counted but gets no directory
line" — is now **fixed rather than flagged**: it has a `blockers` line and a
remedy that names it.

## PR round 2 — three platform findings fixed, four architect rulings recorded in the spec

Round 2 on `de0163de`: fidelity **APPROVE**; the independent gate raised three
platform P2s and four fidelity items were routed to the architect. The four
rulings are now recorded in the spec at the end of `## Contract reference`, under
`### PR-gate round 3 rulings (PR #309 review, 2026-09-19) — amendments to Tables
K, W and Y`, with `*Amended at PR-gate round 3 — see the end of this section.*`
at the start of cells **K2**, **W4** and **Y1** and nothing else in those cells
changed. The Exact-contracts `@returns` is now the five-field shape.
**These are ARCHITECT rulings.** Nothing in this repository records the owner
approving, accepting, ratifying or signing any of them, and the spec section
says so in its own first paragraph.

1. **R-Y1 — the Y1 leak was only half-closed.** Round 1 moved a failing `lstat`
   on a shelf *file* up to its containing directory, but a nested **directory**
   named `zzTOKENzz` whose own `readdir` failed was still reported by its own
   path, and that name is the user's text exactly as a filename is. Now every
   unreadable object below a shelf root — any depth, `readdir` or `lstat` — is
   reported by **the shelf root it sits under** and the code. `readShelfNames`
   and `countShelfTree` carry a `shelfRoot` down the walk, `redacted/` becomes
   the reporting path for everything beneath it, and `pushUnreadable` still
   de-duplicates. `unreadable[].dir` is therefore always a shelf root or
   `<state>`. New `[QU-13]` asserts it at depth 1 **and** depth 2, on the
   refusal and on `--dry-run`; two `manifest.test.js` tests cover the return
   value and the `redacted/` attribution.
2. **R-W4-win32 — the remedy was POSIX-only on a three-platform product.** A
   Windows user has no `mv` and no `rm -rf`, so the printed remedy was not a
   remedy at all for them, and a `--dry-run` that prints an unrunnable command
   is the disclosure failure W5 exists to prevent. `quarantineBlock` now takes
   the platform (defaulting to `process.platform`) and `remedyLines()` emits
   `mv '<p>' ~/wienerdog-quarantine` / `rm -rf '<p>'` with POSIX `'` → `'\''`
   on darwin and linux, and `Move-Item -LiteralPath '<p>' -Destination
   "$HOME\wienerdog-quarantine"` / `Remove-Item -LiteralPath '<p>' -Recurse
   -Force` with PowerShell `'` → `''` on win32. `quarantineBlock` is exported
   for that seam alone — no production caller imports it. New `[QU-14]` asserts
   both branches by injection over a path holding `'`, `$`, a backtick and `"`,
   and still hands the POSIX token to `/bin/sh` to prove it parses back to the
   literal path; **no test needs a Windows host**. `[QU-12]` (the real-shell
   fixture) gains a non-win32 skip, because a Windows filename cannot contain
   `"` and there is no `/bin/sh` there — the *fixture* is host-shaped, not the
   rule, and `[QU-14]` covers the rule everywhere.
3. **R-QU7 — the ordering fixture was darwin-only.** `[QU-7]` arranged its
   unreadable scheduler root under `~/Library/LaunchAgents`, which is a
   discovery root on darwin alone, so it skipped on linux and the two RED
   declarations citing it were unprovable there. It now injects the failure
   **code** on `<core>/schedules` — `discoverSchedulesOnDisk`'s third root, a
   root on darwin, linux and win32 alike — and has **no platform skip at all**.
   It also carries its own control on a separate install: with the shelf emptied
   the same injection reaches D9 and D9 speaks, which is what makes the
   precedence assertion a measurement rather than a dead fixture. Swept the rest
   of this package's tests: `[QU-12]`'s new skip is the only one left, and it is
   a fixture limitation with `[QU-14]` covering the rule.
4. **`GATE_STRINGS` had rotted into a vacuous assertion.** Round 1 reworded the
   remedy heading to *"Move that somewhere you keep"*, and `[QU-5]`'s absence
   list kept asserting that *"Move that folder somewhere you keep"* never
   appeared — a string the product no longer emits, so that element could never
   fail. Rather than re-transcribe it, `gateStrings()` now **renders the block
   itself** (both message shapes × both platform branches, over a synthetic
   core) and takes every line that does not carry the synthetic path, plus the
   two lines the CLI adds around it. The list cannot drift from the product
   again.
5. **R-K** needed no code change — `blockers` and the W4 item (3b) line shipped
   in round 1 — so this round writes the contract down: the `@returns` shape,
   the K2 entry/bytes rule for a blocker, the item (3b) line text, and the
   remedy's target precedence (first shelf root, else first blocker, else the
   canonical `<state>/quarantine`). The JSDoc field order now matches the
   ruling.

**Re-measured, all six** (the `quse-unreadable-shelf-read-as-empty` anchor moved
again with `pushUnreadable`'s signature and was re-derived; the other five
anchors are unchanged, every one `occurrences: 1`):

| Proof id | Tests that reddened (measured, round 3) | Δ vs round 2 |
|---|---|---|
| `quse-refusal-not-raised` | QU-1, 2, 3, 4, 7, 8, 9, 11, 12 | unchanged |
| `quse-yes-skips-the-refusal` | QU-2, 3, 6, 7, 8, 9, 10, 11, 12, **13** | + the nested-directory arm |
| `quse-unreadable-shelf-read-as-empty` | QU-6, 10, **13** | + the nested-directory arm |
| `quse-block-printed-on-an-empty-shelf` | QU-5 | unchanged |
| `quse-dry-run-block-suppressed` | QU-4, 6, 10, **13** | + the nested-directory arm |
| `quse-shelf-name-case-sensitive` | QU-9, 11 | unchanged |

`[QU-14]` reddens under none of the six, which is correct: it drives the block
builder directly and none of these mutations touches `remedyLines`.

**Host-independence of the ordering declarations, stated explicitly.** The
unfiltered `npm run red-proofs` below was run on **darwin**. Under R-QU7,
`[QU-7]` no longer skips on any platform, so `quse-refusal-not-raised` and
`quse-yes-skips-the-refusal` — the two declarations that cite it — are now
provable on linux too; before this round they would have observed a SKIP there
and the declared identity would not have run.

**One more finding, and the unfiltered lane is what found it.** The first
unfiltered run of this round came back **`RUN: FAILED`** — not on any of this
package's six, which were all PROVEN, but on
`WP-scheduler-replay-manifest-independent`'s `srm-resolution-failure-read-as-external`:

```
[srm-resolution-failure-read-as-external] RED: ["[QU-7] AC7 (W7): …"] failed in
its OWN BODY but is not declared — a red whose reason is not the cell's is not a
measurement
```

The control arm I had added to `[QU-7]` asserted **D9's verdict** ("the injected
scheduler-root failure does abort on its own"), and that declaration has no
`testNamePattern`, so the whole suite runs under it and my undeclared red broke
its equality check. That package is merged and its proofs file is **not in this
spec's Deliverables table**, so the fix is on my side: `[QU-7]` now asserts
**precedence only**, which is its own concern and is insensitive to how D9
classifies a resolution failure. The control it wanted already exists as
`WP-scheduler-replay AC17`, which is the *sole declared red of that very
mutation* — duplicating it here bought nothing and cost a false FAILED. I then
swept **all eleven** of that package's declarations against this suite: none
reddens any `[QU-*]` test. `[QU-7]`'s own membership in
`quse-refusal-not-raised` and `quse-yes-skips-the-refusal` is unchanged, so the
six declared sets did not move.

**Verification on this tip:** `npm test` 3020 tests / 0 fail (12 skipped) ·
`npm run lint` passed · `boundary-check` exit 0 ·
`git status --porcelain tests/golden/` empty · all 12 commands of the spec's
literal verification block exit 0 · **UNFILTERED `npm run red-proofs` →
`RUN: PROVEN`**, all six declarations PROVEN, and **no FAILED, VACUOUS,
UNCONTROLLED or FILTERED verdict anywhere in the run**. Branch base is still
`1d8d4743`; `main` has since moved to `391db43f` (#310), which touches nothing
this package touches, so no rebase was taken.

**Round-2 additions to Decisions made.** (a) `quarantineBlock` is exported
solely as the platform seam R-W4-win32 requires; no production caller imports
it. (b) `gateStrings()` renders the product rather than transcribing it, because
the transcribed list had already rotted once. (c) `[QU-7]` carries no control
arm, for the boundary reason above — recorded here rather than left as a silent
deletion.

**Round-2 addition to Discovered issues.** A declaration with no
`testNamePattern` runs the whole suite, so **any test added to
`tests/unit/uninstall.test.js` by any future package can break a merged
package's RED declaration** merely by reddening under its mutation. Nothing here
is wrong — the lane is behaving exactly as designed — but the coupling is
invisible until the 25-minute unfiltered run, and the remedy (scope the
declaration, or keep the new test insensitive) is a choice the *new* package has
to make blind. Not fixed: `scheduler-replay-manifest-independent.proofs.json` is
not in this spec's Deliverables table.

## PR round 3b — the Windows-path defect, and two test-portability items

Round 3 on `3cbb611c`: fidelity **APPROVE**; the independent gate raised three
P2s. Rulings **R-K′** and **R-QU7′** are appended to the spec's round-3 rulings
section, W4's cell now reads *"carries exactly five things (plus item (3b) when
a blocker is present — see the round-3 rulings)"*, and the Mirrored Surface
Checklist gains two bullets registering the Exact-contracts `@returns` block and
the rulings section as mirrors of Tables K/W/Y. Architect rulings throughout;
nothing here records the owner approving anything.

1. **A real Windows-path defect.** `psQuote` doubled only U+0027, but the
   PowerShell parser also closes a single-quoted string on **U+2018 '**,
   **U+2019 '**, **U+201A ‚** and **U+201B ‛**. `O'Connor` is an ordinary folder
   name on both macOS and Windows, and it ended the quoting early, leaving the
   rest of the path as bare PowerShell code inside a `Remove-Item -Recurse
   -Force`. Measured against a real parser before and after: with the old
   quoting `ParseInput` returns *"The string is missing the terminator: '."*;
   with all five doubled it parses and `-LiteralPath` reads back as the literal
   target. `PS_QUOTE_CHARS` now names all five and `psQuote` doubles each.
2. **`[QU-14]` split, and the rule separated from the fixture.** `[QU-14]` keeps
   every string assertion cross-platform over a path carrying all five
   delimiters plus `$`, a backtick and `"`, and asserts each curly quote is
   doubled rather than passed through. `[QU-14a]` carries the real `/bin/sh`
   invocation, guarded on win32. **`[QU-14b]` is new**: it writes the printed
   line to a file (never through an argv), hands it to `pwsh -NoProfile
   -Command` using
   `[System.Management.Automation.Language.Parser]::ParseInput`, walks the AST
   for `-LiteralPath` and asserts its `.Value` equals the literal target —
   skipped where `pwsh` is not on PATH. Verified non-vacuous: reverting
   `psQuote` to U+0027-only reddens `[QU-14]` **and** `[QU-14b]`, and leaves
   `[QU-14a]` green, which is exactly the split the two tests exist to make.
3. **CLI-level remedy assertions no longer hard-code POSIX.** `[QU-1]` and
   `[QU-11]` selected `mv`/`rm -rf` literally and would have failed on win32 for
   a reason unrelated to what they test. Both now call
   `expectedRemedyLines(target)`, which renders the product's own builder at
   `process.platform` — so the expectation is host-selected *and* cannot drift
   from the product, the same trick `gateStrings()` uses.
4. **R-K′ is asserted, not just written down.** A regular file at the
   `<quarantine>/redacted` position is an ordinary K2 entry under `quarantine` —
   one entry, contributing its size — and **not** a blocker; the refusal fires
   and the remedy names the quarantine root, which exists. New
   `manifest.test.js` test.

**Verification on this tip:** `npm test` 3023 tests / 0 fail (12 skipped) ·
`npm run lint` passed · `boundary-check` exit 0 ·
`git status --porcelain tests/golden/` empty · all 12 commands of the spec's
literal verification block exit 0. **None of the six RED `find` anchors moved** —
`psQuote` is not an anchor and no declaration mutates it — so the six declared
sets stand as measured at round 3 and the unfiltered lane was not re-run for
this commit. Base is still `1d8d4743`.

## Decisions made

1. **Where the gate sits relative to the manifest refusals.** W3 fixes the gate
   *before the first `console.log`* and W7 fixes it *before D2 discovery*;
   neither fixes it relative to the shipped `no install manifest found` /
   `install manifest is corrupted` refusals. Simpler option taken: the gate goes
   **after** them (immediately after the `vaultPath` read at `:360`, before the
   discovery call). Those refusals keep their shipped precedence, and a
   directory that was never a Wienerdog install is unaffected. Every W3/W7
   requirement still holds — nothing is printed before `:400`.
2. **The "nothing was removed" lead is on the refusal only.** `quarantineBlock`
   is shared byte-for-byte between the refusal and `--dry-run` (same counts,
   byte totals, directories and hedge, as W5 requires); the
   `wienerdog uninstall stopped — nothing was removed.` line of W4 item (1) is
   prepended by the thrower, because on a `--dry-run` nothing was going to be
   removed anyway and the line would misdescribe the surface.
3. **When both NON-EMPTY and UNREADABLE hold, the message takes the UNREADABLE
   form.** W4 says the unreadable variant names each directory and its `code`
   "instead of a count"; it does not say what happens when both are true.
   Simpler option: one branch, unreadable wins (it is the state that cannot be
   reasoned about). Items (1), (4) and (5) — nothing removed, the hedge, the
   remedy — are in both forms.
4. **The remedy's shell lines name a concrete destination.** W4 item (5) asks
   for "two literal shell lines the user can copy". `mv '<dir>' ~/wienerdog-quarantine`
   and `rm -rf '<dir>'` are literally runnable; a `<placeholder>` would not be.
   The path is POSIX **single**-quoted (round 1, finding 3) so a core under a path holding `$`, a backtick or a quote still copies to the literal path.
5. **`readdirSync` results are sorted.** Not required by Table K, but the
   refusal's per-directory lines and the walk order are then deterministic
   across platforms, which is what makes criterion 5's byte-identity comparison
   meaningful.
6. **An enumeration failure on `<state>` itself is reported as `unreadable`.**
   K1 says the shelves are *located* by enumerating `<state>`, and K3 says any
   non-absence code "from any `readdir` or `lstat` in the walk" is UNREADABLE.
   That enumeration is part of the walk, and if it fails we cannot know whether
   a shelf exists — so it fails closed. (K4's contrary "a failure enumerating
   `<state>` itself still throws" clause is scoped to `disposeCoreMechanics`
   via X13, which is package B's and untouched here.)
7. **Criterion 5's "captured expectation" is a same-build baseline run.** The
   criterion asks for byte-identity "to the same run at `5b77865f`". A unit test
   cannot run an old commit, and a static golden of a temp-path-bearing plan
   would drift. `[QU-5]` instead runs BOTH fixtures — (a) no shelf directory,
   (b) both shelf directories present and empty — normalises each run's own
   `<CORE>`/`<ROOT>`, and asserts the **complete stdout** of `--dry-run` and of
   `uninstall --yes` are equal between them, plus that none of the five gate
   strings appears in any of the four outputs. Fixture (a) is the captured
   expectation and it is unchanged from `5b77865f` by construction, because the
   gate prints and throws nothing when `entries === 0 && unreadable.length === 0`
   (asserted: the `Removed N item(s)` line and the full mechanics plan are inside
   the compared text). The round-6 divergence — three removed paths where the
   baseline reports one — reddens the comparison.
8. **`disposeCoreMechanics`'s JSDoc was left byte-unchanged** (see Discovered
   issues #4).

1. **Four acceptance criteria still carry deletion-side clauses that the split
   moved to `WP-uninstall-shelf-deletion-guards`, and this package's own
   verification step forbids implementing them.** The step
   `! grep -Fn 'preservedQuarantine' src/core/manifest.js` must succeed
   silently, so none of these is reachable here:
   - **Criterion 4**, final clause: *"in which `<state>` no longer appears under
     the machine-generated-state heading and the preserved shelf does"* — needs
     X4/X15. `[QU-4]` asserts everything up to it and that today's plan still
     prints unchanged.
   - **Criterion 5**'s *(Table X row **X4**)* attribution and its
     "`removed` granularity" reasoning — X4 is package B's.
   - **Criterion 6**, final sentence: *"And at the deleter: `disposeCoreMechanics`
     given the same injected failure **preserves** `<state>` and reports it in
     `preservedQuarantine`"* — package B's. `[QU-6]` asserts the gate half in
     full (EACCES and EIO abort naming the directory and the code, zero
     removals, `--dry-run` reports and continues, ENOENT does not abort).
   - **Criterion 11**, arms **(b)** and **(c)** — both drive the sweep and both
     assert `preservedQuarantine`. `[QU-9]` implements arm **(a)**.
   Same class: **Table W row W5**'s sentence *"Under the carve-out that plan
   correctly no longer lists `<state>` … and instead lists the preserved shelf
   from `preservedQuarantine` (**X4**)"* is **false for this package alone** —
   on this diff the `--dry-run` plan is today's plan, unchanged, and `<state>`
   still appears once. That is the round-6 behaviour W5's own last sentence
   describes, so no user-visible contradiction ships.
2. **Two criterion cross-references kept the unsplit document's numbering.**
   Criterion 2 says *"Criterion 3 holds identically for `wienerdog uninstall
   --yes`"* and criterion 4 says *"the same counts and byte total as criterion
   3's refusal"*. Under this file's numbering both mean **criterion 1**
   (criterion 3 here is the no-filename/no-byte one). Implemented as
   criterion 1; `[QU-2]` and `[QU-4]` assert against `[QU-1]`'s fixture.
3. **Table B's prose undercounts its own new-code declarations.** It says *"five
   carry a pre-measurable anchor and one mutates code this package authors
   (`quse-unreadable-shelf-read-as-empty`)"*, but the table marks **two** rows
   *"new — authored by this package"* — `quse-shelf-name-case-sensitive` is the
   other. The table is right; the prose sentence is wrong.
4. **Current state item 2 assigns the `disposeCoreMechanics` JSDoc erratum to
   this package, but the Deliverables table does not.** Item 2 says the
   *"none manifest-tracked, none user-authored"* sentence at `:1112-1116`
   (now `:1450-1453`) *"is false on this tree and must be corrected by this
   package (Table X row **X9**)"* — while the `src/core/manifest.js` row says
   **`quarantineInventory` ONLY** and X9 is canonical in
   `WP-uninstall-shelf-deletion-guards`. Followed the Deliverables table: the
   JSDoc is byte-unchanged and the correction travels with X9.
5. **Table B's declared anchor for `quse-dry-run-block-suppressed` is not a
   usable mutation site on this implementation, and the declaration deviates.**
   The declared anchor is `item(s) would be removed, ${skipped.length} skipped.`
   (still unique, `:418`), which sits *inside* the `--dry-run` arm — but W5
   requires the block to print **before** today's plan, i.e. before the manifest
   headline at `:400`, so no find-string built around that line can suppress it.
   The declaration mutates the `--dry-run` arm of the gate itself (code this
   package authors) with exactly the specified semantics — *drop the block from
   the `--dry-run` arm while leaving the refusal intact* — and it PROVES.
   The other five use their declared anchors, including both `quse-refusal-not-raised`'s
   `const vaultPath = …` and `quse-yes-skips-the-refusal`'s
   `const yes = argv.includes('--yes');`.
6. **Measured: for `quarantine` and `redacted` specifically, an explicit ASCII
   fold and `String.prototype.toLowerCase` are observationally identical.** K8
   requires the explicit fold because `toLowerCase`'s Unicode mappings are
   broader than the property being tested. Swept the whole Unicode range
   (U+0080–U+10FFFF, surrogates excluded) for a code point whose
   `toLowerCase()` is a single ASCII character occurring in `quarantine` or
   `redacted`: **zero hits.** So no fixture can distinguish the two functions on
   these two names — the rule is a correct defensive constraint, not an
   observable behaviour, and the reddening evidence for K8 comes from the
   byte-equality mutation (`quse-shelf-name-case-sensitive`), which `[QU-9]`
   catches. Recorded because a reviewer looking for a `toLowerCase`-vs-fold test
   will not find one and should know why.
7. **Pre-existing, unrelated:** running `node --test tests/unit/uninstall.test.js`
   directly (rather than through `npm test` / `tests/run.js`) fails
   *"Table T forgotten-seam determinism … the suite guard is inherited"*,
   because `WIENERDOG_TEST_NO_REAL_SCHEDULER` is set by the runner, not by the
   file. Reproduced on `origin/main` with this branch stashed. Not touched.
8. **A non-directory occupying a shelf root path is counted but gets no
   directory line.** K2 makes it one entry; W4 item (3) gives a line only to
   *"shelf directories that hold at least one entry"*, and such an object is not
   a shelf directory. The refusal therefore reports `1 file(s), 0 bytes` with no
   path line in that (adversarial, same-user) state. Implemented as specified;
   flagged because a reviewer may read it as a gap. Asserted in
   `tests/unit/manifest.test.js`.

## Lessons

- WP-adr-0019-quarantine-uninstall-gate: a RED-proof `expectRed` entry must fail
  with `ERR_ASSERTION`, so a test that snapshots the filesystem *after* the run
  it is asserting about will fail the lane with `ENOENT` instead of reddening —
  the mutation deleted the tree the snapshot walks. Assert the exit status and
  the tree's existence **before** the snapshot; the ordering is load-bearing
  evidence plumbing, not style. Cost two FAILED verdicts to find.
- WP-adr-0019-quarantine-uninstall-gate: when a design gate's document is split
  into two packages, the acceptance criteria are the surface the split leaks
  through. Contract rows were re-homed with their ids intact, but four criteria
  and one table cell still describe the sibling's return fields — and one of
  this package's own verification greps exists precisely to reject them. Check
  every criterion against the Deliverables table before implementing, not after.
- WP-adr-0019-quarantine-uninstall-gate: tag a package's tests with a unique
  bracketed token (`[QU-n]`) and give every assertion in a test the same `S`
  prefix. `testNamePattern` then scopes each RED declaration to the package's
  own assertions, and whichever assertion reddens carries the signal — without
  it you must predict which assertion fires and put the signal only there.
- WP-adr-0019-quarantine-uninstall-gate: `quarantineInventory` is *located* by
  ASCII fold but the two shelf names contain no character any Unicode code point
  lowercases into (swept the full range, zero hits), so the fold-vs-`toLowerCase`
  rule is unobservable for them. Worth stating in the spec when a rule is
  defensive rather than testable, so an implementer does not burn a round
  hunting for the fixture that separates them.
- WP-adr-0019-quarantine-uninstall-gate: `npm test` is not `node --test <file>`
  in this repo — `tests/run.js` sets `WIENERDOG_TEST_NO_REAL_SCHEDULER`, and a
  bare `node --test` on `tests/unit/uninstall.test.js` fails a guard test that
  has nothing to do with your change. Use the scoped `--test-name-pattern` form
  for iteration, and `npm test` for the verdict.

---
Generated-by: Claude Opus 5


thegyula:
## Gate round 1 on tip `25e9e91e` — fidelity APPROVE, independent gate three P2s (one a Y1 leak); CI 7/7 — back to the implementer

> The gate handles ordinary populated shelves, but error handling exposes private filenames and the suggested cleanup commands can target the wrong path. Capitalized non-directory shelf entries also produce an ineffective recovery instruction.
>
> - [P2] Shell-escape paths in the suggested cleanup commands — `src/cli/uninstall.js:361-362`. If HOME or WIENERDOG_HOME contains `$`, backticks, or double quotes, these copyable commands do not address the literal quarantine path. Double quotes still permit variable expansion and command substitution, so copying the `rm -rf` remedy can delete a different directory or execute embedded commands. These characters are accepted by `getPaths`; use proper shell quoting rather than interpolating the path inside double quotes.
> - [P2] Keep failed file-stat paths out of the disclosure — `src/core/manifest.js:1523-1526`. When `lstatSync` fails on a quarantined file with EIO or EACCES, `full` includes the sensitive note filename. `quarantineBlock` subsequently prints every `unreadable[].dir` verbatim, leaking that filename in both the refusal and dry-run output. Report the containing shelf directory and error code instead, preserving the no-filename requirement in Table Y1. The current disclosure test exercises only successful stats and misses this error path.
> - [P2] Preserve the actual path for non-directory shelf entries — `src/cli/uninstall.js:339`. On a case-sensitive filesystem, a file or symlink named `state/Quarantine` is counted as a blocking entry but omitted from `roots`. This fallback consequently tells the user to move or delete nonexistent `state/quarantine`; `rm -rf` silently succeeds without removing the blocker, and uninstall refuses again indefinitely. Retain the actual matched path for these entries and use it in the remedy, consistent with Table K's case-folded identity and non-directory handling.

**wd-reviewer (spec fidelity, executed under a redirected temp HOME): APPROVE, no blocking finding.** `npm test` 2990 / 0 fail; lint; boundary exact; all 12 spec greps; six RED declarations observed = declared; 11 sibling declarations unbroken; `manifest.js` diff has one removed line (the old `module.exports`) so every out-of-scope function is byte-unchanged by construction; refusal fires with stdout length 0 before any print, with and without `--yes`, filesystem snapshot identical, refusal text byte-identical between the two, no filename/content token in stdout+stderr (Y1 on the success path), counts match the worked example; W5 block precedes the headline and is byte-shared with the refusal; W6 byte-identical against the *pre-gate build* `86f8669a` on both the absent-shelf and empty-shelf fixtures; K4 with a real `chmod 000` — UNREADABLE wins over NON-EMPTY, `--dry-run` still prints the plan; K8 casings refuse and are named by their on-disk path, and the reviewer's own U+0080–U+10FFFF sweep confirms no non-ASCII code point lower-cases into either name; ADR-0019 amendment and pointer byte-exact, `:3`/`:4`/`:52-54` untouched; one sentence / one clause in the two docs, #305's neighbours intact; the only owner-word line is the mandated Status line. Findings (spec-side, routed to the architect): four acceptance criteria carry deletion-side clauses this package's own `! grep preservedQuarantine` step forbids — the gate-side-only reading is ruled the only consistent one; W5's cell contradicts itself for this package; `[QU-5]` asserts same-build identity while criterion 5 demands identity with the pre-gate build (closed by the reviewer's own two-worktree run); the `mv` destination is owner item 2's declined candidate (b) — a W4 wording question; a non-directory at a shelf root prints a block with no list line; `countShelfTree` has no depth bound.

**Adjudication.** Astra's second item is a **Y1 contract violation on the error path** (the success path was tested, the `lstat`-failure path was not) — HEAVY. The first and third are remedy-correctness defects the fidelity gate's finding 5 brushed against: on a case-sensitive filesystem a non-directory `state/Quarantine` blocks forever with a remedy naming a path that does not exist, and a HOME containing `$` or a backtick makes the copyable `rm -rf` line address a different path. All three are implementation fixes within W4/K2/Y1 as written; the spec is untouched. Round 2 runs both gates.

**Done-flip errata** (the split leaked into this spec — one split-errata pass): criteria 4/5/6/11 deletion-side clauses; W5 cell split; Table B prose "five … one" → four/two; Table B anchor for `quse-dry-run-block-suppressed` → "new — authored by this package"; Table B Criterion column re-derived (3→1, 4→2, 8→6, 7→5, 6→4, 15→11); criteria 2/4 "criterion 3" → 1; Current-state item 2's X9 → deletion-guards; W9 "will not remove this folder" → "while it still holds a file"; cites re-pinned from `5b77865f` to `86f8669a` (the PR body has them); GLOSSARY "disposable" beside the new clause (sibling's range); register the acceptance-criteria list and Table B's anchor/criterion columns as mirrors. ADR-0019 amendment — **owner signature pending**.

thegyula:
## Gate round 2 on tip `de0163de` — fidelity APPROVE (with four spec-side routings), independent gate three P2s (platform); CI 7/7 — round 3 with architect rulings

> The change introduces Windows command/test failures and RED-proof declarations that cannot pass on Linux. Full filesystem-based tests were not run because the environment is read-only.
>
> - [P2] Emit remedy commands appropriate for Windows — `src/cli/uninstall.js:385-386`. On native Windows, the refusal prints POSIX commands that do not work reliably in PowerShell: `rm -rf` is not valid `Remove-Item` syntax, and POSIX apostrophe escaping is incompatible with PowerShell. Users following the advertised copyable remedy can remain blocked from uninstalling. Emit appropriately escaped PowerShell commands on Windows, or explicitly identify the required shell.
> - [P2] Make the declared ordering proof executable on Linux — `tests/red-proofs/adr-0019-quarantine-uninstall-gate.proofs.json:40-44`. Both `quse-refusal-not-raised` and `quse-yes-skips-the-refusal` require QU-7 to fail, but QU-7 explicitly skips every non-Darwin host. On Linux, `evaluateBaseline()` rejects that SKIP before either mutation runs, so these new proofs cannot become PROVEN. Make the ordering fixture platform-independent rather than declaring a skipped test as mandatory evidence.
> - [P2] Guard the POSIX quoting fixture on native Windows — `tests/unit/uninstall.test.js:1955-1956`. This unconditional test fails on Windows before reaching its assertions because the fixture directory contains `"`, which is forbidden in Windows filenames. Its later `/bin/sh` invocation also assumes a POSIX host. Add a platform guard for this POSIX-specific test and use a Windows-compatible fixture for native-shell coverage.

**wd-reviewer (spec fidelity, round 2, executed): APPROVE.** All three round-1 closures verified: six injected-fault fixtures through the real CLI — file `lstat` EACCES/EIO, nested-dir `lstat`, `readdir` EIO on `redacted`, symlinked shelf file, five failures de-duplicated to one line — no note filename on any surface; three blocker states each followed by executing the printed `rm -rf` line and a successful re-run; `shQuote` under a HOME with `$x`, a backtick, `"`, `'`, spaces and non-ASCII parses back to the literal path in `/bin/sh`; W6 byte-identical against the real base `1d8d4743` on four surfaces; block byte-shared; all six RED sets observed = declared; 17 sibling declarations unbroken; ADR-0019 amendment byte-exact; the only owner word is the mandated Status line. Findings routed to the architect: (1) a `readdir` failure on a *nested* directory under the shelf prints that directory's own name — Y1 does not decide whether a non-product directory name is inside the no-filename rule; (2) the new `blockers` return field is not in the Exact-contracts `@returns` (no canonical table cell is contradicted — Table K does not enumerate the shape); (3) W4's "exactly five things" does not enumerate the blocker line; (4) one of five `GATE_STRINGS` (`Move that folder somewhere you keep`) is a string the product never emits — that arm of `[QU-5]` is vacuous; (5) `countShelfTree` depth (unreachable under `PATH_MAX`); (6) K2 does not decide a root-path file's byte contribution. Plus three intra-cell criterion-number staleness items (Y1, Y7, Platform scope) from the split renumbering.

**Adjudication / architect rulings (PR-gate round 3, recorded in the spec on this branch):**
- **R-Y1.** An unreadable object *below* a shelf root is reported by the **shelf root it sits under** and the code — never by its own path. Closes finding 1 without ruling on authorship.
- **R-K.** `quarantineInventory` returns `{roots, entries, bytes, unreadable, blockers}`; `blockers` lists the actual on-disk path of every non-directory found at a shelf-root position (K2's "counts as one entry", contributing 0 bytes). Exact-contracts `@returns` and Table K gain it. W4 gains item **(3b)**: one line per blocker, `<path> — not a folder; something else is sitting where the quarantine folder goes`, and the remedy targets the first shelf root, else the first blocker.
- **R-W4-win32.** W4's two remedy lines are the host shell's: POSIX `mv`/`rm -rf` with single-quote escaping on darwin/linux; on win32 `Move-Item -LiteralPath '<p>' -Destination "$HOME\wienerdog-quarantine"` and `Remove-Item -LiteralPath '<p>' -Recurse -Force` with PowerShell `''` escaping. Chosen by `process.platform`, unit-tested by injection for both branches.
- **QU-7** must be platform-independent — inject the scheduler-root failure code as the Platform-scope section already prescribes for K4, so the two ordering declarations are provable on linux; the POSIX quoting test is guarded to non-win32 and a win32 quoting test of the block builder is added. `GATE_STRINGS` corrected.

Round 3 runs both gates. **Done-flip errata** (split-errata pass + this round's): criteria 4/5/6/11 deletion-side clauses; W5 cell split; Table B prose four/two, anchor for `quse-dry-run-block-suppressed`, Criterion column re-derived; criteria 2/4 "criterion 3"→1; Y1 "criterion 5"→3, Y7 "3 and 4"→1 and 2, Platform scope "criterion 8"→6; W6/criterion 5 pin `5b77865f`→`1d8d4743`; all stale cites (PR body has them); Current-state item 2 X9→deletion-guards; W4 remedy destination vs owner item 2 (b) — one clause; GLOSSARY "disposable" (sibling's range); K2 root-path file bytes (0). ADR-0019 amendment — **owner signature pending**.

thegyula:
## Gate round 3 on tip `3cbb611c` — fidelity APPROVE; independent gate three P2s (Windows portability + one PowerShell escaping edge); CI 7/7 — LIGHT round 3b, then merge

> Recovery-command escaping fails for valid Windows paths, and the added tests contain two Windows portability defects. The escaping failure was reproduced with PowerShell's parser; full filesystem tests were unavailable in the read-only sandbox.
>
> - [P2] Escape smart single quotes in PowerShell path literals — `src/cli/uninstall.js:334-335`. On Windows, a valid core path such as `C:\Users\O’Connor\.wienerdog` produces invalid recovery commands: PowerShell treats `’` as a string delimiter, but this helper escapes only ASCII apostrophes. PowerShell's parser confirms a missing-terminator error for the generated command. Escape all PowerShell-recognized single quotes so the W4 copyable remedy preserves the literal path.
> - [P2] Assert platform-specific recovery commands in CLI tests — `tests/unit/uninstall.test.js:1733-1734`. On Windows, the CLI correctly emits `Move-Item` and `Remove-Item`, so these unconditional `mv`/`rm -rf` assertions fail. QU-11 repeats the same assumption. Select the expected commands by platform rather than requiring POSIX output.
> - [P2] Guard the POSIX shell round-trip on Windows — `tests/unit/uninstall.test.js:2081-2084`. On native Windows, QU-14 still enters this loop and launches `/bin/sh`, causing `ENOENT` before reaching its Windows assertions. Keep the string assertions cross-platform, but guard the real-shell invocation as QU-12 already does.

**wd-reviewer (spec fidelity, round 3, executed): APPROVE.** R-Y1: 30 scenarios (depth 1–3 under both roots, file and directory, `readdir`/`lstat` faults) plus a 24-shape Y1 leak fuzz with tokens in basenames and contents — zero leaks, only shelf roots named, de-duplicated; R-K: file/symlink/fifo blockers each yield the actual path, the printed remedy executed and the re-run completed; R-W4-win32: `pwsh` installed locally — PowerShell's own `[Parser]::ParseInput` returns the literal target for both lines against a path with `'`, `$x`, backticks and `"`; POSIX lines round-trip through `/bin/sh`; R-QU7: no skip, ubuntu CI green, all 11 `scheduler-replay` declarations swept against the `[QU-*]` suite with none reddening; `gateStrings()` proven product-derived by mutation; six RED sets observed = declared; W6 byte-identical to the real base on four surfaces; ADR-0019 amendment byte-exact; `src/` removes exactly two `module.exports` lines; no owner claim. Findings routed to the architect: (1) R-K's "every non-directory at a shelf-root position" is implemented at `<state>/<quarantine>` only — a file at `<quarantine>/redacted` counts as an ordinary entry with its size (refusal fires, remedy terminates; nothing leaks); (2) R-QU7's blanket "every test is platform-independent" is false for `[QU-12]` (guarded skip) and `[QU-14]` (unguarded `/bin/sh`); (3) W4's "exactly five things" vs item (3b); (4) the Exact-contracts `@returns` and the rulings section are unregistered mirrors.

**Adjudication / architect rulings (round 3b, recorded in the spec's round-3 section as amendments):**
- **R-K′ (narrows R-K):** the blocker rule applies to the `<state>/<quarantine>` position only. A non-directory at the `<quarantine>/redacted` position is an ordinary entry under `quarantine` per K2 as written (one entry; a regular file contributes its size) — the refusal fires and the remedy names the quarantine root, which exists and whose removal removes the blocker. Shipped behaviour is therefore correct; the cell narrows to match.
- **R-QU7′ (narrows R-QU7):** every test runs unguarded on the CI matrix (darwin, linux); a test whose *fixture* needs a POSIX filename or a POSIX shell is guarded on win32 by an explicit skip, and every expected remedy line in a CLI-level assertion is selected by `process.platform`. R-W4-win32's PowerShell quoting escapes every code point PowerShell treats as a single-quote delimiter (U+0027, U+2018, U+2019, U+201A, U+201B), each doubled.
- W4's cell text gains "(3b)" in its enumeration; the `@returns` block and the rulings section are registered in the Mirrored Surface Checklist.

Astra's escaping item is a real defect for a valid Windows path and is closed by R-QU7′'s quoting clause; the two test-portability items are closed by its guard/selection clauses. **LIGHT: one commit, verified mechanically by the orchestrator (PowerShell parse re-run locally), then merge — no fresh external round.**

**Done-flip errata:** the round-1/2 list unchanged (split-errata pass: criteria 4/5/6/11, W5, criteria numbering in 2/4/Y1/Y7/Platform scope, W6/criterion-5 pin, Table B prose/anchor/Criterion column, Current-state item 2, cites re-pinned, W4 destination clause, GLOSSARY "disposable"), plus: K2's byte rule for the `redacted` position (R-K′), R-QU7′, PR-body line numbers stale. ADR-0019 amendment — **owner signature pending**.

thegyula:
## Round 3b on tip `0458b6a9` — LIGHT fix verified mechanically by the orchestrator; CI 7/7; merging

Delta `3cbb611c..0458b6a9`: `src/cli/uninstall.js` — `psQuote` now doubles all five PowerShell single-quote delimiters (`PS_QUOTE_CHARS`), nothing else in `src/`; tests — `[QU-14]` cross-platform string assertions, `[QU-14a]` the win32-guarded `/bin/sh` round-trip, `[QU-14b]` a real `pwsh` AST parse (skips when `pwsh` is absent), `[QU-1]`/`[QU-11]` select remedy lines by `process.platform`, a `manifest.test.js` assertion for R-K′; spec — R-K′ and R-QU7′ appended, W4's enumeration amended, two checklist bullets. Six RED `find` anchors untouched (`psQuote` is not one), so the round-3 measured sets and `RUN: PROVEN` stand.

Orchestrator verification on a detached worktree at the tip: rendered the win32 `Remove-Item` line for `C:\Users\O’Connor ‘x‛ $y `id`\.wienerdog\state\quarantine`, parsed it with PowerShell's own `Parser::ParseInput` — **0 parse errors, `-LiteralPath` `-ceq` the literal target: True**; `tests/unit/uninstall.test.js` 70/71 under bare `node --test`, the one failure being the pre-existing "Table T forgotten-seam determinism" artefact that needs `tests/run.js`'s guard env (reproduces on `origin/main`). Seven checks pass.

Owner item 1 (ADR-0019 amendment) shipped under the standing process — **owner signature pending**. Done-flip errata as listed in the round-3 comment; the split-errata pass over this spec's criteria/W5/Table B is the largest single item in that list.

PR #10, opened 2026-07-02
feat(dream): implement dream runtime pipeline (WP-017)

thegyula:
Spec: docs/specs/WP-017-dream-validate-commit.md

## Deliverables checklist

- [x] Every changed file is listed in the spec's Deliverables table
- [x] Spec status flipped to In-Review

All changed files are in the Deliverables table (plus the spec's own status
flip); `scripts/boundary-check.js` exits 0.

## Verification output

```
$ npm test -- --test-name-pattern dream-validate
# tests 22, pass 22, fail 0

$ npm test -- --test-name-pattern dream-integration
# tests 7, pass 7, fail 0 (full run/injection, no-new, dry-run, dirty tree,
#   watchdog timeout+kill, live lock, stale-lock steal)

$ npm test        # full suite
# tests 102, pass 102, fail 0

$ npm run lint
--- markdownlint ---  Summary: 0 error(s)
--- shellcheck ---    (no .sh files touched)
--- frontmatter check --- passed: 2 spec(s), 4 agent(s)
lint passed

# End-to-end via the fake brain (no real model, no network):
$ WIENERDOG_DREAM_CMD="$PWD/tests/fixtures/dream/fake-brain.js" node bin/wienerdog.js dream --yes
wienerdog: dream committed cee1b22 — 2 notes, 0 skills; 2 reverted, 1 out-of-vault.

$ git -C "$WIENERDOG_VAULT" log --oneline -1
cee1b22 dream: 2026-07-02 — 2 notes, 0 skills

$ git -C "$WIENERDOG_VAULT" ls-files 06-Identity 05-Skills   # injected.md / weak-skill absent
05-Skills/README.md
06-Identity/goals.md
06-Identity/instructions.md
06-Identity/preferences.md
06-Identity/profile.md
06-Identity/valid-identity.md

$ cat "$WIENERDOG_VAULT/reports/dreams/2026-07-02.md"
# Dream report — 2026-07-02

Consolidated recent sessions.

## Reverted by orchestrator (policy enforcement)
- `05-Skills/weak-skill/SKILL.md` — Tier-3 floor not met (derived_from_untrusted=false, confidence=0.4, recurrence=1; requires false, >=0.85, >=3)
- `06-Identity/injected.md` — Tier-3 floor not met (derived_from_untrusted=true, confidence=0.95, recurrence=5; requires false, >=0.85, >=3)
- `.../state/dream-scratch/EVIL.json` — brain wrote into the read-only scratch dir; deleted

[...]

PR #312, opened 2026-09-21
feat(uninstall): guard every deletion against the secret quarantine (WP-uninstall-shelf-deletion-guards)

thegyula:
Spec: docs/specs/WP-uninstall-shelf-deletion-guards.md

## Deliverables checklist

`node scripts/boundary-check.js docs/specs/WP-uninstall-shelf-deletion-guards.md $(git diff --name-only origin/main...HEAD)` → exit 0.
`git status --porcelain tests/golden/` → empty.

## Verification output

```
$ npm test
✔ vendor: verifyCurrentContainment accepts a contained prod tree and rejects an out-of-root symlink (WP-157) (3.058583ms)
✔ vendor: installStance is decided by containment, fails closed, and never consults ownership (8.940167ms)
✔ vendor: no .js under src/ outside the launcher reads WIENERDOG_DEV (with a non-vacuity control) (5.891416ms)
✔ vendor: launcher.liveStance ≡ vendor.installStance over every install shape (6.717417ms)
✔ vendor: readVersion refuses a tampered package.json version, and accepts real ones (1.657458ms)
✔ vendor: an attended sync carries containment forward or refuses — no DATA-shaped A7 write moves it (162.410042ms)
ℹ tests 3055
ℹ suites 0
ℹ pass 3043
ℹ fail 0
ℹ cancelled 0
ℹ skipped 12
ℹ todo 0
ℹ duration_ms 54769.265334

$ npm run lint

--- markdownlint ---
Linting: 715 file(s)
Summary: 0 error(s)
--- shellcheck ---
--- PSScriptAnalyzer ---
--- frontmatter check ---

lint passed

### preservedQuarantine (X4)
1907: * @param {string[]} removed @param {string[]} preservedQuarantine
1910:function disposeStateTree(paths, prot, removed, preservedQuarantine) {
1919:    preservedQuarantine.push(state); // UNANSWERABLE — preserve and report (X17 (4))
1929:      preservedQuarantine.push(state);
1946:      preservedQuarantine.push(state);
1963:      if (name !== QUARANTINE_DIRNAME) preservedQuarantine.push(child);
1972:      preservedQuarantine.push(child);
1993:      if (!isAbsentCode(code)) preservedQuarantine.push(level);
1997:      preservedQuarantine.push(level); // a symlink, file or socket — never followed
2012:      preservedQuarantine.push(level);
2066: *            preservedQuarantine: string[]}} `removed` KEEPS ITS ORIGINAL
2073: *   unchanged. `preservedQuarantine` lists the directories this call left in
2079:  /** @type {string[]} */ const preservedQuarantine = [];
2111:          for (const r of plan.roots) if (r.entries > 0) preservedQuarantine.push(r.dir);
2113:            if (!preservedQuarantine.includes(u.dir)) preservedQuarantine.push(u.dir);
2116:            if (!preservedQuarantine.includes(b)) preservedQuarantine.push(b);
2121:      disposeStateTree(paths, /** @type {any} */ (prot), removed, preservedQuarantine);
2138:        preservedQuarantine.push(dir);
2165:  return { removed, skippedForVault, preservedQuarantine };

### shelfGuarded (X22)
946: *            discoveredSchedules: string[], shelfGuarded: string[]}}
950: *            `shelfGuarded` is the entries the PRE-DISPATCH SHELF GUARD skipped
952: *            rendering; `shelfGuarded` is a strict subset of it, exposed
1032:  /** @type {string[]} */ const shelfGuarded = [];
1169:        shelfGuarded.push(entry.path);
1416:    discoveredSchedules: discoveredUnloaded, shelfGuarded,

### contains() byte-unchanged (X7)
1503:function contains(outer, inner) {
2097:      if (vaultPath && isDir(dir) && contains(dir, vaultPath)) {
2125:    if (vaultPath && contains(dir, vaultPath)) {

### recursive: true (Table V)
817:  if (!dryRun) fs.rmSync(entry.path, { recursive: true, force: true });
869:  if (!dryRun) fs.rmSync(entry.path, { recursive: true, force: true });
1975:    fs.rmSync(child, { recursive: true, force: true }); // a failure PROPAGATES (X13)
2142:    if (!dryRun) fs.rmSync(dir, { recursive: true, force: true });

# The three states of each new grep (absent → red, compliant → green, violating → red)
# ABSENT (no deliverable) -> must be RED
  exit=1 (preservedQuarantine)
  exit=1 (shelfGuarded)
# COMPLIANT -> GREEN
  exit=0 (preservedQuarantine, 19 hits)
  exit=0 (shelfGuarded, 6 hits)
# VIOLATING (field stripped) -> must be RED
  exit=1 (preservedQuarantine)
  exit=1 (shelfGuarded)
```

### RED proofs — the MEASURED reddening sets (Table B)

Every `expectRed` set below was measured by hand-applying the mutation and running the targeted suite BEFORE the declaration was written.

| Proof id | reds | MEASURED reddening tests |
|---|---|---|
| `quse-carve-out-dropped` | 15 | [SG-1] [SG-2] [SG-3] [SG-4] [SG-8] [SG-10] [SG-11] [SG-12] [SG-14] [SG-15] [SG-16] [SG-17] [SG-18] [SG-19] [SG-20] |
| `quse-symlinked-state-descended` | 1 | [SG-9] |
| `quse-shelf-validated-leaf-first` | 1 | [SG-10] |
| `quse-mechanics-failure-swallowed` | 1 | [SG-12] |
| `quse-dry-run-plans-by-assuming` | 1 | [SG-13] |
| `quse-reverse-shelf-guard-dropped` | 5 | [SG-14] [SG-15] [SG-17] [SG-18] [SG-21] |
| `quse-reverse-guard-from-below-only` | 1 | [SG-15] |
| `quse-absent-shelf-read-as-unanswerable` | 10 | [SG-7] [SG-9] [SG-14] [SG-16] [SG-17] [SG-18] [SG-19] [SG-20] [SG-21] [SG-22] |
| `quse-step2-deletes-through-alias` | 2 | [SG-2] [SG-4] |
| `quse-state-alias-unlinked` | 5 | [SG-3] [SG-10] [SG-15] [SG-18] [SG-19] |
| `quse-symlink-entry-breaks-chain` | 1 | [SG-18] |
| `quse-protected-set-is-anchors-only` | 2 | [SG-4] [SG-18] |
| `quse-chain-node-over-protects` | 1 | [SG-6] |
| `quse-hypothetical-root-not-subtree` | 1 | [SG-17] |
| `quse-alias-into-swept-dir-unlinked` | 1 | [SG-19] |
| `quse-alias-ancestor-of-target-unlinked` | 1 | [SG-19] |
| `quse-manifest-deleted-after-preserve` | 6 | [SG-14] [SG-15] [SG-17] [SG-18] [SG-20] [SG-21] |
| `quse-w10-checks-first-sweep-only` | 1 | [SG-20] |
| `quse-shelf-guard-skip-not-propagated` | 1 | [SG-21] |
| `quse-w10-blocks-on-absent-guarded-path` | 2 | [SG-9] [SG-22] |

## ⚠️ BLOCKER — `npm run red-proofs` is `RUN: FAILED`, on TWO of package A's declarations, and the fix is OUT OF THIS PR'S BOUNDARY

All **20** of this package's declarations are `PROVEN`. Two of
`WP-adr-0019-quarantine-uninstall-gate`'s are not, and both for the same reason:

```
FAILED  quse-refusal-not-raised  (WP-adr-0019-quarantine-uninstall-gate criterion 1)
        the declared identity ["[QU-3] AC3 (K6/Y1): the refusal prints no shelf FILENAME
        and no shelf BYTE"] did not fail under the mutation — the assertion is vacuous,
        or the mutation does not reach it
FAILED  quse-yes-skips-the-refusal  (WP-adr-0019-quarantine-uninstall-gate criterion 2)
        (same declared identity, same reason)
```

**Why it happens, and why it is the package WORKING rather than breaking.** Both mutations
blind the **gate** — the inventory reads every shelf as empty (or the refusal is gated on
`!yes`) — so the run proceeds past it. Before this package that meant the shelf was deleted
and the command exited **0**, which is what made `[QU-3]`'s `assert.equal(res.status, 1)`
redden. With the deletion side in place the blinded gate no longer decides anything: the
carve-out preserves the shelf, **W10** stops before the manifest delete, the command exits
**1**, and the message names the shelf DIRECTORY and no filename — so every one of
`[QU-3]`'s assertions now holds and it stays green. That is exactly Table X row **X3**'s
claim ("the guarantee is the kernel's, not our check's") observed from the outside.

**The fix is one edit, and it is in a file this PR may not touch.** Remove the `[QU-3]`
entry from the `expectRed` array of `quse-refusal-not-raised` and of
`quse-yes-skips-the-refusal` in
`tests/red-proofs/adr-0019-quarantine-uninstall-gate.proofs.json`. **MEASURED:** with only
that edit applied locally, both declarations report `PROVEN`; the file was then reverted and
is unchanged in this PR. `scripts/boundary-check.js` rejects it for this work package, so it
is reported here rather than fixed (CLAUDE.md: "Found something else broken? Note it under
Discovered issues; do not fix it").

A second, smaller interaction with the same two mutations IS fixed here, in commit 2: their
`replace` constructs `{roots, entries, bytes, unreadable}` and omits ruling **R-K**'s
`blockers`. **W10** and the **X15** planner both SPREAD that field, so an undefined `blockers`
turned a missing disclosure into a crashed uninstall and reddened `[QU-5]` undeclared. Both
reads are now `|| []`-guarded — that is this package's own robustness bug, not package A's.

## Decisions made

1. **One declaration file, one suite: `tests/unit/uninstall.test.js`.** `scripts/red-proofs.js`
   takes ONE `suite` per declaration file; the Deliverables table creates exactly one proofs
   file but names TWO test files. The simpler option was to host every declared reddening
   identity in the CLI suite — where the full-command harness already lives — and to keep the
   `tests/unit/manifest.test.js` additions as unit-level contracts of the two functions
   (`preservedQuarantine`, the granularity, the `ENOTEMPTY` climb, the `lstat` classification,
   the vault-guard precedence, `shelfGuarded`) rather than as declared identities.
2. **The new tests are tagged `[SG-n]`, not `[QU-n]`.** `WP-adr-0019-quarantine-uninstall-gate`'s
   declarations scope with `testNamePattern: "\\[QU-"`, which also matches `[QUSE-…]`. A distinct
   tag keeps the two packages' RED lanes disjoint.
3. **The `dryRun: true` planner reports what it would keep as: every inventory root with
   `entries > 0`, every `unreadable` level, and every `blockers` path** — rather than
   re-deriving the exact level the live climb would stop at. Table X row X15 requires the plan
   to name the preserved shelf; the simpler read of the same inventory does that.
4. **Table W row W10's stop reuses `uninstall.js:428-432`'s exit shape** — a thrown
   `WienerdogError`, so the command exits non-zero — because the row says it "reuses this
   wording and that exit".
5. **The vault guard keeps its SINGLE `contains(dir, vaultPath)` call site**, now preceded by
   `isDir(dir) &&` so a symlinked or non-directory `<state>` never reaches that helper's
   realpath-both-sides resolution before Table X row X1 step 0b has classified it with `lstat`.
   `contains` itself is byte-unchanged (Table X row X7).
6. **Fixture roots are `fs.realpathSync`'d.** On macOS `os.tmpdir()` sits behind `/var ->
   /private/var`, which would put a host artifact into every protected set the fixture builds
   and make the measured RED sets platform-dependent. `tempEnv()` gained one `realpathSync`.
7. **`quse-absent-shelf-read-as-unanswerable` is aimed at `shelfProtection`'s root-walk
   classification**, not at a literal single `catch`: that walk is where this implementation
   distinguishes Table X row X17's outcome (3) from outcome (4).

- **`WP-adr-0019-quarantine-uninstall-gate`'s `[QU-6]` ENOENT arm was FALSIFIED by this
  package's contract, and its fixture is corrected in this PR.** It faulted `readdirSync` on a
  POPULATED `<state>/quarantine` to `ENOENT` and asserted the uninstall completes with the core
  gone. Under Table X row **X1** the live deleter's guarantee is the KERNEL's — the shelf comes
  off only through `rmdirSync`, which answers `ENOTEMPTY` whatever an injected `readdirSync`
  reports — so the user's file is preserved and **W10** stops the run. The arm's subject is the
  GATE's classification of `ENOENT`, so its fixture is emptied first; its assertions are
  unchanged.
- **Table B's predicted `expectRed` sets are narrower than the measured ones, for five
  declarations.** `quse-carve-out-dropped` is described as carrying "**Two `expectRed` entries**,
  each with a ONE-element `test` path"; MEASURED: **15**. Likewise
  `quse-reverse-shelf-guard-dropped` (measured 5), `quse-absent-shelf-read-as-unanswerable`
  (10), `quse-state-alias-unlinked` (5) and `quse-manifest-deleted-after-preserve` (6). The
  declaration carries the measured sets.
- **Acceptance criterion 9's `copied-skill` clause cannot be asserted on the entry shape its
  prose implies.** `reverseCopiedSkill`'s ownership proof (parent resolves equal to a skills
  root, `wienerdog-*` basename, real-directory `lstat`, `hashDir` match) admits no path that is
  AT OR UNDER the shelf. The arm is therefore asserted on the only shape that reaches its
  recursive `rmSync` at all — a real `wienerdog-*` skill directory that **CONTAINS** the shelf,
  reached by a `<state>` alias into it, which is the from-above half the clause is about.
- **`init` REFUSES to write under a symlinked core** ("the Wienerdog core is a symlink or not a
  directory"). Criteria 2 and 9 read as though a symlinked-core install can be created directly;
  it cannot, so those fixtures install into a real directory and then move it and link.
- **`src/core/dream/validate.js`'s `quarantinePreserve` line numbers in the spec's Current state
  are stale** (`:936`/`:950`/`:953` vs `:948-958` on `main`). Not touched — that file is out of
  scope.

## Lessons

- WP-uninstall-shelf-deletion-guards: a JSDoc line containing `**X19**/**X20**` **terminates the
  block comment** — the closing `**` plus the `/` is `*/`, and the next line parses as code
  (`SyntaxError: Unexpected token '**'`). Bold table-row ids are everywhere in this repo's
  comments; never let a `**` sit immediately before a `/`.
- WP-uninstall-shelf-deletion-guards: `fs.lstatSync` classifies a path's FINAL component only, so
  a component-by-component walk must build each step under the **resolved** prefix. Test fault
  injections keyed on the lexical path then never fire — on macOS `os.tmpdir()` sits behind
  `/var -> /private/var`, so an injected code has to answer to both spellings of the same object.
- WP-uninstall-shelf-deletion-guards: `fs.realpathSync(fs.mkdtempSync(...))` in fixtures, always.
  An unresolved temp root puts a host symlink into anything that records a resolution chain, and
  that turned one declared RED set from 1 test into 12 on this host alone.
- WP-uninstall-shelf-deletion-guards: a "the bytes survive" fixture that plants the shelf file
  BEFORE the run never reaches the deleter — the gate refuses first, and the test passes for the
  wrong reason. Every such arm has to seam the preserve INTO the run, after the guard's anchors
  are computed and before the entry that would delete it is replayed.
- WP-uninstall-shelf-deletion-guards: manifest ENTRY ORDER decides whether a from-above fixture
  bites. `reverseVendoredTree`'s ownership test resolves through `<core>/app`; if the install's
  own `app` entry is replayed first it unlinks the alias, and the forged entry is then refused
  for a reason that has nothing to do with the guard under test.
- WP-uninstall-shelf-deletion-guards: `assert` on `fs.readFileSync` of a file the mutation
  deletes reddens as `ENOENT`, not `ERR_ASSERTION`, and `scripts/red-proofs.js` refuses that red.
  Read through a `readOrNull`/`isLink` helper so a destroyed original reddens as an assertion
  carrying its signal.
- WP-uninstall-shelf-deletion-guards: `pgrep -f '<script>'` in an `until` loop matches the
  waiter's OWN command line and never terminates. Wait on the PID.

---
Generated-by: Claude Opus 5


thegyula:
## Gate round 1 on tip `ba273b10` — fidelity APPROVE, independent gate three P1 guard bypasses + one P2; CI 7/7 — back to the implementer with architect rulings

> In-memory probes reproduced three guard bypasses that delete quarantined copies and a false resolution failure that blocks uninstall. The full test suite could not run in the read-only sandbox.
>
> - [P1] Canonicalize case aliases before testing shelf containment — `src/core/manifest.js:1786-1790`. On case-insensitive macOS filesystems, `walkChain()` retains the caller's spelling rather than the canonical directory name. A hash-less manifest entry naming `<state>/QUARANTINE/2026-note.md` therefore bypasses the guard when the stored directory is `quarantine`. If a dream creates that copy after the pre-plan gate, `reverse()` deletes it. Canonicalize existing components while retaining the resolution chain so case aliases receive the protection required by X12.
> - [P1] Resolve relative link targets before collapsing parent segments — `src/core/manifest.js:1780-1781`. With `<state>/quarantine -> ../jump/../recovery` and `<core>/jump -> <core>/logs/inner`, the kernel resolves the shelf to `<core>/logs/recovery`, but `path.resolve()` prematurely normalizes its target to `<core>/recovery`. A direct `disposeCoreMechanics()` sweep consequently deletes `logs`, including the quarantined copy, while reporting the quarantine link preserved. Preserve the relative target's components until preceding symlinks have been resolved, as required by X17.
> - [P1] Distinguish parent traversal from names beginning with two dots — `src/core/manifest.js:1722-1723`. When the quarantine symlink targets `<core>/logs/..recovery`, containment from `logs` produces the relative path `..recovery`. This predicate incorrectly treats that legitimate child name as parent traversal, so the mechanics sweep recursively deletes `logs` and the quarantined copy while reporting the link preserved. Reject only an actual `..` component, not every name with that prefix, to satisfy X18's recursive-deletion guard.
> - [P2] Reset the link-hop budget for each shelf-root resolution — `src/core/manifest.js:1861-1862`. The same `ctx.hops` counter is reused across independent root walks. A valid 21-link alias chain consumes 21 hops resolving `quarantine`, then falsely reports `ELOOP` while resolving `quarantine/redacted`, even when both shelves are absent. `reverse()` consequently aborts every retry on a resolvable layout. Share the collected chain nodes, but reset the hop budget for each root.

**wd-reviewer (spec fidelity, executed in three throwaway worktrees under a redirected temp HOME): APPROVE.** `npm test` 3055 / 3043 / 0 fail; lint; boundary exact (7 files incl. the ruled row); `recursive: true` exactly Table V's four sites; single `contains(dir, vaultPath)` site with the `isDir(dir) &&` guard; the carve-out executed end to end with the gate blinded — shelf bytes survive, `<state>` survives, W10 stops with the manifest still on disk, no filename/content token on any surface; 12 deleter/guard scenarios incl. symlinked shelf root, escape symlink, hardlink, `/tmp`→`/private/tmp` alias, forged `vendored-tree`/`copied-skill` from above, chain `symlink` kind (X20), hash-less `file` from below — all preserved with recursive `rmSync` never invoked; W6 byte-identical to base on both fixtures; dry-run zero mutating calls; the completion arm (X19 alias outside all roots, no shelf) exits 0 and does not block permanently; all 20 RED sets observed = declared; both gate-package mutations hand-applied on this tip with `[QU-3]` staying green; `[QU-6]` change is one added statement; 37 sibling declarations unbroken; no owner claim; ADR-0019 untouched. Findings (non-blocking): (1) the PR body's pasted verification block is from an earlier tree (pastes a failing lint, two `contains` sites, cites 5 lines off) — must be re-run on the tip; (2) W10's stop message says "your quarantined copies" for an *unanswerable-resolution* stop that may involve no shelf at all; (3) W8's summary prints only `keptQuarantine[0]`; (4) the per-entry shelf guard runs before the allowed-root bound, so an out-of-root unresolvable entry now blocks indefinitely where it used to be skipped; (5) the two test-file Deliverables cells describe a split that did not ship; (6) `shelfProtection`'s fold lookup `readdirSync` reads through a symlinked `<state>` (a read, adds anchors only); (7) the planner's `isDir` short-circuit; (8) the round-1 ruling's quoted row literal ≠ the row as shipped; (9) a manifest-supplied shelf-file path can echo to stderr (not Y1 — the product never puts one there); (10) `quse-w10-checks-first-sweep-only` `criterion: "10"` should be 4. Criterion 9's `copied-skill` clause ruled unsatisfiable as written; the implementer's reading accepted.

**Adjudication.** Astra's three P1s are real guard bypasses in the resolution code (a case alias on a case-insensitive filesystem; a relative link target normalised by `path.resolve` before preceding links resolve; a `..`-prefix test that treats `..recovery` as traversal) and the P2 is a false permanent block — all HEAVY, all implementation against X12/X17/X18 as written. Architect rulings for the fidelity items: **W8′** — every preserved path is printed, one per line (W8's literal widened); **W10′** — the stop message branches: a shelf-derived stop says the quarantined copies are still there and how to move or delete them; an unanswerable-resolution stop names the path that could not be read and its code and says the fix is the permission, never "quarantined copies"; **X16′** — the per-entry shelf guard applies to entries that pass `withinAllowedRoot`; an entry the root bound rejects is skipped exactly as before this package (no `shelfGuarded`, no W10), because nothing would have deleted it; the round-1 ruling's quoted literal is amended to the row as shipped; the two test-file Deliverables cells are rewritten to where the tests landed (`uninstall.test.js` owns declared identities); the `criterion` field 10→4. Round 2 runs both gates.

**Done-flip errata (beyond the above):** criterion 9 wording; Table B's five under-predicted `expectRed` cells struck; predecessor spec path now `docs/specs/done/…`; `quarantinePreserve` pins +3; Current-state item 9 already corrected by the gate package; every `manifest.js` pin off by hundreds of lines (one sentence); criteria 2/5/9's absolute fixture paths and the `init`-refuses-symlinked-core fixture shape; Security-checklist "every stop has a terminating remedy" narrowed per findings 2/4; X10 clarifying sentence (type classification vs name enumeration); X17 canonical-extraction candidate (2–4k-character cells carrying 5–7 rules).

thegyula:
## Gate round 2 on tip `dfb3cba5` — independent gate two more P1 bypasses + two P2s; fidelity REQUEST-CHANGES on three contract cells; CI 7/7 — round 3 with an architect ruling that freezes the resolver's design

> Two path-resolution gaps allow quarantined originals to be deleted, and two preservation-reporting issues incorrectly block uninstall. Targeted read-only and in-memory probes reproduced these issues; the full test suite was not run under the read-only sandbox.
>
> - [P1] Canonicalize non-ASCII components before checking containment — `src/core/manifest.js:1793-1796`. On a case-insensitive volume, consider `quarantine -> ω/recovery` with the sibling stored as `Ω/recovery`. This shortcut leaves the protected target spelled `ω`, while enumeration supplies `Ω`; the containment check misses the overlap and `disposeStateTree()` recursively deletes the sibling, including its quarantined originals, while reporting the link preserved. Canonicalize existing non-ASCII components too; the resolved-target protection required by X17 cannot assume that only ASCII letters have case aliases.
> - [P1] Anchor relative manifest paths before walking their components — `src/core/manifest.js:1861-1866`. `validateEntry()` accepts relative paths, but this walker starts them at an empty root and effectively discards leading `..` components. With CWD `<core>/work`, `<core>/cache -> <state>/quarantine`, and a hash-less entry `../cache/n.md`, the guard checks `work/cache/n.md` instead. The reverser's subsequent `fs.realpathSync()` resolves the actual shelf file and deletes it if it arrived after the initial gate, bypassing X16. Anchor relative inputs at CWD before processing components, without lexically collapsing symlink-sensitive `..` segments.
> - [P2] Reject resolved out-of-root targets from the shelf guard — `src/core/manifest.js:1051-1055`. Accepting either the literal or resolved path does not reproduce `withinAllowedRoot()`. With a symlinked home `/home -> /real-home`, a recorded skill link retargeted to `/real-home` is out of bounds for the reverser, but its literal location passes this check and its target matches a protected chain node. It consequently enters `shelfGuarded`, and W10 blocks every retry despite empty shelves. Use the resolved boundary when resolution succeeds, as required by X16′.
> - [P2] Reconcile preserved aliases with successful directory removals — `src/core/manifest.js:2100-2101`. On a case-insensitive filesystem, an empty directory stored as `Quarantine` is added to `preservedQuarantine` here, but the subsequent lowercase `rmdirSync()` removes that same directory and then `state`. The result reports `state` removed and its nonexistent child preserved, contrary to X4. W10 then aborts the uninstall claiming quarantined copies remain, unnecessarily requiring another run.

**wd-reviewer (spec fidelity, round 2, executed): REQUEST-CHANGES on contract cells; "on behaviour I have no reservations".** All four round-1 shapes reproduce on `ba273b10` and are closed here; differential fuzz of the resolver against `fs.realpathSync.native` over 2 600 layouts / 17 600 probes with 0 mismatches (ASCII case, `.`, empty and trailing segments, absolute/relative/`..` targets, dangling links, cycles, 25-link chains); hop budget over-protects only; W8′/W10′/X16′ verified as behaviour on A/B fixtures; remedies terminate; 20/20 RED sets exact, `[SG-23..26]` red on `ba273b10`; `[QU-3]` green under both gate mutations; 17 sibling declarations intact; six safety fixtures — no shelf byte lost, `<state>` never removed while non-empty, `secrets/` removed in all six; TOCTOU on the spelling lookup lands inside `R-alias-outside-closure` (cited, not re-opened). Blocking: (1) `reverse()` gained `shelfUnanswerable` — recorded in no canonical table, the Exact-contracts JSDoc still lists one new field, no test pins it; (2) the vault-guard lines were changed against X7/X8's "untouched" and the comment above them still says "untouched"; (3) the round-1 ruling's quoted row is still truncated by 68 characters. Non-blocking: X16's cell carries only the pointer; criterion 10 not amended for W10′ (nor 4 for W8′); `[SG-24]`/`[SG-25]` carry no RED declaration; the `[SG-1]–[SG-22]` range claim; the per-entry stderr notice at `:1200` still says "secret quarantine" for an unanswerable entry; a stale `6 hits` in the PR body; small nits. X17 (≈4k characters, 5–7 rules, findings in three consecutive rounds) is the canonical-extraction candidate.

**Adjudication.** The resolver has now been re-derived by hand five times (ASCII case, relative targets, `..`-prefix, non-ASCII case, relative inputs) and each round's fuzz was blind to the next round's shape because the fuzz generated only what the previous round had imagined. The convergence rule applies: **the design is frozen by handing the question of "what does this existing path canonically resolve to" to the kernel** — architect ruling **X17′** below — and the package's own walk is reduced to what the kernel cannot answer (chain nodes, and the absent suffix). Astra's two P1s are then closed by construction, and the fuzz is extended to Unicode case pairs and relative inputs so the next shape cannot hide. The two P2s are ruled **X16″** (resolved boundary when resolution succeeds) and **X4′** (the preservation report is reconciled against what still exists after the sweep). The three fidelity blockers are contract records to be made true, not code. Round 3 runs both gates; after it, prose-only findings go to errata.

**Done-flip errata (accumulating):** Current-state items 1/9 void (package A landed); every `src/` pin historical (one sentence); `quarantinePreserve` pin `:950`→`:953` only; Table B "nine/eleven" inverted, `quse-carve-out-dropped` "two" → measured; unsplit criterion numbers at `:245/:256/:807` (7→3), `:257/:259` (14), `:309/:868` (18), `:521` (8), `:845` (9→4); predecessor path → `done/`; Security-checklist termination attribution (X16′, W10′) and the unregistered security bullets; checklist `:372-373`; `quse-w10-checks-first-sweep-only` `why` prose vs field; X17 canonical extraction.

thegyula:
## Gate round 3 on tip `0bfe5ec8` — identity/spelling family CLOSED by X17′; independent gate three P1s in the chain-collection family + one P2; fidelity two mechanical blockers; CI 6/7 (ubuntu red on `[SG-27]`) — round 4 is the last product round

> In-memory replay/sweep probes reproduced three incomplete-protection cases that delete quarantined copies, plus a deterministic dry-run mismatch. Syntax checks passed; full filesystem tests could not run in the read-only sandbox.
>
> - [P1] Preserve raw relative targets during chain collection — `src/core/manifest.js:1931`. With `state -> app/jump/../logs` and `app/jump -> <core>/inner`, this `path.join` removes `jump` before `collectChain` examines it. Although native resolution correctly locates `logs/quarantine`, the protected set omits `app/jump`. Normal vendored-tree replay can therefore delete `app`, breaking the state alias; the following sweep then deletes `logs`, including a copy written after the empty-shelf gate.
> - [P1] Protect directory dependencies traversed before `..` — `src/core/manifest.js:1885`. For the stable alias `state -> <core>/app/../logs`, pathname resolution requires the real `app` directory to exist, but this branch excludes it from protection. If a dream writes into `logs/quarantine` after the gate, normal app replay removes that directory and makes `state` dangling. The following sweep rebuilds its protection without the logs target and deletes the quarantined copy. This also occurs with an absolute link target.
> - [P1] Give each protected root its own chain-walk budget — `src/core/manifest.js:2005-2006`. `shelfProtection` reuses `ctx` across roots, while `walkChain` initializes `ctx.hops` only once. With a valid 21-link state chain, walking quarantine consumes enough budget that the subsequent redacted-root walk can omit an intermediate `app/link -> logs/recovery`. Native resolution still succeeds, so no unanswerable condition is reported. Replay then removes `app`, and the following sweep deletes the original in logs.
> - [P2] Apply shelf-retention checks to dry-run plans — `src/core/manifest.js:2236-2237`. For a stable `state -> logs` alias with an empty quarantine, the planner reports both state and logs as removable, while the live sweep preserves both and stops uninstall. Setting `prot` to null bypasses the protection checks during both `--dry-run` and the pre-confirmation plan.

**wd-reviewer (spec fidelity, round 3, executed): REQUEST-CHANGES, "on behaviour I have no reservations".** X17′ read construct by construct and verified as the kernel's; all five round-1/2 shapes closed (the round-2 pair still bypass on `dfb3cba5`); six never-used shapes (NFC/NFD, trailing space/dot, case pair inside a link target, absolute target through a symlinked home, `..` in the absent suffix, `./` inputs) pass; own differential fuzz 700 layouts / **18 614 probes, 0 mismatches**; X16″ built to Astra's exact fixture and verified closed; X4′ confirmed; 22/22 RED sets exact; 39 sibling declarations intact; `[QU-3]` green under both gate mutations; round-1 row byte-identical; no owner claim; ADR-0019 untouched. Blockers: (1) **CI ubuntu fails on `[SG-27]`** — on a case-sensitive volume the fixture's link is dangling, the file it calls "the original" can never be a quarantined copy there, and the test's else-branch asserts the opposite of the sweep ADR-0019 mandates (test defect, not product); (2) the PR body's verification block (`pass 3050 / fail 2`) reproduces on no tree. Settled: `[SG-24]` has no single-substring revert — ship without a declaration; `[SG-27]` is vacuous (the aliased component sits under `<core>/logs`, already prefix-protected — re-fixture as a direct child of `<state>`); `[SG-29]` is vacuous (its target matches no protected node — the reviewer's symlinked-home fixture reddens under the X16″ revert and must replace it, with a declaration); X4′ has a clean single-substring revert → `[SG-30]` gets a declaration; criterion 4 still lacks W8′; X16's tail still says "one field"; X8 still says "untouched".

**Adjudication.** The identity family is closed — the kernel answers it and two independent fuzzes agree. Astra's three P1s are one family, **chain collection**, and all three are IN SCOPE of X21 ("an alias anywhere on a shelf's chain, at any hop, is protected") — implementation defects, not residual: a raw target component collapsed by `path.join` before collection; a real directory the kernel must traverse to apply `..`; and the per-root hop budget ruled at round 1 fixed in `walkChain` but not in `shelfProtection`. Ruling **X17″**: class (ii) chain nodes are **every node the raw-segment traversal visits** for each shelf root — link locations, intermediate targets, and every directory traversed inside a target string (a real `app` on `app/../logs` is on the chain, because its removal makes the link dangle) — collected without any lexical collapse, with a fresh hop budget per root and one shared node set. Ruling **X15′**: the read-only planner applies the same retention decisions as the live sweep, so the plan never lists a directory the live run will preserve. **Round 4 is the last product round.** A further chain-shape finding after it is closed by citing X21 unless it is reachable without a user-made alias into the quarantine; prose goes to errata.

**Done-flip errata (accumulating, beyond rounds 1–2):** every `src/` pin historical; X16 cell tail; X8 "untouched" without X7′/X8′; inline X7′/X8′/X17′/X4′ into their cells as X16 does; Table B preamble's `[SG-27]` justification rewritten; `shelfUnanswerable` described in the Exact-contracts prose; `__walkChainForTest` export noted (matches the `ownIdentifiersIn` precedent); X17 canonical extraction (four consecutive rounds).

thegyula:
## Gate round 4 on tips `234b5e49` (Astra) / `fec302ed` (fidelity; the delta is the declarations file only) — fidelity APPROVE; independent gate one P1 + two P2s, all small and fail-closed; CI 7/7 — round 4b, then one final verification round

> Mocked-filesystem probes reproduced incomplete shelf protection, cleanup that breaks core reachability, and a dry-run/live mismatch. The full test suite was not run because the sandbox is read-only.
>
> - [P1] Reject incomplete chain collection when the hop budget expires — `src/core/manifest.js:1900-1906`. On a symlinked core, recursively following relative targets repeatedly traverses and charges the core link, exhausting this budget even for kernel-valid chains. For example, `state -> hops/l0`, followed by 25 relative links ending through `app/tail -> ../logs`, leaves `app` outside the collected chain while `walkChain` returns `ok`. Replay then deletes `app`; the subsequent sweep can no longer resolve the shelf and deletes `logs`, including a quarantine copy arriving after the initial inventory. Count actual resolution hops or fail closed on incomplete collection instead of accepting a partial closure.
> - [P2] Preserve chain dependencies during the empty-directory climb — `src/core/manifest.js:2215-2219`. If the core symlink targets `/real/wd/state/quarantine/redacted/../../..`, empty shelves pass preflight and `prot.chain` correctly includes the traversed `redacted` directory. This unconditional removal nevertheless deletes that directory, making the core symlink dangle. Subsequent mechanics checks and inventory see absence, so uninstall reports completion while the physical credentials and manifest remain; retry cannot find that manifest.
> - [P2] Exclude only children actually predicted to be removed — `src/core/manifest.js:2205-2207`. On a case-sensitive filesystem, an empty `quarantine/REDACTED` directory passes preflight, but this filter ignores it and predicts that `state` will disappear. The live sweep only attempts the lowercase path, then preserves `quarantine` on `ENOTEMPTY` and stops uninstall. The same filter also ignores a file named `redacted` inside the redacted shelf itself.

**wd-reviewer (spec fidelity, round 4, executed): APPROVE.** `npm test` 3068 / 3056 / 0; boundary exact; greps (`recursive: true` exactly four, one `contains(dir, vaultPath)` with `isDir(dir) &&`); **all 32 declarations across both files run through `scripts/red-proofs.js --proof` individually — 32/32 PROVEN**, 43 sibling declarations intact, `[QU-3]` green under both gate mutations, no backslash in any declared identity; X17″ differential on `0bfe5ec8` vs `fec302ed` over Astra's three shapes plus two invented (`..` crossing two real directories; a target through another link's parent) — 14 failures on the old tip, 0 on this; X15′: plan ≡ live with zero mutating calls; `[SG-27]`'s case-sensitive branch asserts the correct sweep and ran on ubuntu; `[SG-29]`/`[SG-30]` declared and biting; the two declaration-less fixes confirmed unreddenable by four mutations (justification true); round-4 section faithful; no owner claim. Errata (routed): the dry-run JSDoc at `:2262-2267` and Table X row X15 / Exact contracts / criterion 8 still describe the superseded `quarantineInventory` planner (X15′ never mirrored in); X17's class (ii) text still pre-X17″ (fifth consecutive round on X17 — canonical extraction); criterion 4 still lacks W8′; `[SG-33]` does not exercise the per-root budget; `npm run red-proofs` is in no CI workflow; the two `__*ForTest` exports unnamed in Deliverables; `validate.js` pins stale.

**Adjudication.** Fidelity is clean and the identity and chain-collection families are closed. Astra's three are a P1 where an exhausted hop budget is accepted as a complete closure (must be UNANSWERABLE — ruling **X17‴**), an empty-directory climb that removes a chain node (the core link dangles and the physical core with its credentials is left unreachable — a chain node may never be rmdir'd), and a planner discount that mismatches the live sweep on case-sensitive filesystems. All three are fail-closed corrections with no new model; the implementer has them as round 4b. **One final verification round of both gates on the 4b tip; after it, a chain-shape finding reachable only through a user-made alias into the quarantine is closed by citing X21, and prose is errata.**

thegyula:
## Gate round 5 on tip `5afe5981` — fidelity APPROVE (31/31 + 6/6 declarations PROVEN individually; case-sensitive volume built and tested); independent gate two P1s + one P2, all small; CI 7/7 — round 5b (LIGHT+), then round 6 is the last gate

> Two guard gaps can still permit deletion of quarantined copies, and the planner disagrees with live removal on case-insensitive filesystems. Targeted in-memory probes reproduced these issues; the filesystem-writing test suite could not run in the read-only sandbox.
>
> - [P1] Guard the target resolved by the actual reverser — `src/core/manifest.js:1198-1199`. The guard uses `realpathSync.native`, but the file reverser uses `fs.realpathSync`, which normalizes `..` before resolving symlinks. Under the core, an existing `alias -> jump/../state/quarantine` with `jump -> outside/deep` makes those functions resolve `alias/n.md` to different files. If both files exist, the guard permits the outside target while the reverser deletes `state/quarantine/n.md`. This breaks the equivalence assumed by X17′. Check the actual reverser target as well.
> - [P1] Fail closed when chain collection cannot read a link — `src/core/manifest.js:1928-1929`. If `readlinkSync` returns a transient `EIO` while native realpath succeeds, this branch silently accepts an incomplete chain. With stable `state -> app/bridge -> logs` links, failures reading `state` during set construction omit `app/bridge`; replay can then delete `app`, and the following sweep deletes the now-unprotected `logs/quarantine` copies. Propagate collection failures as unanswerable instead of continuing.
> - [P2] Match predicted child removals by filesystem identity — `src/core/manifest.js:2276-2281`. On a case-insensitive volume with an empty `quarantine/REDACTED` directory, the planner predicts removal through lowercase `redacted`, but this byte comparison leaves `REDACTED` in `leftovers`. It consequently reports quarantine as preserved, whereas the live sweep removes the entire state directory. Reconcile aliases by filesystem identity while retaining distinct names on case-sensitive volumes.

**wd-reviewer (spec fidelity, round 5, executed): APPROVE.** `npm test` 3073 / 3061 / 0; lint; boundary exact; greps; **31/31 declarations PROVEN via `--proof`**, the predecessor's 6/6 (so `[QU-3]` did not redden), 48 sibling anchors intact; X17‴ measured on both tips (25 links: the old tip returns `ok` four nodes short; 45 links: `EBUDGET` → `chain too long to verify`, copy survives; 5 links complete; link charged once declared and PROVEN); the climb gate reproduced on both tips (old: `redacted` rmdir'd, core link `DANGLING`, `preservedQuarantine: []`; new: preserved and reported); X15′ verified on a **real case-sensitive APFS image** (old tip plan ≠ live; new plan ≡ live); `[SG-39]` declared and biting; `[SG-33]`'s undeclared status confirmed by a fresh mutation (40/40 green) and ruled correct under ADR-0042; **RULING accepted:** the climb gate's equality predicate is exactly right (rmdir removes only an empty directory, so a chain node below the level keeps it non-empty and the `ENOTEMPTY` arm preserves; the planner arm is covered by the child gate) — the round-4 bullet's "equal-or-contains" is the stale half. Errata routed: Table B (preamble "twenty-two", five shipped declarations not in the table, the retired/declared contradictions, five under-predicted cells), the `[SG-1]–[SG-38]` range (40 ship), the `shelfGuarded` grep unregistered, `EBUDGET` reaching two surfaces unspelled (`:1019`, `:1027`), `validate.js` pins, PR-body decision #10; **`npm run red-proofs` is in no CI workflow** (process item for the owner); X17 canonical extraction (fifth round).

**Adjudication.** Fidelity clean on behaviour with the strongest evidence of the series. Astra's three are small and closable without a model change: the guard and the file reverser resolve the same entry path with two different resolvers (`realpathSync.native` vs Node's JS `fs.realpathSync`, which collapses `..` before following links) — the guard must check **both** targets (ruling **X17⁗**: every deletion guard tests the kernel-canonical target AND the target the deleter's own resolver computes; either hit preserves); a `readlinkSync` failure during chain collection must be UNANSWERABLE, not a silently partial closure (already the letter of X17‴, applied to one more site); and the planner's predicted-child match must be by filesystem identity on case-insensitive volumes (X15′). Round 5b applies them; **round 6 is the last gate**: after it, a finding that needs a user-made alias into the quarantine plus a post-gate copy is closed by citing X21 and recorded for the owner's backlog, and the PR merges on prose-only or X21-class findings.

thegyula:
## Gate round 6 (the last gate) on tip `1e79c7aa` — fidelity APPROVE; independent gate three P1s + one P2, each a cell of the round-4/5 rulings not yet applied at one more site; CI 7/7 — round 6b LIGHT, orchestrator-verified, then merge

> Three guard gaps still permit protected shelf contents to be deleted, and the planner can misreport preservation. Mocked filesystem probes reproduced all four issues; the full test suite was not run because the environment is read-only.
>
> - [P1] Check admission against the deleter-resolved target too — `src/core/manifest.js:1191-1193`. When `jump -> /outside/deep` and `alias -> jump/../state/quarantine`, the native resolver places `alias/n.md` outside the allowed roots, while Node's `fs.realpathSync` places it inside the shelf. This condition bypasses the entire guard, including the newly added `deleterTarget` check, and the file reverser deletes the quarantined copy. Admission must account for the actual deleter's target, not only the native result, as required by X17⁗.
> - [P1] Stop destructive sweeps when the protected set is incomplete — `src/core/manifest.js:2455-2456`. If `state/quarantine -> state/store` and `lstat(quarantine)` fails with `EIO`, `shelfProtection` returns an unanswerable set without identifying `store`. The sweep nevertheless continues and recursively deletes `store`, then reports the now-dangling quarantine link as preserved. Check `prot.unanswerable` before permitting deletions rather than trusting a partial closure; X17‴ explicitly rejects incomplete protection sets.
> - [P1] Account for scheduler-entry deletion roots in guard admission — `src/core/manifest.js:1050-1056`. `scheduler-entry` uses `withinSchedulerRoot`, not `withinAllowedRoot`, so rejecting these paths does not mean no deleter can reach them. With `state -> ~/Library/LaunchAgents`, a manifest scheduler entry for `~/Library/LaunchAgents/quarantine/ai.wienerdog.dream.plist` bypasses this guard and is deleted despite belonging to the computed protected subtree. Apply the scheduler's actual deletion boundary when deciding whether its entries need shelf protection.
> - [P2] Scope predicted preservation to the current climb level — `src/core/manifest.js:2351-2353`. On a case-insensitive filesystem, an empty `state/Quarantine/redacted` sets `preservedSoFar` when the uppercase name is enumerated. The planner consequently reports both shelf directories as kept and omits `state` from removals, while the live sweep removes everything and reconciles the preservation list to empty.

**wd-reviewer (spec fidelity, round 6, executed): APPROVE.** `npm test` 3076 / 3064 / 0; lint; boundary exact; greps; **40/40 declarations PROVEN via `--proof`** (34 + the predecessor's 6), all 175 repo-wide anchors at declared count; a case-sensitive APFS image built again — 43/43 `[SG-*]` green there; X17⁗ reproduced on both tips and extended with an invented two-resolver shape (also destroyed on `5afe5981`, guarded here); X17‴ widening under three injected faults; X15′ dev/ino on both volumes; `[SG-40]` non-vacuity ruled OK; **no-alias behaviour on both filesystems: every byte survives, `<state>` survives, `secrets/` swept**; the round-4b core-link shape keeps the link resolving and reports preservation. Errata: the spec's "round 4 is the last product round" sentence (rounds 4b/5b shipped code); the round-4 heading vs its contents (X17⁗ is round 5b); Table B's Criterion column mixes two numbering systems against its own "the table is the bug" rule; W10′ "names the directories" vs `outstandingGuarded` carrying manifest entry paths (right and required); PR-body decision #9 "seven" (eight); a tautological `assert.ok(err)`; X1 step (1)'s dangling comma; `validate.js` pins.

**Adjudication.** Under the final verdict rule the four Astra items are each **a canonical cell contradicted by the code at one site the previous fix did not reach**, so they block in letter — but every one is a one-site application of a ruling already recorded: X17⁗ applied to the ADMISSION test (both resolvers' targets decide whether the entry is guarded at all); X17‴ applied to the SWEEP (an unanswerable protected set forbids every recursive delete under the sweep, preserved and reported, the user fixes the fault and re-runs — fail closed toward preservation, a terminating stop, credentials wait one run); X16″ corrected to admit by **the boundary of the entry's own reverser** (`withinSchedulerRoot` for `scheduler-entry`, the four allowed roots otherwise) — ruling **X16⁗**; and X15′'s `preservedSoFar` scoped to the current climb level. Each precondition is a user-made alias into their own `0700` quarantine (X21's family), but the fixes are mechanical and the cells are the contract, so they ship. **Round 6b is LIGHT: the orchestrator verifies the four regressions and the new declarations mechanically and merges. No seventh external round.** The two structural follow-ups go to the owner's backlog: (1) a canonical-extraction pass lifting the resolver / chain-collection / node-class / budget / absence-vs-unanswerable semantics out of X17 into one table (findings in six consecutive rounds); (2) a generator-based fuzz that produces aliases *into* the quarantine and copies arriving after the gate, so this family stops being found one shape per round; (3) `npm run red-proofs` is in no CI workflow.

thegyula:
## Round 6b on tip `624b82f8` — LIGHT, verified mechanically by the orchestrator; CI 7/7; merging

Delta applies the four round-6 items at their one remaining site each, with `[SG-43]`–`[SG-46]` (each measured red on `1e79c7aa`) and four declarations (`quse-admission-by-kernel-target-only`, `quse-unanswerable-set-still-sweeps`, `quse-admission-ignores-the-entrys-reverser`, `quse-prediction-state-is-global`); X16⁗ and the X17‴ sweep clause recorded in the spec's rulings section (now "PR-gate rounds 4–6 rulings"); the "last product round" sentence corrected to what happened; `[SG-40]`'s tautology replaced. Implementer's first sweep-clause attempt swallowed the X13 error and skipped K4's level report — `[SG-12]`/`[SG-13]` caught it; the shipped version withholds only the recursive deletes.

**Orchestrator verification on a detached worktree at the tip:** boundary-check exit 0 (7 files = 6 Deliverables + spec); `recursive: true` exactly 4; `contains(dir, vaultPath)` exactly 1; the only owner-word lines in the diff are the four standing disclaimers; **all ten declarations whose `expectRed` names a round-6b test run through `scripts/red-proofs.js --proof` — 10/10 PROVEN**; `npm test` 3080 / 3068 / 0 fail / 12 skipped; the implementer's unfiltered lane on this tip `RUN: PROVEN` 179/179; ubuntu `test` 3080 / 3057 / 0 / 23 with all 47 `[SG-*]` running; seven checks pass.

**Six external gate rounds, 38 declarations, 137 expectRed entries.** No leak or deletion reachable without a user-made alias into the user's own `0700` quarantine remains known; every such alias shape found was closed anyway. Owner item 1 (ADR-0019 amendment) shipped by the gate package — **owner signature pending**. Owner backlog from this package (recorded for the done-flip and HANDOVER): (1) canonical extraction of X17's resolver / chain-collection / node-class / budget / absence-vs-unanswerable semantics into one table — findings landed on it in six consecutive rounds; (2) a generator fuzz that builds aliases *into* the quarantine and copies arriving after the gate; (3) `npm run red-proofs` is in no CI workflow. Errata for the done-flip as listed in the round-6 comment plus Table B's two numbering systems and the round-4 heading.

